PR #27, opened 2012-08-31
Enhanced suburb search with Nominatim

fraulyoner:
Hello,

we are a company from Karlsruhe (Germany) developing a web service for truck routing. For geocoding we use our own instance of Nominatim. Since the results searching for suburbs of a city with Nominatim's special phrases were not sufficient for us, we added some functionality. Now we get the suburbs (including the suburbs that are of type administrative) of a city by parsing Nominatim's detail page of this city. For this reason we added suburb.php (modified details.php) and two templates so that we get the results as json or html. 
Maybe someone else can need this functionality too.

Commit of Aug 23, 2012:
Added two templates (html, json) for displaying results of suburb
search. Used details.php as pattern to search for suburbs
(boundary:administrative, place:suburb, place:village), modified it,
result is suburb.php

### DIFF
--- a/lib/template/search-json.php
+++ b/lib/template/search-json.php
@@ -36,6 +36,9 @@
 		$aPlace['lat'] = $aPointDetails['lat'];
 		$aPlace['lon'] = $aPointDetails['lon'];
 		$aPlace['display_name'] = $aPointDetails['name'];
+		
+		//add shortname for further use
+		$aPlace['short_name'] = $aPointDetails['placename'];
 
 		$aPlace['class'] = $aPointDetails['class'];
 		$aPlace['type'] = $aPointDetails['type'];
--- a/lib/template/suburb-html.php
+++ b/lib/template/suburb-html.php
@@ -1,0 +1,40 @@
+<?php
+header("content-type: text/html; charset=UTF-8");
+?>
+<html>
+  <head>
+    <title>OpenStreetMap Nominatim: Suburbs</title>
+  </head>
+
+<?php
+
+if (sizeof($aParentOfLines)) {
+    $aGroupedAddressLines = array();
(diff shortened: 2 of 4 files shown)


derTobsch:
Hey, will anything happen here? Will it be merged or is there another way to do this? We are using this as well.

polapi:
Is the change going to be merged? I don't want to put any pressure, but I think it takes a little bit long...

matkoniecz:
Maybe after 5 years it would be a good idea to accept this change or explain why it is rejected and close PR?

sanitas2:
Some time ago I [expressed my opinion on mailing](https://lists.openstreetmap.org/pipermail/dev/2017-February/029682.html) list how the project is being managed. Among the other things I raised the issue of keeping (at that time) 5 year old pull request open. The answer was:

> One thing that contributes to the impression
> that "pull requests/issues are ignored" is that Nominatim aims to be a
> good, or at least a functioning, geocoder for the whole planet.
> Contributors (understandably - that's how Open Source works) often
> scratch their own itch, they find a problem with Spanish addresses and
> submit a fix - but they don't notice (or care) that it breaks geocoding
> elsewhere.

Why the original contributor was not notified about that is a mystery to me. Anyway, for some time such opened pull request annoyed me, but now I see it as an easy way to check on project.

@matkoniecz  but if you really want to see it closed, the best way to achieve it is to "disrespect" project maintainers.

lonvia:
This is unfortunately too specific for inclusion in the general code. I've left it open anyway because it was obviously useful to some people and so it seemed a good idea to make it easy for them to find. However, given that the main code now has significantly moved forward, it is time to close this.

Thank you  for your contribution.

PR #92, opened 2013-10-16
added option to ignore nonsearch terms in geocoding requests

manly:
The option is not enabled by default.
Allows frequently indexed search terms to be ignored rather than filtered out. This greatly help geocoding requests without a road type (ex: 'street' not specified in '500 rivard street') as well as Canadian geocoding requests.


lonvia:
the current master branch is doing something similar now per default. Closing here.

PR #93, opened 2013-10-27
Update postgresql.h

sroycode:
fix to compile on mac


lonvia:
I'm not sure why this compiled but it looks like it is geared toward one specific setup. The more generic way would be to find out where endianess and bswap are defined on mac, add a check in cmake and then include the proper file.

Closing this one as we have moved to cmake and a completely new PR will be needed.

PR #112, opened 2014-02-13
Handle external db servers

aredridel:
This adds what I think is everything needed to work with an external database server.

It doesn't handle the installation of `nominatim.so`, and in fact, I'm trying to deploy to Amazon RDS, so there's no way to load a C extension and I've dummied the two functions in my attempts.

It should however do everything else needed.

Password prompts happen. Use `.pgpass`.

### DIFF
--- a/utils/setup.php
+++ b/utils/setup.php
@@ -676,6 +676,7 @@
 
 	function pgsqlRunScriptFile($sFilename)
 	{
+		global $aCMDResult;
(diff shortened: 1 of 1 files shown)


lonvia:
Closing this, as the requested changes are still missing and #552 is similar but has less merge conflicts.

billsinc:
@aredridel can you elaborate more on your workaround for `nominatim.so`?

PR #120, opened 2014-03-22
Add more Italian articles and a conjunction

dforsi:
This patch adds the Italian word for "and" and two articles.


lonvia:
We need a different solution for stop words. I won't add any new ones as the existing ones already cause enough trouble.

PR #210, opened 2014-12-15
Added feature that can be used to to restrict results to belong to city specified in input address

Smarre:
I was getting results from different city, specifically because "Majavakatu 9, Vantaa" does not actually exist in Vantaa, but in Helsinki, there is Majavakatu 9, so it uses it as the street name wholly matches to it. So I added an optional check that will reject a match belonging to different city. It seems to fall back matching against only a street name further on street, so it actually will return the correct address from Vantaa with this code.

Strange debugging note: for some reason, without Wikipedia data, it will return the address in Vantaa anyway. I didn’t really investigate why that’s so as that token parser is fairly annoying to debug.

### DIFF
--- a/lib/Geocode.php
+++ b/lib/Geocode.php
@@ -1379,6 +1379,36 @@
 									$aPlaceIDs = $aRoadPlaceIDs;
 								}
 
+								// ensure it’s from same city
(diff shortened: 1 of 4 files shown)


lonvia:
This kind of check is rather specific and easier done on the client side. Ask for addressdetails=1 and then check for the city part.

PR #219, opened 2014-12-31
Fix case sensitivity of cyrillic ghe

mizabrik:
Fixes bug described in https://trac.openstreetmap.org/ticket/2758.


mizabrik:
The only solution that comes to my mind is to create tool, that re-normailzes all strings containing given characters. So, is modification of utils/update.php an appropriate solution?

lonvia:
Re-normalization would be a possibility, yes. If we go down that road, I'd prefer a general renormalization function that simply recomputes everything because then we could also change entries in the abbreviation table, which suffers from the same problem.

Note that it is not enough to go through the word table because different string might be pointing to the same normalized word before the change in the utf table but not anymore after. It really would be necessary to recompute the name_vector and nameaddress_vector columns in search_name.

lonvia:
I've merged this now as interpolations introduced a breaking change anyway. Note that the instance on osm.org currently doesn't run master. A reimport is planned in a few weeks.

lonvia:
Just to let you know, this change is now live on osm.org.

Issue #222, opened 2015-01-16
Allow multiple values in reverse lookup by OSM ID


saerdnaer:
Currently one can only ask for one reverse lookup by OSM ID per request:
Examples: 
 * http://nominatim.openstreetmap.org/reverse?format=json&osm_type=N&osm_id=271239092
 * http://nominatim.openstreetmap.org/reverse?format=json&osm_type=W&osm_id=27169996

If you want to use this endpoint to reverse lookup a list e.g. returned from an overpass query, you have to wait one second per result as the usage policy requests this.

I propose to add a osm_ids parameter, to allow to retrieve multiple addresses in one request, e.g. http://nominatim.openstreetmap.org/reverse?format=json&osm_ids=N271239092,W27169996

What's your opinion? Should I provide an patch?

lonvia:
Would be ok for me, although I wonder if it wouldn't be better to provide something like that in an extra endpoint, something like '/address' maybe? There should also be a server configurable limit, how many items can be retrieved.

saerdnaer:
Please ignore 52b21d8.

saerdnaer:
Looks like this issue can be closed:
https://nominatim.openstreetmap.org/lookup.php?osm_ids=R146656,W104393803,N240109189

Thanks to all involved!

lonvia:
There is still documentation missing at the http://wiki.openstreetmap.org/wiki/Nominatim wiki page. I wouldn't mind if somebody beats me to it.

mtmail:
I'll add that later today.

saerdnaer:
Here is my first first draft for the new section: http://wiki.openstreetmap.org/wiki/Nominatim#Address_lookup

I'm not so sure about the json_callback and accept-language parameters, so please delete these sections when they are not there.

mtmail:
awesome. Looks good to me

lonvia:
Wow, that was really quick. Thanks!

PR #223, opened 2015-01-18
New endpoint 'places' for looking up multiple osm_id's address #222

saerdnaer:


website/places.php
```diff
@@ -0,0 +1,70 @@
...
+			echo 'Bulk User: Only ' .  CONST_ReverseSearch_Max_IDs . " ids are allowed in one request.\n";
+			exit;
+		}
+		
+		$type = ''; 
+		$id = 0;
+		$aPlaces = array();
+		foreach ($osm_ids AS $item) 
+		{
+			$type = $item[0];
```
lonvia: Fails for empty items (e.g. places.php?osmids=N1,,N2)

website/places.php
```diff
@@ -0,0 +1,70 @@
...
+	{
+		$oPlaceLookup = new PlaceLookup($oDB);
+		$oPlaceLookup->setLanguagePreference($aLangPrefOrder);
+		$oPlaceLookup->setIncludeAddressDetails($bShowAddressDetails);
+		
+		$osm_ids = explode(',', $_GET['osm_ids']);
+		
+		if ( count($osm_ids) > CONST_Places_Max_ID_count ) 
+		{
+			echo 'Bulk User: Only ' .  CONST_ReverseSearch_Max_IDs . " ids are allowed in one request.\n";
```
lonvia: Should that be CONST_Places_Max_ID_count?

lib/template/places-json.php
```diff
@@ -0,0 +1,45 @@
+<?php
+	$aFilteredPlaces = array();
+
+	if (!sizeof($aPlaces))
+	{
+		if (isset($sError))
+			$aFilteredPlaces['error'] = $sError;
+		else
+			$aFilteredPlaces['error'] = 'Unable to geocode';
```
lonvia: I think it is better to return an empty list if none of the places could be found. (Same for XML response.) Is $sError even set anywhere?

website/places.php
```diff
@@ -0,0 +1,70 @@
...
+
+	if (strpos(CONST_BulkUserIPs, ','.$_SERVER["REMOTE_ADDR"].',') !== false)
+	{
+		$fLoadAvg = getLoadAverage();
+		if ($fLoadAvg > 2) sleep(60);
+		if ($fLoadAvg > 4) sleep(120);
+		if ($fLoadAvg > 6)
+		{
+			echo "Bulk User: Temporary block due to high server load\n";
+			exit;
```
lonvia: Use userError() here, which creates a proper 400 response. Same below.

lonvia:
I've added comments to the code for a few minor things. Other things I noticed:

 * I get odd results, when querying a nonexistent object, e.g. /places.php?osm_ids=N1. 
 * In the xml response the addressdetails are outside the result. The whole thing should look more like the answers to search queries, maybe even calling it `place` instead of `result`.

Could I also ask you to add basic tests similar to `tests/features/api/reverse_simple.php`? The minimum would be to test for valid json/xml. Bonus points for testing different input parameters.

website/places.php
saerdnaer: In my tests it did work. I only get an "Notice: Uninitialized string offset" when I add error_reporting(E_ALL); 
But I will add some code to handle this case better.

website/places.php
saerdnaer: of course

lib/template/places-json.php
saerdnaer: I basically used https://github.com/twain47/Nominatim/blob/master/lib/template/address-jsonv2.php as base for my new file. This part was copied from there.

saerdnaer:
> "I get odd results, when querying a nonexistent object, e.g. /places.php?osm_ids=N1."
can you explain or give an more details example what happens?

I can't really test the code in production, as I don't have a good testing environment.

mtmail:
There are some syntax errors in PHP (one bracket not closed) and Python (method called with too many variables). It's hard to write code when, as you say, you don't have a good testing environment setup. I will post my changes in a couple of hours and then it should be ready to be merged. All thanks goes to you of course!

@lonvia: the CONST_Places_Max_ID_count is set to 300 in settings.php. On my machine (with a remote worldwide database) looking for 300 existing nodes took 30 seconds. Should the limit be lower by default to avoid slowing down default Nominatim installations?

lonvia:
Yeah, I think a limit of 50 is sufficient.

As for test environment, it would be completely sufficient to have a very small Nominatim installation of, say, Liechtenstein to code and test such a change. Of course, that still means going through the whole installation procedure, which isn't the most simple or fast thing to do. I wonder if it's time to have a look at docker to provide a convenient environment where you only have to run the setup script.

mtmail:
I have https://www.vagrantup.com/ scripts for Nominatim I can share.

saerdnaer:
This pull request can be closed now, as #283 was merged which implemented the same feature.

PR #240, opened 2015-02-23
Added support for matching details in search response

agunnarsson:
Adds the option to get values matching the search input in the response. This is needed for external scoring calculation of the response.
Example: A user searches for the Swedish name "Eiffeltornet" and get "Eiffel tower" as response since using a client in English. To be able to evaluate the quality of the returned hit we would like to know it has the Swedish name "Eiffeltornet" too.

The implementation and usage follows the pattern of address details. Type is added to the beginning of each key to indicate the origin.
Request: ...&q=eiffeltornet&matchingdetails=1
Response: ..."matching":{"attraction:name:sv":"Eiffeltornet"}...

### DIFF
--- a/lib/Geocode.php
+++ b/lib/Geocode.php
@@ -6,6 +6,7 @@
 		protected $aLangPrefOrder = array();
 
 		protected $bIncludeAddressDetails = false;
+		protected $bIncludeMatchingDetails = false;
 
 		protected $bIncludePolygonAsPoints = false;
 		protected $bIncludePolygonAsText = false;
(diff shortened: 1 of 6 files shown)


lonvia:
A couple of comments about this approach.

You only get the exact matches missing the fuzzy matches that Nominatim made. For example, when searching for `Zurich`, Nominatim will match terms with `Zürich` but your match 
algorithm won't find those. Working with the token ids might be better but can be tricky.

I also suspect that the 'like' queries are relatively expensive. Again, the tokens might improve the situation.

In general, this looks like specifically geared towards the use case of reranking. I'm not sure how useful it is for general use. How about generalizing this a bit and instead extending the addressdetails to return the full name hash for each address line unconditionally if it matches or not. This would give you a lot more freedom in implementing your reranking algorithm (including better fuzzy matching than Nominatim is doing), be much more useful for other use-cases and could also be implemented more efficiently without any additional database queries.

agunnarsson:
Woups, I totally forgot to answer this issue. Yes, fuzzy matching would be preferable. Don't know if all information needed related to tokens is preserved. Maybe the best way would be to generate tokens on the fly for input and all unique details.

Yes, it would be geared towards re-ranking, However, returning all details would mean scoring calculations of how well each detail matches each search input. I guess it would be more expensive than some exact string matching, and scoring calculations on the result. That depends the amount of information that can be ruled out by the first step. Making that first step outside of Nominatim would mean duplicating some functionality, and I question if there would be much need for improved fuzziness when Nominatim already decided what should end up in the response.

lonvia:
We have the `namedetails` parameter now, which can be used to implement this client-side. Closing now here.

Issue #232, opened 2015-02-05
Get rid of way-nodes index on planet-osm-way


lonvia:
The `planet_osm_ways_nodes` index on the `planet_osm_ways` table is 120GB for a full planet and takes quite a few hours to create. Removing it would improve database size and import time significantly.

Currently it is used in exactly one place [here](https://github.com/twain47/Nominatim/blob/master/sql/functions.sql#L1433) to determine if a node is on some way. There is no way to reformulate this query efficiently using spacial operations on placex (in my experiments import of rank 30 slowed down by factor 3, i.e. increasing from 20 hours to 60 hours), so we'd need to get rid of this lookup as far as possible. A simple provision would be to move the node-is-on-way test after the check for addr:* tags and associated-street relations. That would leave only about 20% of nodes to be checked. But I wonder if it would be ok to get rid of the the check completely.

It's a bit difficult to check the DB for actual cases where this query has an effect on the indexing result, so some guessing is required. I can come up with three cases at the moment:

1. Interpolations. They should be anyways removed from placex into another table where the interpolations are computed on the fly. As an intermediate step, I'd suggest to change the code to compute parents for interpolations directly and just copy the result to the intermediate house numbers. That would save us already indexing of a few million points.
2. POIs/entrances on buildings. This case is now covered by the address inheritance from buildings [here](https://github.com/twain47/Nominatim/blob/master/sql/functions.sql#L1407). We'd only loose indirect inheritance from associated-street relations but I could live with that.
3. POIs like junctions and bus_stops on ways. That case should be sufficiently covered by finding the spatially nearest road.

@twain47, any other cases you had in mind for nodes on ways?

lonvia:
Already implemented.

PR #247, opened 2015-02-27
Avoid indexing of interpolated house numbers and drop way-node index

lonvia:
Combines two relatively major changes:

First of all, interpolated addresses are no longer separately indexed. Instead, parenting is done for the interpolation way and then simply copied to the interpolation points. That requires to identify the interpolated points, so the type has changed to the so far unused `place=address`.

The second change drops the `planet_osm_ways_nodes` index as already discussed in #232. So far, I've left all the parenting checks in. I've just reordered them so that the spatial lookup for ways on the node is avoided if anyhow possible.

Performance-wise the two changes cancel each other out in terms of indexing of rank 30. Using the spatial index for parenting slows down indexing by about 30% while at the same time about 30% of rank 30 was interpolated addresses which are no longer computed. What we gain are 120-150GB of space and 6-12 hours import time from the index creation.

Note: requires changes in osm2pgsql which are not yet part of this pull request, see https://github.com/lonvia/osm2pgsql/tree/remove-way-node-index


lonvia:
There was one more place in osm2pgsql where the index was used (but not needed). After fixing that, import and updates are running smoothly now. If there are no other comments, I'll merge this tomorrow. This change should be backward compatible for existing databases.

PR #251, opened 2015-03-03
Reverse search by multiple osm ids

mtmail:
This pull request and builds upon https://github.com/twain47/Nominatim/pull/223 for issue https://github.com/twain47/Nominatim/issues/222 by @saerdnaer.

Improvements:
* not-found places are not added to the output
* use of the `search-<format>.php` output templates. This makes parsing in our test suite and for existing client libraries easier
* lowered default maximum bulk requests from 300 to 50
* put tests into separate file
* make sure setup.php creates a URL redirect

My testcase for debugging was http://localhost:8089/nominatim/places.php?osm_ids=,R146656,W104393803,,,N240109189,N240109189,X1,N1,n,&format=xml
Note that the node N240109189 will be in the output twice. I guess it's arguable if the results should be de-duplicated. Results are ordered the way they're inputted in the request.

untitled2

I'm not happy with the name. `places` kind of describes the output and not the search method. I would prefer `lookup` because instead of a search we use a direct index.

Thoughts? 
Happy to make any adjustments.


tests/features/api/places.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Places by osm_type and osm_id Tests
+    Simple tests for internal server errors and response format.
+    These tests should pass on any Nominatim installation.
```
lonvia: Better to remove this comment. The tests as written require a planet installation.

tests/features/api/reverse_simple.feature
```diff
@@ -50,3 +50,4 @@ Feature: Simple Reverse Tests
         Then the result is valid json
         When looking up jsonv2 coordinates 36.791966,127.171726
         Then the result is valid json
+
```
lonvia: Nitpicking here but if you could get rid of this empty line change, that would be nice.

website/places.php
```diff
@@ -0,0 +1,91 @@
...
+	// $aPlaces (the format of reverse search) => $aSearchResults (the format of search)
+	$aSearchResults = array();
+	foreach ($aPlaces as $oPlace){
+		$oResult = $oPlace;
+		unset($oResult['aAddress']);
+		$oResult['address'] = $oPlace['aAddress'];
+		unset($oResult['langaddress']);
+		$oResult['name'] = $oPlace['langaddress'];
+		// importance score only adds confusiion: the results are ordered exactly as inputted
+		// unset($oResult['importance']);
```
lonvia: I would leave it in. Some people might find the information useful for further processing.

website/places.php
```diff
@@ -0,0 +1,91 @@
...
+				}
+			}
+		}
+	}
+
+
+	// we want to use the search-* output templates, so we need to convert
+	// $aPlaces (the format of reverse search) => $aSearchResults (the format of search)
+	$aSearchResults = array();
+	foreach ($aPlaces as $oPlace){
```
lonvia: Any particular reason you don't do that already in the loop above?

lonvia:
I like the reuse of the search template but I wonder if we should change the outer xml element (the one with searchresults), it doesn't really fit. 

Renaming from `places` to `lookup` sounds good.

As for deduplication, one argument in favour is to have less database lookups. I don't think we should guarantee any order of the results, just to keep options open for the future.

saerdnaer:
@mtmail Do you have time to finalize this pull request or should I try to continue it?

mtmail:
Almost forgot about this one. Will be ready tomorrow or the next day. It's near the top of my inbox again.

saerdnaer:
This pull request can be closed now, as #283 was merged which implemented the same feature.

PR #267, opened 2015-04-27
Add polygon simplification

a1exsh:
New query string parameter polygon_threshold=<0.0..1> is introduced.
The float value of this parameter (defaults to 0) is passed to
ST_SimplifyPreserveTopology() on geometry we're about to output in
one (or many) requested formats such as GeoJSON, KML, etc.

This is useful when getting border polygons for whole countries, but
rendering them at large scale, when most of the high resolution details
cannot be seen anyway.  For example, the unsimplified polygon data for
Germany in GeoJSON format currently makes for about 3 MB response body.
With use of this new parameter, the application can greatly reduce the
amount of downloaded data and server response time while providing its
users with the same picture.  On a typical laptop screen resolution,
zooming out to fit the whole country borders on screen, only 1/100 amount
of details could be well enough.

### DIFF
--- a/lib/Geocode.php
+++ b/lib/Geocode.php
@@ -1618,7 +1618,16 @@
 					if ($this->bIncludePolygonAsKML) $sSQL .= ",ST_AsKML(geometry) as askml";
 					if ($this->bIncludePolygonAsSVG) $sSQL .= ",ST_AsSVG(geometry) as assvg";
 					if ($this->bIncludePolygonAsText || $this->bIncludePolygonAsPoints) $sSQL .= ",ST_AsText(geometry) as astext";
-					$sSQL .= " from (select place_id,centroid,ST_SimplifyPreserveTopology(geometry,".$this->fPolygonSimplificationThreshold.") as geometry from placex where place_id = ".$aResult['place_id'].") as plx";
+					$sFrom = " from placex where place_id = ".$aResult['place_id'];
+					if ($this->fPolygonSimplificationThreshold > 0)
+					{
+						$sSQL .= " from (select place_id,centroid,ST_SimplifyPreserveTopology(geometry,".$this->fPolygonSimplificationThreshold.") as geometry".$sFrom.") as plx";
+					}
+					else
+					{
+						$sSQL .= $sFrom;
+					}
+
 					$aPointPolygon = $this->oDB->getRow($sSQL);
 					if (PEAR::IsError($aPointPolygon))
 					{
--- a/lib/Geocode.php
+++ b/lib/Geocode.php
@@ -12,6 +12,7 @@
 		protected $bIncludePolygonAsGeoJSON = false;
 		protected $bIncludePolygonAsKML = false;
 		protected $bIncludePolygonAsSVG = false;
+		protected $fPolygonSimplificationThreshold = 0.0;
 
 		protected $aExcludePlaceIDs = array();
 		protected $bDeDupe = true;
@@ -100,6 +101,11 @@
 		function setIncludePolygonAsSVG($b = true)
 		{
 			$this->bIncludePolygonAsSVG = $b;
+		}
+
+		function setPolygonSimplificationThreshold($f)
+		{
+			$this->fPolygonSimplificationThreshold = $f;
 		}
 
 		function setDeDupe($bDeDupe = true)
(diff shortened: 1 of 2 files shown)


lonvia:
Very nice. I'll review more closely in the next few days.

lib/Geocode.php
```diff
@@ -1612,7 +1618,7 @@ function lookup()
 					if ($this->bIncludePolygonAsKML) $sSQL .= ",ST_AsKML(geometry) as askml";
 					if ($this->bIncludePolygonAsSVG) $sSQL .= ",ST_AsSVG(geometry) as assvg";
 					if ($this->bIncludePolygonAsText || $this->bIncludePolygonAsPoints) $sSQL .= ",ST_AsText(geometry) as astext";
-					$sSQL .= " from placex where place_id = ".$aResult['place_id'];
+					$sSQL .= " from (select place_id,centroid,ST_SimplifyPreserveTopology(geometry,".$this->fPolygonSimplificationThreshold.") as geometry from placex where place_id = ".$aResult['place_id'].") as plx";
```
lonvia: I wonder if we should preserve the original SQL here for the frequent case that `polygon_threshold = 0.0`.

a1exsh: On Tue, Apr 28, 2015 at 9:38 PM, Sarah Hoffmann <notifications@github.com>
wrote:

> In lib/Geocode.php
> <https://github.com/twain47/Nominatim/pull/267#discussion_r29280174>:
>
> > @@ -1612,7 +1618,7 @@ function lookup()
> >  					if ($this->bIncludePolygonAsKML) $sSQL .= ",ST_AsKML(geometry) as askml";
> >  					if ($this->bIncludePolygonAsSVG) $sSQL .= ",ST_AsSVG(geometry) as assvg";
> >  					if ($this->bIncludePolygonAsText || $this->bIncludePolygonAsPoints) $sSQL .= ",ST_AsText(geometry) as astext";
> > -					$sSQL .= " from placex where place_id = ".$aResult['place_id'];
> > +					$sSQL .= " from (select place_id,centroid,ST_SimplifyPreserveTopology(geometry,".$this->fPolygonSimplificationThreshold.") as geometry from placex where place_id = ".$aResult['place_id'].") as plx";
>
> I wonder if we should preserve the original SQL here for the frequent case
> that polygon_threshold = 0.0.
>
Well, we totally could, but that would make the code a little bit more
complicated.  And I'm not sure if x == 0.0 works all that great with
floating point?  For that we could also set the default value to -1, and
check for < 0.

website/search.php
```diff
@@ -44,6 +44,8 @@
 		$bAsKML = (boolean)isset($_GET['polygon_kml']) && $_GET['polygon_kml'];
 		$bAsSVG = (boolean)isset($_GET['polygon_svg']) && $_GET['polygon_svg'];
 		$bAsText = (boolean)isset($_GET['polygon_text']) && $_GET['polygon_text'];
+		$fThreshold = 0.0;
+		if (isset($_GET['polygon_threshold'])) $fThreshold = (float)$_GET['polygon_threshold'];
```
lonvia: It would be nice to have this for format=html, too. Allows to try out the different degrees of simplification directly on the website.

lib/Geocode.php
lonvia: Problem is that SimplifyPreserveTopology is not cheap. For Germany I get 25 ms on average for `SELECT geometry FROM placex WHERE place_id = x` and 1073 ms for `select geometry from (select ST_SimplifyPreserveTopology(geometry, 0.0) as geometry from placex where place_id = x) as p`. I would check for x > 0.0.

a1exsh: On Tue, Apr 28, 2015 at 10:19 PM, Sarah Hoffmann <notifications@github.com>
wrote:

> In lib/Geocode.php
> <https://github.com/twain47/Nominatim/pull/267#discussion_r29283700>:
>
> > @@ -1612,7 +1618,7 @@ function lookup()
> >  					if ($this->bIncludePolygonAsKML) $sSQL .= ",ST_AsKML(geometry) as askml";
> >  					if ($this->bIncludePolygonAsSVG) $sSQL .= ",ST_AsSVG(geometry) as assvg";
> >  					if ($this->bIncludePolygonAsText || $this->bIncludePolygonAsPoints) $sSQL .= ",ST_AsText(geometry) as astext";
> > -					$sSQL .= " from placex where place_id = ".$aResult['place_id'];
> > +					$sSQL .= " from (select place_id,centroid,ST_SimplifyPreserveTopology(geometry,".$this->fPolygonSimplificationThreshold.") as geometry from placex where place_id = ".$aResult['place_id'].") as plx";
>
> Problem is that SimplifyPreserveTopology is not cheap. For Germany I get
> 25 ms on average for SELECT geometry FROM placex WHERE place_id = x and
> 1073 ms for select geometry from (select
> ST_SimplifyPreserveTopology(geometry, 0.0) as geometry from placex where
> place_id = x) as p. I would check for x > 0.0.
>
I've had some different figures, but I agree if the simplify function is
non-trivial in case of 0.0, we should check that ourselves.  Should I
submit an update to this pull request or a new one?

lonvia: There might be a difference between postgis versions. I've tested against 2.1.4.

You can update this pull request (if you know how, just amend the commit, if not, just add more commits).

a1exsh:
OK, I've added the suggested changes in a commit on top of the branch.  Please have a look :)

lonvia:
Merged, thanks a lot for this. The new feature will show up on nominatim.osm.org in a few days.

a1exsh:
Great, thank you!

PR #268, opened 2015-04-28
during reverse geocode lookup we also check the (optional) Tiger data

mtmail:
When we don't find a result with rank 30 (building name, house number etc) but find a road, then inside the US we also check the nearest Tiger house numbers who have the street as parent.

The `osm_type` is 'T' and the `osm_id` is the place_id. Not entirely sure if that makes sense but `Geocode.php` sets the same.

In `PlaceLookup.php` right after `if ($this->bIsTiger)` it would be possible to use a `UNION` to combine both SQL queries. But since it's a direct lookup by place_id and we already know the place is a Tiger place a single query is enough.

I've setup a server at http://46.101.149.108:9876/nominatim were you can test the results. For example http://46.101.149.108:9876/nominatim/reverse.php?format=xml&lat=40.6859534063625&lon=-112.055149368461 which will return `6882, Kings Estate Drive, West Valley City, Salt Lake County, 84128, United States of America`.

The server has only data loaded for the Salt Lake City, Utah area. See the bounding box on http://bboxfinder.com/#40.310949,-112.376404,41.391234,-111.549683 

This feature got sponsored/paid for by Guru Labs. Thanks a lot!

### DIFF
--- a/lib/ReverseGeocode.php
+++ b/lib/ReverseGeocode.php
@@ -69,6 +69,7 @@
 		{
 			$sPointSQL = 'ST_SetSRID(ST_Point('.$this->fLon.','.$this->fLat.'),4326)';
 			$iMaxRank = $this->iMaxRank;
+			$iMaxRank_orig = $this->iMaxRank;
 
 			// Find the nearest point
 			$fSearchDiam = 0.0004;
@@ -112,9 +113,8 @@
 				$bIsInUnitedStates = ($aPlace['calculated_country_code'] == 'us');
 			}
 
-
 			// Only street found? If it's in the US we can check TIGER data for nearest housenumber
-			if ($bIsInUnitedStates && $iPlaceID && $aPlace['rank_search'] == 26) 
+			if ($bIsInUnitedStates && $iMaxRank_orig >= 28 && $iPlaceID && ($aPlace['rank_search'] == 26 || $aPlace['rank_search'] == 27 )) 
 			{
 				$fSearchDiam = 0.001;
 				$sSQL = 'SELECT place_id,parent_place_id,30 as rank_search ';
--- a/lib/ReverseGeocode.php
+++ b/lib/ReverseGeocode.php
@@ -75,6 +75,8 @@
 			$iPlaceID = null;
 			$aArea = false;
 			$fMaxAreaDistance = 1;
+			$bIsInUnitedStates = false;
+			$bPlaceIsTiger = false;
 			while(!$iPlaceID && $fSearchDiam < $fMaxAreaDistance)
 			{
 				$fSearchDiam = $fSearchDiam * 2;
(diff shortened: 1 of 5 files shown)


lib/ReverseGeocode.php
```diff
@@ -107,12 +109,53 @@ function lookup()
 				}
 				$iPlaceID = $aPlace['place_id'];
 				$iParentPlaceID = $aPlace['parent_place_id'];
+				$bIsInUnitedStates = ($aPlace['calculated_country_code'] == 'us');
+			}
+
+
+			// Only street found? If it's in the US we can check TIGER data for nearest housenumber
+			if ($bIsInUnitedStates && $iPlaceID && $aPlace['rank_search'] == 26)
```
lonvia: rank_search == 27 might be relevant, too (named service roads, tracks and footways)

lib/ReverseGeocode.php
```diff
@@ -107,12 +109,53 @@ function lookup()
...
+					$iPlaceID = $aPlaceTiger['place_id'];
+					$iParentPlaceID = $aPlaceTiger['parent_place_id']; // the street
+				}
 			}
 
 			// The point we found might be too small - use the address to find what it is a child of
 			if ($iPlaceID && $iMaxRank < 28)
 			{
-				if ($aPlace['rank_search'] > 28 && $iParentPlaceID)
+				if ($aPlace['rank_search'] > 28 && $iParentPlaceID && !$bPlaceIsTiger)
```
lonvia: $iMaxRank contains the user-supplied `zoom` parameter. If it is less than 28, the TIGER lookup shouldn't be done in the first place.

tests/features/api/reverse.feature
```diff
@@ -11,3 +11,23 @@ Feature: Reverse geocoding
...
+    @Tiger
+    Scenario: TIGER house number
+        Given the request parameters
+          | addressdetails
+          | 1
+        When looking up jsonv2 coordinates 40.6863624710666,-112.060005720023
+        # Then exactly 1 result is returned
+        # Then result addresses contain
+        # | ID | house_number | road               | postcode | country_code
+        # | 0  | 7094         | Kings Estate Drive | 84128    | us
```
lonvia: This one only works against xml format, so use `When looking up coordinates` above. Also you should explicitly test for Tiger results with `And result 0 has not attributes osm_id,osm_type`.

The tests below don't do anything.

mtmail:
Thanks for reviewing.

   - added `rank_search==27` for non-highways
   - on smaller zoom levels (<17) we now skip looking at Tiger data
   - improved test. You're right, it even passed against an unchanged Nominatim installation

Running the test case:
`NOMINATIM_SERVER=http://46.101.149.108:9876/nominatim lettuce features/api/reverse.feature --tag=Tiger` => pass
`NOMINATIM_SERVER=http://nominatim.openstreetmap.org lettuce features/api/reverse.feature --tag=Tiger` => fail (expected)

lonvia:
Merged, thanks a lot. The new feature will be deployed on nominatim.osm.org in the next couple of days.

PR #274, opened 2015-05-03
update import script to use Tiger 2014 data

mtmail:
2014 is available and it parses fine. 3233 input files, 37,457,051 SQL lines created (compared to 37,359,579 for 2013). It's unlikely that anybody would import 2010 data so I removed that part from the script.

Once merged I'll update the instructions on http://wiki.openstreetmap.org/wiki/Nominatim/Installation as well.

### DIFF
--- a/utils/imports.php
+++ b/utils/imports.php
@@ -3,6 +3,7 @@
 
 	require_once(dirname(dirname(__FILE__)).'/lib/init-cmd.php');
 	ini_set('memory_limit', '800M');
+	$sDefaultDataPath = CONST_BasePath.'/data/tiger2014';
 
 	$aCMDOptions = array(
 		"Create and setup nominatim search system",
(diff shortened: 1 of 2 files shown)


utils/imports.php
```diff
@@ -3,6 +3,7 @@
 
 	require_once(dirname(dirname(__FILE__)).'/lib/init-cmd.php');
 	ini_set('memory_limit', '800M');
+	$sDefaultDataPath = CONST_BasePath.'/data/tiger2014';
```
lonvia: How about moving this into settings.php? Then it can be reused in `./utils/setup.php` below. Also, I'd leave out the year by default. It will just cause again confusion next year.

lonvia:
Merged with the additional change of adding the TIGER path to settings. I'm not sure how best to proceed with the wiki page as the current (and next) release will still have the old version.

Issue #33, opened 2013-01-06
Nominatim installation behind a proxy


xdriver:
I just tried to install Nominatim on a server behind a proxy. Unfortunately some of the scripts don't work out, because the proxy doesn't allow a direct connection. The following warning is reported:

Warning: file_get_contents(http://wiki.openstreetmap.org/wiki/Special:Export/Nominatim/Special_Phrases/VI): failed to open stream: Connection refused in {path to}/Nominatim-2.0.0/utils/specialphrases.php on line 55

According to the installation instructions at openstreetmap, the following commands seem to be effected:

"./utils/specialphrases.php --wiki-import > specialphrases.sql"
"./utils/update.php --import-osmosis-all --no-npi"

Is it possible to enhance the functionality to optionally enable a proxy?

Thanks!

PR #276, opened 2015-05-19
allow usage behind an HTTP Proxy

simonlopez:
This fixes the case where the server is not able to do direct http request and needs to use a proxy.

### DIFF
--- a/settings/settings.php
+++ b/settings/settings.php
@@ -9,6 +9,13 @@
 	@define('CONST_Max_Word_Frequency', '50000');
 	@define('CONST_Limit_Reindexing', true);
 
+	// Proxy settings
+	@define('CONST_HTTP_Proxy', false);
+	@define('CONST_HTTP_Proxy_Host', 'proxy.mydomain.com');
+	@define('CONST_HTTP_Proxy_Port', '3128');
+	@define('CONST_HTTP_Proxy_Login', '');
+	@define('CONST_HTTP_Proxy_Password', '');
+
 	// Software versions
 	@define('CONST_Postgresql_Version', '9.3'); // values: 9.0, ... , 9.4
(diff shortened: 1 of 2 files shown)


lonvia:
There was a different solution proposed quite some time ago in issue #33. Did you have a look at that?

simonlopez:
@lonvia I didn't foud the issue before, sorry. I've made changes to use @dhalbert proposal but with clean settings vars and support for proxy for https connexion if needed in future.

settings/settings.php
```diff
@@ -102,3 +109,26 @@ function user_busy_cost()
...
 	@define('CONST_Log_File_SearchLog', '');
 	@define('CONST_Log_File_ReverseLog', '');
+
+	// handle http proxy when using file_get_contents
+	if (CONST_HTTP_Proxy) {
+		$proxy = 'tcp://' . CONST_HTTP_Proxy_Host . ':' . CONST_HTTP_Proxy_Port;
+		$aHeaders = array();
+		if(CONST_HTTP_Proxy_Login != null && CONST_HTTP_Proxy_Login != '' && CONST_HTTP_Proxy_Password != null && CONST_HTTP_Proxy_Password != '') {
+			$auth = base64_encode(CONST_HTTP_Proxy_Login . ':' . CONST_HTTP_Proxy_Password);
+			$aHeaders = array("Proxy-Authorization: Basic $auth", "Authorization: Basic $auth");
```
lonvia: Is the Authorization header correct here? I'd expect that Proxy-Authorization is sufficient.

lonvia:
Thanks, that is so much nicer than a commented-out section. Two tiny cosmetic changes would be nice before that can be merged: remove the empty line change in `lib/lib.php` and move the section that configures the proxy into `lib/init-cmd.php`.

simonlopez:
@lonvia , I've made the suggested changes. Should be ok now.

PR #283, opened 2015-06-16
new method /lookup.php - Reverse search by multiple osm ids

mtmail:
This pull request and builds upon https://github.com/twain47/Nominatim/pull/223 for issue https://github.com/twain47/Nominatim/issues/222 by @saerdnaer. It also replaces https://github.com/twain47/Nominatim/pull/251 because it accumulated 11 commits, so this pull request is a bit cleaner.

Improvements:
* called `lookup.php` instead of `places.php`
* `importance` data is back on the output
* the `search-xml.xml` is reused and special care is taken to avoid uninitialized variables

My testcase for debugging was http://localhost:8089/nominatim/lookup.php?osm_ids=,R146656,W104393803,,,N240109189,N240109189,X1,N1,n,&format=xml
Note that the node N240109189 will be in the output twice. I guess it's arguable if the results should be de-duplicated. Results are ordered the way they're inputted in the request (though we don't guarantee that).

untitled

Test case output:

```
 NOMINATIM_SERVER=http://localhost:8089/nominatim lettuce features/api/lookup.feature

Feature: Places by osm_type and osm_id Tests                               # features/api/lookup.feature:1
  Simple tests for internal server errors and response format.             # features/api/lookup.feature:2

  @mtm
  Scenario: address lookup for existing node, way, relation         # features/api/lookup.feature:5
    When looking up xml places N158845944,W72493656,,R62422,X99,N0         # steps/api_setup.py:127
    Then the result is valid xml                                           # steps/api_result.py:83
    exactly 3 results are returned                                         # steps/api_result.py:117
    When looking up json places N158845944,W72493656,,R62422,X99,N0        # steps/api_setup.py:127
    Then the result is valid json                                          # steps/api_result.py:83
    exactly 3 results are returned                                         # steps/api_result.py:117

  @mtm
  Scenario: address lookup for non-existing or invalid node, way, relation# features/api/lookup.feature:14
    When looking up xml places X99,,N0,nN158845944,ABC,,W9                 # steps/api_setup.py:127
    Then the result is valid xml                                           # steps/api_result.py:83
    exactly 0 results are returned                                         # steps/api_result.py:117

1 feature (1 passed)
2 scenarios (2 passed)
9 steps (9 passed)
```


website/lookup.php
```diff
@@ -0,0 +1,93 @@
...
+	$hLog = logStart($oDB, 'place', $_SERVER['QUERY_STRING'], $aLangPrefOrder);
+
+	$aSearchResults = array();
+	if (isset($_GET['osm_ids']))
+	{
+		$oPlaceLookup = new PlaceLookup($oDB);
+		$oPlaceLookup->setLanguagePreference($aLangPrefOrder);
+		$oPlaceLookup->setIncludeAddressDetails($bShowAddressDetails);
+		
+		$osm_ids = explode(',', $_GET['osm_ids']);
```
lonvia: Minor complaint about style: please continue the variable style of camel case and tagging with the type (i.e. $aOsmIds) in php files. Also for $type and $id below.

website/lookup.php
```diff
@@ -0,0 +1,93 @@
...
+			{
+				$oPlaceLookup->setOSMID($type, $id);
+				$oPlace = $oPlaceLookup->lookup();
+				if ($oPlace){
+					// we want to use the search-* output templates, so we need to fill
+					// $aSearchResults and slightly change the (reverse search) oPlace
+					// key names
+					$oResult = $oPlace;
+					unset($oResult['aAddress']);
+					$oResult['address'] = $oPlace['aAddress'];
```
lonvia: Needs a if-isset, as aAddress is undefined when addressdetails=0 is given as a parameter.

website/lookup.php
```diff
@@ -0,0 +1,93 @@
...
+			}
+		}
+	}
+
+
+	if (CONST_Debug) exit;
+
+	$sXmlRootTag = 'lookupresults';
+	// we initialize these to avoid warnings in our logfile
+	$sQuery = '';
```
lonvia: I wonder if we might as well use that to list the ids queried.

lonvia:
Apart from my tiny nitpicking comments above it looks all good to merge now.

mtmail:
Thanks. I incorporated your feedback. For setting the query string in the output I made sure only the parts that looks valid are used as repeating raw user input is usually a bad idea.

PR #290, opened 2015-06-25
first version of Vagrant installation instructions

mtmail:
New files
* VAGRANT.md
* Vagrantfile
* vagrant-provision.sh

to setup a development environment inside a virtual machine.

### DIFF
--- a/VAGRANT.md
+++ b/VAGRANT.md
@@ -102,8 +102,8 @@ and then
 
 ## Running unit tests
 
-    cd ~/Nominatim/tests
-    phpunit
+    cd ~/Nominatim/tests-php
+    phpunit ./
 
 
 
--- a/VAGRANT.md
+++ b/VAGRANT.md
@@ -1,0 +1,167 @@
+# Install Nominatim in a virtual machine for development and testing
+
+This document describes how you can install Nominatim inside a Ubuntu 14
+virtual machine on your desktop/laptop (host machine). The goal is to give
+you a development environment to easily edit code and run the test suite
+without affecting the rest of your system. 
+
+The installation can run largely unsupervised. You should expect 1-2h from
+start to finish depending on how fast your computer and download speed
+is.
+
+## Prerequisites
+
+1. [Virtualbox](https://www.virtualbox.org/wiki/Downloads)
+
+2. [Vagrant](https://www.vagrantup.com/downloads.html)
+
+3. Nominatim 
+
+        git clone --recursive https://github.com/twain47/Nominatim.git
+
+    If you haven't used `--recursive`, then you can load the submodules using
+    
+        git submodule init
+        git submodule update
+
+
+
+## Installation
+
+1. Start the virtual machine
+
+        vagrant up
+
+2. Log into the virtual machine
+
+        vagrant ssh
+
+3. Import a small country (Monaco)
+
+    You need to give the virtual machine more memory (2GB) for an import, see `Vagrantfile`.
+    
+    See the FAQ how to skip this step and point Nominatim to an existing database.
+
+  ```
+  # inside the virtual machine:
+  cd Nominatim
+    wget --no-verbose --output-document=data/monaco.osm.pbf http://download.geofabrik.de/europe/monaco-latest.osm.pbf
+    utils/setup.php --osm-file data/monaco.osm.pbf --osm2pgsql-cache 1000 --all | tee monaco.$$.log
+    ./utils/specialphrases.php --countries > data/specialphrases_countries.sql
+    psql -d nominatim -f data/specialphrases_countries.sql
+    ```
+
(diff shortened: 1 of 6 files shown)


lonvia:
I've run in an odd issue with osm2pgsql not being able to find the protobuf lib:

~~~
==> default: depbase=`echo parse.lo | sed 's|[^/]*$|.deps/&|;s|\.lo$||'`;\
==> default:    /bin/bash ./libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H -I.  -pthread -I/usr/include/postgresql -I/usr/include/libxml2 -I/usr/include -I/usr/include -I/usr/include -DOSM2PGSQL_DATADIR='"/usr/local/share/osm2pgsql"' -Igeos-fallback -I/usr/include/lua5.2 -I/usr/include   -g -O2 -MT parse.lo -MD -MP -MF $depbase.Tpo -c -o parse.lo parse.cpp &&\
==> default:    mv -f $depbase.Tpo $depbase.Plo
==> default: libtool: compile:  g++ -DHAVE_CONFIG_H -I. -pthread -I/usr/include/postgresql -I/usr/include/libxml2 -I/usr/include -I/usr/include -I/usr/include -DOSM2PGSQL_DATADIR=\"/usr/local/share/osm2pgsql\" -Igeos-fallback -I/usr/include/lua5.2 -I/usr/include -g -O2 -MT parse.lo -MD -MP -MF .deps/parse.Tpo -c parse.cpp  -fPIC -DPIC -o .libs/parse.o
==> default: In file included from parse-pbf.hpp:34:0,
==> default:                  from parse.cpp:7:
==> default: fileformat.pb-c.h:7:35: fatal error: protobuf-c/protobuf-c.h: No such file or directory
==> default:  #include <protobuf-c/protobuf-c.h>
==> default:                                    ^
==> default: compilation terminated.
==> default: make[2]: 
==> default: *** [parse.lo] Error 1
~~~

The header can only be found in `/usr/include/google` on the VM but that path is somehow missing, yet `./configure` happily runs through.

mtmail:
That's weird. I destroyed my VM and recreated it but it doesn't report that error.

I do see 
```
==> default: libtool: link: ln .libs/parse-pbf.o .libs/libosm2pgsql.lax/lt1-parse-pbf.o || cp .libs/parse-pbf.o .libs/libosm2pgsql.lax/lt1-parse-pbf.o
==> default: ln:
==> default: failed to create hard link '.libs/libosm2pgsql.lax/lt1-parse-pbf.o' => '.libs/parse-pbf.o'
==> default: : Operation not permitted
```
Since . (/Nominatim) is really on the host computer it cannot create a hardlink from inside the VM to outside the VM. It didn't affect importing though.

vagrant-provision.sh
```diff
@@ -0,0 +1,166 @@
...
+
+
+
+
+
+###
+### Setup Apache/website
+###
+
+createuser -SDR www-data
```
lonvia: Needs a 'sudo -u postgres'.

VAGRANT.md
```diff
@@ -0,0 +1,167 @@
...
+
+    You need to give the virtual machine more memory (2GB) for an import, see `Vagrantfile`.
+    
+    See the FAQ how to skip this step and point Nominatim to an existing database.
+
+  ```
+  # inside the virtual machine:
+  cd Nominatim
+    wget --no-verbose --output-document=data/monaco.osm.pbf http://download.geofabrik.de/europe/monaco-latest.osm.pbf
+    utils/setup.php --osm-file data/monaco.osm.pbf --osm2pgsql-cache 1000 --all | tee monaco.$$.log
```
lonvia: I would recommend also catching errors, i.e "... -all 2>&1 | tee monaco.$$.log".

Also, there is a "./" missing at the beginning.

lonvia:
I've tried again with a fresh clone of Nominatim and with that it worked perfectly fine.

Two things I've noticed: first, the vagrant setup has overwritten my settings/local.php which was rather unexpected. How about only creating one, if there isn't one already? Second, I would prefer if the default memory settings could be tweaked so that the example import with monaco runs through. Either increase the memory of the virtual machine or decrease --osm2pgsql-cache in the example, possibly both.

mtmail:
Cool. I'll add your suggestions and experiment with the memory settings again.

mtmail:
Thanks again, I added your input

* allocated RAM: the documentation already said you need to allocate more before running the import, but 2GB is the default now
* `settings/local.php`. If the file is present, then a `settings/local-vagrant.php` gets created instead

With Vagrant already setup and the `apt-get` cache filled from a previous run I can recreate the VM (`vagrant destroy; vagrant up; vagrant ssh`) in 2.5 minutes. Importing Monaco takes another 2.5 minutes.

lonvia:
Merged. Thanks again.

PR #313, opened 2015-08-22
more tests for lib/lib.php

mtmail:
While chasing a bug I wrote some more tests. No feature changed, just more tests of existing features.


tests-php/Nominatim/NominatimTest.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 namespace Nominatim;
-require 'lib/lib.php';
+require '../lib/lib.php';
```
lonvia: This does not work for me. As the phpunit.xml is in the root directory, I also have to run phptest from the root directory. The README in test-php seems to be wrong on that subject or maybe I'm running the tests in the wrong way.

tests-php/Nominatim/NominatimTest.php
```diff
@@ -74,4 +74,76 @@ public function test_looksLikeLatLonPair()
...
+		// 20 words => 169766 sets
+		// 28 words => 397594 sets
+		$this->assertEquals(
+			8,
+			count( getWordSets(array_fill( 0, 4, 'a'),0) )
+		);
+
+
+		$this->assertEquals(
+			8,
```
lonvia: Shouldn't that be 397594? (It actually is 3505699 when I run it, not sure why).

lonvia:
Any news on this?

mtmail:
Thanks for pushing!

- `require '../lib/lib.php';` is according to documentation because `tests-php/README.txt` and `VAGRANT.md` both say/assume you run the command inside the `tests-php/` directory

- You're probably right on the loop count for 28 words. On my machine it runs out of memory (4GB) even trying. I changed the tests to only run 18 words which works if PHP gets 128M RAM.

PR #322, opened 2015-09-25
Add --drop option to setup.php

woodpeck:
This adds a --drop option to setup.php, allowing users to immediately remove all tables that are only required during import and for updates, but not for a readonly database. The --drop option can be used together with --all, or in a separate step with an existing database. 

The implementation is potentially a bit dangerous because it uses a positive selection of tables to keep, and deletes everything else - including any tables that the unsuspecting user might have manually created.

Using --drop will also get rid of a potentially created flat node file from osm2pgsql; it will not, however, propagate the --drop option to osm2pgsql itself.

### DIFF
--- a/utils/setup.php
+++ b/utils/setup.php
@@ -39,6 +39,7 @@
 		array('index-output', '', 0, 1, 1, 1, 'string', 'File to dump index information to'),
 		array('create-search-indices', '', 0, 1, 0, 0, 'bool', 'Create additional indices required for search and update'),
 		array('create-website', '', 0, 1, 1, 1, 'realpath', 'Create symlinks to setup web directory'),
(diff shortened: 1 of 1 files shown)


lonvia:
The `country_name` table is missing which causes errors in the postgres logs. Also, the 'place_classtype_*` tables for special phrases should be better kept. 

I'm going to make those changes and commit the feature with an experimental warning for now, so people can test it out if they like. In the long run, we really should have a list of tables to delete but I think that's not too hard to do later starting from this PR.

lonvia:
There was also the minor issue that the drop was always called even when `--drop` was not set. Fixed that, too, and merged everything.

Thanks for the code.

PR #345, opened 2016-01-04
Updated FIPS map in utils/tigerAddressImport.py

netsgnut:
The FIPS map, previously generated by an auto-script (which is defunct right now), does not reflect the current map. In particular, importing of newer TIGER data results in a Python error complaining of missing indices. This PR aims to fix these errors by adding the new county FIPS.

These changes can be found on [Changes to Counties or County Equivalent Entities](https://www.census.gov/geo/reference/county-changes.html) on the US Census Bureau in the 2000s and 2010s. Apart from adding new FIPS, name changes are also made to the list. Removal of obsolete FIPS codes, however, is not made in order to maintain backward compatibility towards older datasets.

Some other changes are made by coarse visual comparison with the [national county list](http://www2.census.gov/geo/docs/reference/codes/files/national_county.txt).


lonvia:
Thanks for the update.

PR #350, opened 2016-01-20
Mobile responsive

mtmail:
Compared to the last checkpoint I added
- responsive reverse-html.php
- responsive details-error-html.php
- shared header and footer section
- updated leaflet to 0.7.7
- more accurate distance calculation on details-html.php page

On the reverse search page the search position is also shown (can probably be improved) and you can click on the map to start a new search. I had to switch off the "continuous world" on zoom level 1 because clicking too far left ended up on e.g. longitude -350 and it's hard to explain to users why the search returns no results while the map shows places.

In the test suite I had to comment out `tidylib`. It can't parse/tidy HTML5 documents (neither can the W3C validator these days). The relevant tests pass.

I'd be grateful if you run the full test suite on your machine. Also the intro text is a placeholder, I wasn't sure what to put in there.

You mentioned it would be good if the "report problems with results" wouldn't default to the trac/git issue tracker if it's a non-OSM installation (which for example might have imported just a subset or outdated data). Not sure how to add that.


lonvia:
Really nice. A few minor points:

* I would prefer smaller headers, as the search results are the important part and should get as much screen estate as possible. I've made a pull request with an example of what I mean. Similarly, details could be a bit more compact to avoid scrolling (also, with my screen size, it isn't really obvious that there is a address section following after the main details because it is outside the screen.)
* Has leaflet keyboard controls? In particular pageup/down for zooming in and out. I use that a lot via two-finger-scroll on the trackpad.
* In reverse, when you click on the map, it should also add the current zoom to the reverse query (parameter zoom). Maybe even add zoom as a third input field.

Test have run fine.

mtmail:
- made header, titles use less space, same for tables on detail page. Thanks for the PR. Sitting behind 2x27" monitors I wasn't even aware
- keyboard controls (plus/minus keys) work but you have to focus the map (single click) first otherwise leaflet doesn't receive the events. If that's too confusing I can make the whole page (except the text boxes) intercept those and send them to the map
- I had scrollwheel zoom switched off because on touch devices (mobile, tables) if you scroll through a page you suddenly end up zooming instead of scrolling. I found a way to disable that for touch devices only. We can still assume that most users will use desktop browsers on that page anyway.
- for reverse search there's already a `zoom` parameter which restricts the search. I added a `prevmapzoom` parameter instead and it get ignored on initial (full world) pageview. If you want to debug the `zoom` feature we can also add an optional dropdown listing 2..21.
- thanks for running the tests. Running against my worldwide installations there were a couple of false positives

lonvia:
The compact design looks good now on my tiny screen and scrollwheel works like a charm.

For reverse, I was indeed talking about debugging the zoom parameter we already have. It would be cool to be able to zoom to the desired level on the map, click and get the result with zoom parameter attached. Maybe that idea is impractical, I don't know. A dropdown box would do as well.

settings/local-vagrant.php
```diff
@@ -0,0 +1,10 @@
+<?php
+   // General settings
+   // @define('CONST_Database_DSN', 'pgsql://@/nominatim');
+   @define('CONST_Database_DSN', 'pgsql://postgres@localhost:9999/nominatim');
+   // Paths
+   @define('CONST_Postgresql_Version', '9.3');
+   @define('CONST_Postgis_Version', '2.1');
+   // Website settings
+   @define('CONST_Website_BaseURL', 'http://localhost:8089/nominatim/');
+
```
lonvia: Was this file intentionally added? I understood it should be created locally.

mtmail:
You're right, I accidentally checked in a local file. I changed the zoom parameter to a dropdown. I took the labels from `ReverseGeocode.php`.

PR #358, opened 2016-02-02
Fix #4895

Xevib:
That change should fix this https://trac.openstreetmap.org/ticket/4895


lonvia:
We need a different solution for stop words. I won't add any new ones as the existing ones already cause enough trouble.

PR #378, opened 2016-02-26
spaces-to-tabs for all *.php files

mtmail:
The majority (90%, just guessing) of PHP files are intended by tabs. This PR converts the remaining files to make it consistent, The largest difference is the new header and footer HTML files.

I tested using `grep -r '^  '  lib/ website/ settings/ tests-php/ | grep '.php'` (that's two spaces)

A `git diff --ignore-space-change --ignore-blank-line` shows there were no other changes.

Let me know if you prefer spaces though. For the *.sql files it's mostly 2 space indentation, for the python files I see 4 space indentation.


lonvia:
I would prefer to leave the python and sql files as they are.

PR #379, opened 2016-02-27
First basic working version on poisnear feature, only xml render

jipipayo:
hi, i am building a new feature i think would be useful: Find points of interest near a point with a lat, long given, this code is based on the reverse location code.

sample url: [nominatimbaseurl]/poisnear.php?lat=40.40739&lon=-3.71533&class=aeroway&type=aerodrome

give me some feedback, thanks

### DIFF
--- a/lib/PoisNear.php
+++ b/lib/PoisNear.php
@@ -1,0 +1,69 @@
+<?php
+	class PoisNear
+	{
+		protected $oDB;
+		protected $fLat;
+		protected $fLon;
+		protected $iMaxRank = 28;
+
+		protected $aLangPrefOrder = array();
+
+		function __construct(&$oDB)
+		{
+			$this->oDB =& $oDB;
+		}
+
+		function setLanguagePreference($aLangPref)
+		{
+			$this->aLangPrefOrder = $aLangPref;
+		}
+
+		function setLatLon($fLat, $fLon)
+		{
+			$this->fLat = (float)$fLat;
+			$this->fLon = (float)$fLon;
+		}
+
+		function setRank($iRank)
+		{
+			$this->iMaxRank = $iRank;
+		}
+
+        function setClassAndType($sClass, $sType){
+            //TODO: make one only param poitype, coercing type and class mixed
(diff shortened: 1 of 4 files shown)


jipipayo:
any feedback on this feature? do you think it deserves to be added to the Nominatim main branch? feel free to say wathever you think about :smile:

mtmail:
Nominatim already has a similar feature when searching for `<keyword> near <place>`. For example

http://nominatim.openstreetmap.org/search.php?q=bus+stop+near+40.40739,-3.71533
http://nominatim.openstreetmap.org/search.php?q=airport+near+madrid
http://nominatim.openstreetmap.org/search.php?q=Aeropuertos+cerca+de+madrid

The keywords as well as 'near/in' are translated to multiple languages, see http://wiki.openstreetmap.org/wiki/Nominatim/Special_Phrases

Since it's part of the `/search.php` endpoint, bounded searches, XML/JSON output, number of results and other parameters already work.

I can see that setting a radius and using a class=type format can make sense, after all that's what we do internally in `Geocode.php` (look for `$sNearPoint`) when parsing the incoming query string, but so far the existing query parsing was good enough. Adding a new endpoint would duplicate some logic. Most importantly we'd have to make sure the endpoint also works for JSON, configurable radius and has a tests in our test suite (often more work that writing the feature).

Did you know about the radius search? Would that work for you?

jipipayo:
thanks for your kind feedback @mtmail . I didnt know this feature already exists in the search.php! great news because it works in xml and json formats, even with the "alias" words for class:key values. I completely agree  my feature is not necessary. You can close this pull request. 
One question: i noticed the radius is not working properly, maybe i could focus on fixing that.

on this sample in malaga city, i dont get airports near:
http://nominatim.openstreetmap.org/search.php?q=airport+near+36.71770215,-4.4092259483129&format=json&radius=2676
and the documentation of nominatim params do not say nothing about radius.

http://wiki.openstreetmap.org/wiki/Nominatim#Parameters

if you give me some advice i will try to fix the radius feature.

PR #381, opened 2016-02-28
Replace autotools with cmake and update osm2pgsql

lonvia:
This move is required because osm2pgsql has moved to cmake. It's also much faster.

The main difference is that we are now building out-of-tree, which requires the scripts to know about source and build directory. As a consequence, scripts must be run out of the build directory at the moment. To run:

    mkdir build
    cd build
    cmake ..
    make
    ./utils/setup.php ...

Still needs README and vagrant scripts adapted before it can be merged.


lonvia:
Also added test execution via ctest, although this will also execute the osm2pgsql tests.

This should be good to go now.

PR #388, opened 2016-03-02
Reverse geocode include geometry

mtmail:
This adds support for `&polygon=1`, `&polygon_kml`, `&polygon_svg=1` `&polygon_geojson=1`, `&polygon_text=1` to the reverse lookup. `&polygon_threshold=` for polygon simplification and the `CONST_PolygonOutput_MaximumTypes` setting is observed/supported.

Added relevant test for forward and reverse search. PHP and Python tests pass.

Even without the `&polygon*` parameters we now try to add a `boundingbox` element to the output.

As before if `CONST_Search_AreaPolygons` isn't set to `true` the bounding box gets calculated using a configured radius which is based on place class and type.

In the past when the result was a node we calculated the polygon-as-text (`$aPolyPoints`) differently depending on the `CONST_Search_AreaPolygons` setting. If set to `true` the logic was

    $iSteps = max(8,min(100,$fRadius * 3.14 * 100000));
    $fStepSize = (2*pi())/$iSteps;

... if not set the logic was ...

    $iSteps = max(8, min(100, ($fRadius * 40000)^2));
    $fStepSize = (2*pi())/$iSteps;

Now we always use the former because it was likely used more often in production.

When a `&polygon*` parameter is set we already did one additional database query for each result to fetch its geometry during forward search. For reverse search we now also have an additional database query.
 
The HTML details page already got the geometry from a direct SQL query so there was no change needed.


lonvia:
It's a bit difficult now because the code is so nicely generalised but would it be possible to not enable the 'polygon=1' format for reverse? It's terribly inefficient to generate and there are plenty of formats available that are easier to parse.

BTW: the branch could do with rebasing now.

lib/ReverseGeocode.php
```diff
@@ -165,7 +220,13 @@ function lookup()
...
 			return array('place_id' => $iPlaceID,
-					     'type' => $bPlaceIsTiger ? 'tiger' : 'osm');
+			             'type' => $bPlaceIsTiger ? 'tiger' : 'osm');
 		}
+
+
+
+
+
+
```
lonvia: Needs cleanup.

lib/template/address-json.php
```diff
@@ -24,6 +24,42 @@
...
 		if (isset($aPlace['sExtraTags'])) $aFilteredPlaces['extratags'] = $aPlace['sExtraTags'];
 		if (isset($aPlace['sNameDetails'])) $aFilteredPlaces['namedetails'] = $aPlace['sNameDetails'];
+
+		if (isset($aPlace['aBoundingBox']))
+		{
+			$aFilteredPlaces['boundingbox'] = array(
+				$aPlace['aBoundingBox'][0],
+				$aPlace['aBoundingBox'][1],
+				$aPlace['aBoundingBox'][2],
+				$aPlace['aBoundingBox'][3]);
```
lonvia: Why not just $aFilteredPlaces['boundingbox'] = $aPlace['aBoundingBox']?

lib/template/address-xml.php
```diff
@@ -27,6 +27,43 @@
...
 		if (isset($aPlace['lat'])) echo ' lat="'.htmlspecialchars($aPlace['lat']).'"';
 		if (isset($aPlace['lon'])) echo ' lon="'.htmlspecialchars($aPlace['lon']).'"';
+		if (isset($aPlace['aBoundingBox']))
+		{
+			echo ' boundingbox="';
+			echo $aPlace['aBoundingBox'][0];
+			echo ','.$aPlace['aBoundingBox'][1];
+			echo ','.$aPlace['aBoundingBox'][2];
+			echo ','.$aPlace['aBoundingBox'][3];
+			echo '"';
```
lonvia: use `join(',', $aPlace['aBoundingBox'])`

lonvia:
A general remark on style: the use of spaces is rather inconsistent. I would be strongly in favour of always having a space after commas.

PR #389, opened 2016-03-04
Reverse geocode include geometry (v2)

mtmail:
[ squaching (git rebase -i) failed miserably, thus I rather create a new pull request
@lonvia's feedback is incorporated:
- rebased to latest master branch
- disabled &polygon=1 for reverse search, adjusted tests
- cleaned up empty lines
- skip bounding box array creation if variable is already an array
- added space after comma where needed. I favor that, too.
]

This adds support for `&polygon_kml`, `&polygon_svg=1` `&polygon_geojson=1`, `&polygon_text=1` to the reverse lookup. `&polygon_threshold=` for polygon simplification and the `CONST_PolygonOutput_MaximumTypes` setting is observed/supported.

Added relevant test for forward and reverse search. PHP and Python tests pass.

Even without the `&polygon*` parameters we now try to add a `boundingbox` element to the output.

As before if `CONST_Search_AreaPolygons` isn't set to `true` the bounding box gets calculated using a configured radius which is based on place class and type.

In the past when the result was a node we calculated the polygon-as-text (`$aPolyPoints`) differently depending on the `CONST_Search_AreaPolygons` setting. If set to `true` the logic was

    $iSteps = max(8,min(100,$fRadius * 3.14 * 100000));
    $fStepSize = (2*pi())/$iSteps;

... if not set the logic was ...

    $iSteps = max(8, min(100, ($fRadius * 40000)^2));
    $fStepSize = (2*pi())/$iSteps;

Now we always use the former because it was likely used more often in production.

When a `&polygon*` parameter is set we already did one additional database query for each result to fetch its geometry during forward search. For reverse search we now also have an additional database query.
 
The HTML details page already got the geometry from a direct SQL query so there was no change needed.

### DIFF
--- a/lib/Geocode.php
+++ b/lib/Geocode.php
@@ -1,4 +1,6 @@
 <?php
+	require_once(CONST_BasePath.'/lib/PlaceLookup.php');
+
 	class Geocode
 	{
 		protected $oDB;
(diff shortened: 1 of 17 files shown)


lib/Geocode.php
```diff
@@ -1634,87 +1636,20 @@ function lookup()
...
-							$aPointPolygon['maxlat'] = $aPointPolygon['maxlat'] + $fRadius;
-						}
-						if (abs($aPointPolygon['minlon'] - $aPointPolygon['maxlon']) < 0.0000001)
-						{
-							$aPointPolygon['minlon'] = $aPointPolygon['minlon'] - $fRadius;
-							$aPointPolygon['maxlon'] = $aPointPolygon['maxlon'] + $fRadius;
-						}
-						$aResult['aBoundingBox'] = array((string)$aPointPolygon['minlat'],(string)$aPointPolygon['maxlat'],(string)$aPointPolygon['minlon'],(string)$aPointPolygon['maxlon']);
-					}
+					$aResult[$k] = $v;
```
lonvia: Looks like a simple array_merge().

website/reverse.php
```diff
@@ -63,6 +93,22 @@
...
+		$oPlaceLookup->setIncludePolygonAsSVG($bAsSVG);
+		$oPlaceLookup->setPolygonSimplificationThreshold($fThreshold);
+
+		$fRadius = $fDiameter = getResultDiameter($aPlace);
+		$aOutlineResult = $oPlaceLookup->getOutlines($aPlace['place_id'],$aPlace['lon'],$aPlace['lat'],$fRadius);
+
+		foreach($aOutlineResult as $k => $v)
+		{
+			$aPlace[$k] = $v;
+		}
```
lonvia: dito array_merge()

lib/template/search-json.php
```diff
@@ -18,11 +18,7 @@
 
 		if (isset($aPointDetails['aBoundingBox']))
 		{
-			$aPlace['boundingbox'] = array(
-				$aPointDetails['aBoundingBox'][0],
-				$aPointDetails['aBoundingBox'][1],
-				$aPointDetails['aBoundingBox'][2],
-				$aPointDetails['aBoundingBox'][3]);
+			$aPlace['boundingbox'] = $aPointDetails['aBoundingBox'][0];
```
lonvia: [0] looks out of place.

lonvia:
Looks almost good to go. Just one minor thing that looks like a bug and makes me wonder: do we have tests for bounding box output in the lettuce test part?

mtmail:
fixed. And I added tests

mtmail:
Not live yet, but this should fix
https://github.com/twain47/Nominatim/issues/376
https://github.com/twain47/Nominatim/issues/150
https://github.com/twain47/Nominatim/issues/138

lonvia:
It is live already on nominatim.osm.org. :) I've closed the other issues.

PR #404, opened 2016-03-16
fix bad paths to nominatim binary and partitionedtags.def in update.p…

IrlJidel:
fix bad paths to nominatim binary and partitionedtags.def in update.php and setup.php

introduces a new CONST  CONST_Nominatim_Binary

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,11 @@ configure_file(
     ${PROJECT_BINARY_DIR}/settings/phrase_settings.php
 )
 
+configure_file(
+    ${PROJECT_SOURCE_DIR}/settings/partitionedtags.def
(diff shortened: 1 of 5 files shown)


IrlJidel:
I had to update lib/init.php to do 

`require_once(CONST_BasePath.'/build/settings/settings.php');`

Things are getting ugly ;-)

lonvia:
Turns out I've commited a wrong version of the cmake port branch which explains why you were having so much trouble. I have now pushed the actual working version. The paths fixes have been added in 4b46c730884f09720f5c0c6ebdff44da8bfc0e1d. Thanks a lot for testing and finding this.

Unfortunately, the new version uses quite a different structure (copying the scripts instead of symlinking them), so you'd have to redo the build directory from scratch, I'm afraid.

PR #410, opened 2016-03-22
Do Tiger-LineInterpolation on demand and not at import time

markigail:
Import the tiger data as lines with start- and endnumber, not as interpolated housenumbers anymore. Do the interpolation later at search time.
Tested for California: reduces the storage demand from 14 GB to 585 MB  for the table location_property_tiger. Search time is not affected.
Additionally return osm_type='tiger' in json/jsonv2/xml request if it was a tiger housenumber.


sql/tiger_import_start.sql
```diff
@@ -32,7 +29,7 @@ BEGIN
 
   numberrange := endnumber - startnumber;
   rangestartnumber := startnumber;
-
+  
   IF (interpolationtype = 'odd' AND startnumber%2 = 0) OR (interpolationtype = 'even' AND startnumber%2 = 1) THEN
     startnumber := startnumber + 1;
     stepsize := 2;
```
lonvia: stepsize is unused.

website/details.php
```diff
@@ -10,7 +10,7 @@
 	   $fLoadAvg = getLoadAverage();
 	   if ($fLoadAvg > 3)
 	   {
-	   echo "Page temporarily blocked due to high server load\n";
+	   echo "Page temporarily blocked due to high server load\n";F
```
lonvia: This looks like an accidental commit.

lonvia:
Two more things about coding style: brackets should go on their own line and '=' should have a space before and after. (Sorry, that really should be written down somewhere.)

I still want to run it on the complete Tiger dataset now to make sure it passes ok for the entire US. Doing that now.

sql/tiger_import_start.sql
markigail: stepsize is used to filter out broken tiger data later, but rangestartnumber is unused.

markigail:
ok, thank you. I will fix it.

lonvia:
Truly impressive. Table size for Tiger data has dropped from 211GB (+ 120GB index) to 7.3GB (+ 1.4 GB index),

lonvia:
I'm getting a large number of test failures, mostly unrelated to this PR, but I need to fix those before merging.
The only one really caused by the Tiger changes is the server error for this query: /search?q=restaurant&addressdetails=1&bounded=1&viewbox=-87.7%2C41.9%2C-87.57%2C41.85 (make sure you have special phrases installed)

PR #439, opened 2016-04-26
Details in json format

agunnarsson:
This adds json fromat for details request. Some of the details has been made optional since loading is a bit expensive. 
New parameters for details are &addressdetails=1&linkedplaces=1&childplaces=1 default are false for all parameters details, but they have been included in links on html pages.

### DIFF
--- a/lib/template/details-html.php
+++ b/lib/template/details-html.php
@@ -161,6 +161,8 @@
 					if ($aPointDetails['wikipedia'])
 					{
 						kv('Wikipedia Calculated' , wikipedia_link($aPointDetails) );
+					} else if ($aPointDetails['aExtraTags']['wikipedia']) {
+						kv('Wikipedia Calculated' , wikipedia_link($aPointDetails['aExtraTags']) );
 					}
 
 					kv('Extra Tags'      , hash_to_subtable($aPointDetails['aExtraTags']) );
--- a/lib/template/details-html.php
+++ b/lib/template/details-html.php
@@ -42,7 +42,7 @@
 
 	function nominatim_link($aFeature, $sTitle)
 	{
-		return '<a href="details.php?place_id='.$aFeature['place_id'].'">'.$sTitle.'</a>';
+		return '<a href="details.php?place_id='.$aFeature['place_id'].'&addressdetails=1&linkedplaces=1&childplaces=1">'.$sTitle.'</a>';
 	}
 
 	function format_distance($fDistance)
--- a/lib/template/details-json.php
+++ b/lib/template/details-json.php
@@ -1,0 +1,91 @@
+<?php
+
+	function getDetails($aLine){
+		$aDetails = array();
+		if ($aLine['place_id'])
+		{
+			$aDetails['place_id'] = $aLine['place_id'];
+		}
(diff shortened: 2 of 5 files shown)


mtmail:
I like it.

One long-term goal is to split off the HTML frontend and let it talk to Nominatim only via the API (JSON/XML output). The reasoning is to make this project easier to manage in terms of writing tests, releasing or having the frontend in a different programming language. Your pull request is one step in that direction.

agunnarsson:
Sounds good! I did not include all information from the html version for now, but it's an easy task if/when the HTML frontend is updated.

lib/template/details-html.php
```diff
@@ -161,6 +161,8 @@ function _one_keyword_row($keyword_token,$word_id){
 					if ($aPointDetails['wikipedia'])
 					{
 						kv('Wikipedia Calculated' , wikipedia_link($aPointDetails) );
+					} else if ($aPointDetails['aExtraTags']['wikipedia']) {
+						kv('Wikipedia Calculated' , wikipedia_link($aPointDetails['aExtraTags']) );
```
lonvia: That's wrong as the wikipedia tag in ExtraTags is different. That is the originally tagged value. 'Wikipedia Calculated' is the wikipedia page that has been used to compute the importance.

lib/template/details-json.php
```diff
@@ -0,0 +1,91 @@
...
+
+	function wikipedia_link($sWikipedia)
+	{
+		if ($sWikipedia)
+		{
+			list($sWikipediaLanguage,$sWikipediaArticle) = explode(':',$sWikipedia);
+			return 'https://'.$sWikipediaLanguage.'.wikipedia.org/wiki/'.urlencode($sWikipediaArticle);
+		}
+		return '';
+	}
```
lonvia: code duplication from details-html.php. It's time to start a new file lib/output.php and move all the functions from details-*.php in there.

lib/template/details-json.php
```diff
@@ -0,0 +1,91 @@
...
+	$aPlace['extra_tags'] = $aPointDetails['aExtraTags'];
+
+	if ($aAddressLines) {
+		$aAddressDetails = array();
+		foreach($aAddressLines as $aAddressLine)
+		{
+			$aAddressDetails[] = getDetails($aAddressLine);
+		}
+		$aPlace['address_details'] = $aAddressDetails;
+	}
```
lonvia: missing: else { $aAddressLines = false; }

Without that you get an undefined variable error later. Dito for $aLinkedPlaces, $aParentOfLines below.

lonvia:
Some general comments:

  * Please have a look at the apache error logs when testing your code. There are many PHP errors thrown which you should resolve.
  * distance, admin_level and is_address need to be included in the address_details output
  * distance and admin_level are also missing for child and parent places
  * polygon output would be needed as well for an HTML frontend

lib/template/details-html.php
agunnarsson: Ok! Then originally tagged values should have a different title, like 'Wikipedia'?

agunnarsson:
Will have a look on the commented issues and get back with an update as soon as I can find some time.

lonvia:
Closing in favour of #968.

PR #442, opened 2016-04-26
add more build* directories to .gitignore

mtmail:
The vagrant script use 

    cd Nominatim
    sudo -u $USERNAME mkdir build-vagrant
    cd build-vagrant

Since it's temporary it's good to make sure it doesn't get checked into git accidentally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.pyc
 
 build
+build-*
 settings/local.php
 
 data/wiki_import.sql


lonvia:
Actually, I was thinking that we should change the vagrant scripts to put the build directory into the vagrant home directory, i.e. on the virtual disk. That way it doesn't interfere with the source code at all.

lonvia:
The new vagrant scripts now build inside the VM. This should solve your issue.

PR #445, opened 2016-04-30
first research script for ubuntu16. Not sure yet how to integrate

mtmail:
I tested Nominatim installation on a Ubuntu 16 LTE virtual machine. Not sure if we can support 14 and 16 in the same script.

Changes necessary compared to 14:
   * apache2 is not installed by default
   * php defaults to php7 (python defaults to 2.7)
   * postgresql and postgis got updated
   * packagename python-Levenshtein is now lowercase

### DIFF
--- a/VAGRANT.md
+++ b/VAGRANT.md
@@ -45,11 +45,11 @@ is.
 
   ```
   # inside the virtual machine:
-  cd Nominatim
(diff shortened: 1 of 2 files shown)


lonvia:
Thanks, I've integrated these changes in the new version of the vagrant scripts.

PR #465, opened 2016-06-08
Add osm_tag flag to reverse geocoding

SuberFu:
Add osm_tag to reverse geocoding will only return result that will
contain at least the given label tag.
For example, a osm_tag of "place:city" will return the nearest "place:city"

### DIFF
--- a/lib/ReverseGeocode.php
+++ b/lib/ReverseGeocode.php
@@ -6,6 +6,7 @@
 		protected $fLat;
 		protected $fLon;
 		protected $iMaxRank = 28;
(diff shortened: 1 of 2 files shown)


SuberFu:
A bit of a question for the "multiple osm_tag". Should I use comma or semicolon to delimit? For example.
Current: `place:city,place:town` searches for nearest city or town.
Semi-colon: `place:city;place:town` searches for nearest city or town.

SuberFu:
Using a different strategy instead of changing `reverse.php` behavior.

PR #469, opened 2016-06-12
Some refactoring of website php

lonvia:
* introduce accessor function for URL parameter
  * remove unused `bAsPoints` from reverse.php
  * update links to OSM objects (\browse\ no longer necessary, allow https)
  * factor out link formatting function and osm type translation
  * introduce chksql function for website code

Most important for the user-facing API is the more strict checking of input parameters. Illegal value now result in a 400 error instead of simply returning nothing.

### DIFF
--- a/lib/website.php
+++ b/lib/website.php
@@ -39,7 +39,7 @@ INTERNALFAIL;
 			echo "<pre>\n".$oSql->getUserInfo()."</pre>";
 		}
 
-		echo "</pre></p></body></head></html>";
+		echo "</pre></p></body></html>";
 		exit;
 	}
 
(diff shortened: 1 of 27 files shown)


lib/website.php
```diff
@@ -0,0 +1,140 @@
...
+		if (CONST_Debug)
+		{
+			var_dump($oSql);
+		}
+		else
+		{
+			echo "<pre>\n".$oSql->getUserInfo()."</pre>";
+		}
+
+		echo "</pre></p></body></head></html>";
```
mtmail: closing </head> misplaced here.

lib/website.php
```diff
@@ -0,0 +1,140 @@
...
+		exit;
+	}
+
+	function failInternalError($sError, $sSQL = false, $vDumpVar = false)
+	{
+		header('HTTP/1.0 500 Internal Server Error');
+		header('Content-type: text/html; charset=utf-8');
+		echo "<html><body><h1>Internal Server Error</h1>";
+		echo '<p>Nominatim has encountered an internal error while processing your request. This is most likely because of a bug in the software.</p>';
+		echo "<p><b>Details:</b> ".$sError,"</p>";
```
mtmail: should the comma be a dot in this case?

lonvia:
@mtmail Thanks, fixed.

I've also added a fix where the reverse HTML would not work anymore when zoom was not set. My error checking was a bit too strict.

mtmail:
Excellent refactoring. trac replaced with github URLs, no longer using a `_GET` value in left-assignment, lots of tiny good changes as well.

PR #476, opened 2016-06-21
travis-ci setup, some tests fail

mtmail:
Working setup for http://travis-ci.org/ TravisCI is free for open-source projects. 

The test run including import of Monaco takes about 12 minutes, 6000 lines output. Example: https://travis-ci.org/mtmail/Nominatim/builds/139336274

Some tests are currently failing. Those relying on a full world-install probably need to be tagged or excluded.

By default the github user who pushes to master (or a branch) gets an email on failure. Those notifications can be send to specific email addresses. Or switched off https://docs.travis-ci.com/user/notifications/

Each pull request is also checked. That can be switched off in the settings (travis-ci.org user interface).

Adding a status image to the `README.md` is trivial
https://docs.travis-ci.com/user/status-images/

* Took me 50 builds to get everything sorted.


lonvia:
Nice. Happy to see this activated here on the repo once you make it pass. (My preference would be to switch mail notification off and enable PR checking.)

I would leave out the API tests completely or restrict them to `features/api/*simple.feature`. Also exclude tagged tests: `-t -Fail -t -Tiger -t -poldi-only`.

I'm not sure about using `verbosity=2` for lettuce, makes the output very hard to read.

Can we hack this thing so that it fails if it produces PHP warnings?

mtmail:
Oh I forgot it's still activated. Happy to leave it running on my github account though if you wanted to move it to yours it's easy, too.

I'll make it pass and check if setting `E_ALL | E_STRICT` or similar will make PHP croak on warnings.

mtmail:
@lonvia Tests pass now, see https://travis-ci.org/mtmail/Nominatim/builds/148606262

I had to add another tag `@RiverRhein` because some tests assumed certain waterways are in the database.

Verbosity is now `1`, basically dots that show progress.

Notifications are switched off. (Only the person doing the checkin got notifications, not the project maintainers)

I didn't figure out how to catch PHP warnings. The default in `php.ini` is already `error_reporting = E_ALL & ~E_DEPRECATED & ~E_STRICT` and when running the test suite no error or warning gets written to `apache.log`.

lonvia:
Erm, those RiverRhein tests are not supposed to fail. They don't make any assumptions, they just add data to an empty database and then query it. If the data is not there, something is wrong. No test in features/db and features/osm2pgsql should fail unless marked with @fail (which none is at the moment).

PR #477, opened 2016-06-21
initialized variable in setup.php

mtmail:
PHP reported a `PHP notice` so it didn't break anything.


lonvia:
Fixed already in #486.

PR #487, opened 2016-07-24
Remove duplicate reverse geocoding code

lonvia:
The reverse geocoding code from lib/ was actually broken now,
so we use ReverseGeocode directly. Note that it currently
cannot support any interpolations, neither OSM nor Tiger. But
neither did the previous version.


mtmail:
Looks good. I noticed the same on Friday when reviewing the reverse logic.

BTW those `getParamFloat` and `chksql` functions made the source code much cleaner!

Issue #333, opened 2015-11-16
Different language specific abbreviations of the same word will not work


simonpoole:
Pro memoria.

Case at hand: street name including  "Santa"  ES, PT -> Sta. , IT -> S. (Nominatim handles the former).

Workaround use short_name on the street for the abbreviated name.

lonvia:
There is a similar clash for doktor between Russian (D-r) and German (Dr.), see also https://trac.openstreetmap.org/ticket/4961

PR #492, opened 2016-07-28
move #map-position to bottom right

damianblanck:
By moving map-position to bottom right, you can see more of the map and put position into white space in the bottom bar

http://nominatim.openstreetmap.org/css/search.css

#map-position {
  position: absolute;
  bottom: -90px;
  right: 20px;
  padding: 0 5px;
  color: #333;
  font-size: 11px;
  background-color: rgba(255, 255, 255, 0.7);
  z-index: 100;
}


lonvia:
I have no idea what exactly is going on with this pull request but it is completely bogus. Actually looks like a github bug because you should not be able to open a PR that comes from the twain47/Nominatim repo.

Can you please try again (create branch on your own fork, commit there, then open new PR from there). It would also be helpful if you could attach a screenshot of what the CSS change is trying to achieve.

lonvia:
Ah, found it. Never mind. Closing this one.

PR #493, opened 2016-07-28
Moves position out of map Update search.css

damianblanck:
Moves position out of map and into white space in bottom right.  Update search.css

### DIFF
--- a/website/css/search.css
+++ b/website/css/search.css
@@ -52,7 +52,7 @@ form label {
 
 #map-position {
   position: absolute;
-  top: 0;
+  bottom: -90px;
   right: 20px;
   padding: 0 5px;
   color: #333;


mtmail:
On my large monitor having the debug info on the top or bottom doesn't make much difference. The `-90px` just had a disadvantage if the lines are longer, then content will be cut. Did you want to move it out of focus or is the debug information not useful at all? I could hide it behind a "show mouse position" link or button instead.

damianblanck:
I think the info is usefull. Sometimes I have multiple windows open and don't have nominatim on the full screen. In those cases I found the info important but distracting. 
nominatimmove-map-position

lonvia:
Thanks for the screenshot. I like that change.

mtmail:
The other pull requests (https://github.com/twain47/Nominatim/issues/498) got merged (https://github.com/twain47/Nominatim/commit/cd6dcfa57), so we can close this. Not live on http://nominatim.openstreetmap.org/ yet

PR #496, opened 2016-07-31
PHP linting with CodeSniffer

mtmail:
New `php-link-rules.xml` for [CodeSniffer](http://pear.php.net/package/PHP_CodeSniffer/). Install using `sudo pear install PHP_CodeSniffer`.

A direct code diff is too complex, you'd need to compare revision by revision. The first two only change whitespace.

I started with the `PSR2` ruleset and fixed all errors. The moved to the more strict `PHPCS` ruleset. The `php-link-rules.xml` contains many exceptions because currently they would caused too many warning. In total I fixed about 4000 warnings and errors. (`PSR2` wants `elseif` while `PHPCS` wants `else if` and both don't agree on bracket rules for multiline if).

To run
`phpcs --standard=./php-lint-rules.xml --report-width=120 --tab-width=4 --colors lib/**.php website/**.php`

It still produces about 20 errors:
   * `TODO` statements cause warnings, can we disabled of course
   * Nominatim still uses PHP4 style constructors, worth fixing
   * formatting of Doc style comments on functions. Probably worth adding
   * one function uses snake_case instead of camelCase in its name

In the future it's worth enabling the maximum line limit again and force every class and function to have documentation.

### DIFF
--- a/lib/Geocode.php
+++ b/lib/Geocode.php
@@ -1,7 +1,7 @@
 <?php
 
-require_once(CONST_BasePath.'/lib/PlaceLookup.php');
-require_once(CONST_BasePath.'/lib/ReverseGeocode.php');
+require_once CONST_BasePath.'/lib/PlaceLookup.php';
+require_once CONST_BasePath.'/lib/ReverseGeocode.php';
 
 class Geocode
 {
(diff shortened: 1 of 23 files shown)


lonvia:
Thanks for going through that all. That's a lot of cosmetic changes to make the linter happy. Definitely catches a lot of good things but there are also some changes I don't like/find unnecessary:

 * change in bracket policy (keep opening bracket on single line)
 * enforced empty lines from PHPCS style
 * end-of-function comments behind closing brackets from PHPCS style

If we really must touch every single line of code (and I do see a point in getting rid of the initial indention step), we should discuss again if it wouldn't be better to switch from tabs to spaces. The tabs seem to trip a lot of people.

PR #498, opened 2016-07-31
HTML map: new button -show map bounds-

mtmail:
Replaces https://github.com/twain47/Nominatim/pull/493 by @damianblanck.

Map position information is hidden behind a button in the bottom right of the map. See screenshot.

untitled


lonvia:
When I click the button in the reverse view, the button disappears but nothing new appears.

General remark: I like the idea with the button but if we are making the area smaller with it, it is probably better to leave it in the upper right corner. The lower right corner is not visible on smaller screens.

mtmail:
not forgotten. Will fix either today or tomorrow.

mtmail:
works on /reverse now, moved to top-right corner.

PR #499, opened 2016-08-02
Travis ci setup - tests pass

mtmail:
Follow-up to https://github.com/twain47/Nominatim/pull/476 The river Rhein tests failed because the numpy package was missing in Ubuntu/centos/travis.


lonvia:
Jay, already one problem found because of CI. ;)

PR #502, opened 2016-08-11
PHP linting with CodeSniffer v2

mtmail:
Replacement for https://github.com/twain47/Nominatim/pull/496

I added more exceptions based on @lonvia's feedback and fixed warnings for `Geocode.php` and `search.php`. The rules file is cleaner and has more comments. If this looks good I can fix almost all other warnings.

The diff touches every line because one rule assumes files to have one indentation (one tab) less inside `<?php` blocks. I've split that into separate commits though.

Personally I prefer spaces so I don't mind switching tabs-to-spaces. Up to others to decide. https://www.youtube.com/watch?v=SsoOG6ZeyUI

### DIFF
--- a/lib/Geocode.php
+++ b/lib/Geocode.php
@@ -1,614 +1,559 @@
 <?php
-	require_once(CONST_BasePath.'/lib/PlaceLookup.php');
-	require_once(CONST_BasePath.'/lib/ReverseGeocode.php');
 
-	class Geocode
+require_once(CONST_BasePath.'/lib/PlaceLookup.php');
+require_once(CONST_BasePath.'/lib/ReverseGeocode.php');
+
+class Geocode
+{
+	protected $oDB;
+
+	protected $aLangPrefOrder = array();
+
+	protected $bIncludeAddressDetails = false;
+	protected $bIncludeExtraTags = false;
+	protected $bIncludeNameDetails = false;
+
+	protected $bIncludePolygonAsPoints = false;
+	protected $bIncludePolygonAsText = false;
+	protected $bIncludePolygonAsGeoJSON = false;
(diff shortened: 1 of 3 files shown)


lonvia:
Thanks for tweaking this some more. From what I understand, CodeSniffer doesn't really deliver rules for different code styles. We can only disable the rules we don't like or write our own. Option 1 is fine for most things in the custom style set here except for the bracket stuff. That I'd like to see checked, which basically means we have to switch style there.

So here is what I would suggest how we should go forward: 
 * get #498 and #504 merged as they are almost ready
 * reformat the entire php code base to using (4) spaces instead of tabs and CodeSniffers bracket style with the opening bracket at the end of the line
 * go forward with this PR

What do yo think?

mtmail:
Styles inherit from other base styles. The list of removed unwanted rules got long, but creating a positive list (copy&paste from the base styles) would be even longer.

I'm not sure I understand what you'd like checked on the bracket rules. All styles seem to favour `{` on the same line for control structures and conditions and on the next line for functions and classes.

I'll take care of the tabs->spaces PR ones those other two are merged. Then update/review this PR before I changing all files (boring work of 3-4 hours last time).

What I'd also like to introduce is a `Nominatim` namespace for the classes (separate PR later). That's mostly code style as well although it makes one error message clearer: we have a DB.php and it uses a PEAR library that contains a DB.php. The library might not be installed and the error message becomes hard to understand.

lonvia:
What I meant about bracket style: the rule set in this PR allows

    if (foo)
    {
       // blah
    }

but it does not enforce it, i.e this will also pass:

    if (foo) {
    }

At least that was my impression. And I could not find any base style that enforces the former.

PR #504, opened 2016-08-20
factor out parameter parsing into its own class

lonvia:
Allows to use the parsing functions in website/search.php and
utils/query.php as well.


mtmail:
I like it.

It'll be hard to write tests because the class reads global `$_GET`, `$_SERVER` and emits `userError` which isn't a really exception handling. Let's leave those for another time.

PR #506, opened 2016-08-22
specify ubuntu as primary Vagrant VM

mtmail:
By default if you ran `vagrant up && vagrant ssh` it started both Ubuntu and CentOS. That usually failed because both try to map the same ports. So start just one you had to use `vagrant up ubuntu` or `vagrant up centos`.

With this change the Ubuntu VM is started if no VM is specified. See "SPECIFYING A PRIMARY MACHINE" on https://www.vagrantup.com/docs/multi-machine/

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -14,7 +14,7 @@ Vagrant.configure("2") do |config|
       checkout = "no"
   end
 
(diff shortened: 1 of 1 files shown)


lonvia:
That just nicely scratched one of my itches. :)

PR #513, opened 2016-08-31
adding sudo where needed

joelmdev:
following commands in Ubuntu instructions, sudo was necessary on several commands but the instructions did not reflect this


lonvia:
Why do you think these sudos should be needed? If you have set $USERHOME to some arbitrary directory you should `chown $USER $USERHOME` before everything else and then be able to continue without any sudo.

lonvia:
Closing as there was no further comment.

PR #516, opened 2016-09-04
travis-ci.org changed from Postgres 9.2 to 9.1, while libpg is 9.5

mtmail:
A couple of weeks ago Travis-ci.org made it possible to specify which Postgres version one wants in the `.travis.yml` file. Before the default was an already started 9.2.
https://github.com/travis-ci/travis-ci/issues/4264
https://docs.travis-ci.com/user/database-setup/#PostgreSQL

Now the default became 9.1 but 9.2-9.5 were also present. So a `service postgresql stop` shows

```
        * Stopping PostgreSQL 9.1 database server
        * Stopping PostgreSQL 9.2 database server
        * Stopping PostgreSQL 9.3 database server
        * Stopping PostgreSQL 9.4 database server
        * Stopping PostgreSQL 9.5 database server   
```

Since that change when specifying 9.3, the version we used in the past, it caused trouble.
- cmake found `libpq.so` version 9.5.4

```
-- Found the following Boost libraries:
--   system
--   filesystem
-- Found PostgreSQL: /usr/lib/x86_64-linux-gnu/libpq.so (found version "9.5.4") 
```
- while `setup.php` found version 9.1

```
Create DB
Setup DB
Postgres version found: 9.1
CREATE EXTENSION
```
- and eventually the build broke `/module/nominatim.so: undefined symbol: palloc`

I ran 10 travis-ci builds to set versions and deinstall/purge/remove the other packages. The only working combination was to use version 9.5. Not ideal of course.

Successful build: https://travis-ci.org/mtmail/Nominatim/builds/157511690

### DIFF
--- a/vagrant/install-on-travis-ci.sh
+++ b/vagrant/install-on-travis-ci.sh
@@ -6,7 +6,8 @@
 # user 'travis'
 # $TRAVIS_BUILD_DIR is /home/travis/build/twain47/Nominatim/, for more see
 #   https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables
-# Postgres 9.2 installed and started. role 'travis' already superuser
+# Postgres 9.5, PostGIS 2.2 installed and started. role 'travis' already superuser.
+# https://docs.travis-ci.com/user/database-setup/#PostgreSQL
 # Python 2.7.10, pip 7.1.2
 
 # Travis has a 4 MB, 10000 line output limit, so where possible we supress
@@ -19,7 +20,7 @@ sudo apt-get update -qq
 sudo apt-get install -y -qq libboost-dev libboost-system-dev \
                         libboost-filesystem-dev libexpat1-dev zlib1g-dev libxml2-dev\
                         libbz2-dev libpq-dev libgeos-c1 libgeos++-dev libproj-dev \
-                        postgresql-server-dev-9.3 postgresql-9.3-postgis-2.1 postgresql-contrib-9.3 \
+                        postgresql-server-dev-9.5 \
                         apache2 php5 php5-pgsql php-pear php-db
 
 sudo apt-get install -y -qq python-Levenshtein python-shapely \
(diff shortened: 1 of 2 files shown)


vagrant/install-on-travis-ci.sh
```diff
@@ -19,7 +20,7 @@ sudo apt-get update -qq
 sudo apt-get install -y -qq libboost-dev libboost-system-dev \
                         libboost-filesystem-dev libexpat1-dev zlib1g-dev libxml2-dev\
                         libbz2-dev libpq-dev libgeos-c1 libgeos++-dev libproj-dev \
-                        postgresql-server-dev-9.3 postgresql-9.3-postgis-2.1 postgresql-contrib-9.3 \
+                        postgresql-server-dev-9.5 \
```
lonvia: Postgis gets automatically installed for 9.5?

mtmail: It's a bit hidden on https://docs.travis-ci.com/user/database-setup/#PostgreSQL "All installed versions of PostgreSQL include PostGIS 2.1"

lonvia:
It would indeed be nice to run travis with the oldest supported version of postgres to make sure that still works. pglib of a higher version shouldn't be a problem. It's the postgresql-server-dev package that needs to match the server version. So, if you want to run against 9.1 server install postgresql-server-dev-9.1 (and only that one). It should also be possible to change the port, when multiple postgres are running, to connect to a different server version.

lonvia:
On second thought: our current cmake setup might actually not allow to detect different versions for the client and server libs for postgresql.

lonvia:
We now have Travis enabled on the repository. Do you mind rebasing this PR on the current master?

mtmail:
Awesome! Will do.

mtmail:
Looks like I abandoned the branch and continued with version 9.5 and those changes got merged already. My master is the same as openstreetmap/Nominatim/master and when you do an empty push you should see a build result like https://travis-ci.org/mtmail/Nominatim/jobs/208789762

Two test failures related to `db/update/linked_places.feature`. Are they expected?

lonvia:
No, there shouldn't be any test failures. I'll look into it then and open a new PR with fixes for those. Thanks for checking.

Closing this one as abandoned.

PR #517, opened 2016-09-04
tabs-to-spaces

mtmail:
One step in preparation for https://github.com/twain47/Nominatim/pull/502#issuecomment-241196058

* changed tabs to 4 spaces
* If all but a couple of lines in a file was indented, removed one level of indentation
* removed two `?>` (end of a `<?php` block) from files. Those are optional anyway
* Added a couple of empty lines for readability
* Checked using `git diff --ignore-all-space` for any side effects/changes. None other than listed above.
* Checked using `for f in `find . -type f | grep -v git | grep -v '.png' | grep -v osm2pgsql | grep -v ./nominatim | grep -v ./module | grep -v ./build | grep -v /leaflet | grep -v Makefile`; do echo $f; egrep '^\t' "$f"; done` which files still have lines starting with tabs. C++ lines and osm2pgsql files mostly.
* Ran all tests in ubuntu VM and they pass
* I'm pretty sure travis-ci build also pass once https://github.com/twain47/Nominatim/pull/516 is applied


lonvia:
Thanks for that. I've added a commit that adapts CONTRIBUTING.md. The other change we should probably get done is the change of bracket style, so that Code Sniffer is happy. That's a big one, too. Do you still feel like doing it? Otherwise I can do the boring work for a change.

mtmail:
I'll take care. I have two flights and three long train rides this week which is ideal for that kind of changes/PRs.

PR #525, opened 2016-09-11
fix indentation and misc errors according to PSR2 coding style guide

mtmail:
Fixes most open errors according to the PSR2 standard
https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md

To be precise
```
* Perl-style comments are not allowed (# instead of //)
* Line indented incorrectly
* Whitespace found at end of line
* Expected 2 blank lines before function
* Expected 1 blank line at end of file
* Expected 1 space before comment text but found 20; use block comment if you need indentation
* Function closing brace must go on the next line following the body
* TRUE, FALSE and NULL must be lowercase
* AS keyword must be lowercase
* Terminating statement must be indented to the same level as the CASE body
```

Tests pass.

I believe it found a couple of bugs this time, "only" in the importWikipedia script though. There were several `case` statements without an explicit `break`. Looking at the logic I added those. Some `case` statements ended in semicolons instead of colons.

In `php-lint-rules.xml` I added ignoring of
```
* Blank comments are not allowed
* Equals sign not aligned with surrounding assignments
* There must be no blank line following an inline comment
* "A file should declare new symbols (classes, functions, constants, etc.) and cause no other side effects, or it should execute logic with side effects, but should not do both."
```

We only have three errors remaining, which can be handled in the next PR. 
```
* Method name "Geocode::Geocode" is not in camel caps format
* Each class must be in a namespace of at least one level (a top-level vendor name)
* Visibility must be declared on method
```

After that we can add style checks to the documentation. Maybe to the test suite. Or even a third-party service like https://styleci.io/ (free for open source).

### DIFF
--- a/lib/Geocode.php
+++ b/lib/Geocode.php
@@ -48,6 +48,7 @@ class Geocode
     protected $sQuery = false;
     protected $aStructuredQuery = false;
 
+
     function Geocode(&$oDB)
     {
         $this->oDB =& $oDB;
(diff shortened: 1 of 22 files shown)


lonvia:
Would it be much work to split this commit and move the 'fixes' into a separate one (not a separate PR, just two commits in this one, feel free to force-push on this branch)? It would make it easier to review and keep a better record for later.

mtmail:
You mean the changes to `php-lint-rules.xml` in one and the fixes in another? No problem

lonvia:
I mean fixes like adding breaks, i.e. where you have changed any php code.

mtmail:
ok, will do

mtmail:
replacing this with https://github.com/twain47/Nominatim/issues/528

PR #528, opened 2016-09-14
Psr2 indentation revised

mtmail:
#Fixes most open errors according to the PSR2 standard
https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md

To be precise
```
* Perl-style comments are not allowed (# instead of //)
* Line indented incorrectly
* Whitespace found at end of line
* Expected 2 blank lines before function
* Expected 1 blank line at end of file
* Expected 1 space before comment text but found 20; use block comment if you need indentation
* Function closing brace must go on the next line following the body
* TRUE, FALSE and NULL must be lowercase
* AS keyword must be lowercase
* Terminating statement must be indented to the same level as the CASE body
```

Tests pass.

I believe it found a couple of bugs this time, "only" in the importWikipedia script though. There were several `case` statements without an explicit `break`. Looking at the logic I added those. Some `case` statements ended in semicolons instead of colons.
[ replaces https://github.com/twain47/Nominatim/pull/525, split into three commits ]

In `phpcs.xml` I added ignoring of
```
* Blank comments are not allowed
* Equals sign not aligned with surrounding assignments
* There must be no blank line following an inline comment
* "A file should declare new symbols (classes, functions, constants, etc.) and cause no other side effects, or it should execute logic with side effects, but should not do both."
```
and removed several exceptions which were relevant for the PEARCS style standard but no longer caused warnings when using PSR2 as base.

To run `phpcs --report-width=120 --colors lib/**.php website/**.php utils/**.php`

We only have three errors remaining, which can be handled in the next PR. 
```
* Method name "Geocode::Geocode" is not in camel caps format
* Each class must be in a namespace of at least one level (a top-level vendor name)
* Visibility must be declared on method
```

After that we can add style checks to the documentation. Maybe to the test suite. Or even a third-party service like https://styleci.io/ (free for open source).


utils/importWikipedia.php
```diff
@@ -389,15 +389,16 @@ function nominatimXMLEnd($hParser, $sName)
             break;
         case 'prefecture japan':
             $aRecord['name'] = trim(str_replace(' Prefecture', ' ', $aRecord['name']));
+            break;
```
lonvia: This looks like the fallthrough is intended. Like all case statements directly below a prefecture is a state.

utils/importWikipedia.php
```diff
@@ -415,6 +416,7 @@ function nominatimXMLEnd($hParser, $sName)
             break;
         case 'settlement':
             $bUnknown = true;
+            break;
```
lonvia: dito

mtmail:
You're right on the two breaks, they should be omitted after all. (The PSR2 linting requires to add a comment if a `break` is not in place.)

PR #539, opened 2016-09-29
There is no build folder

dogmatic69:


mtmail:
In the "Building and Configuration" section of the document
https://github.com/twain47/Nominatim/blob/master/docs/install-on-ubuntu-16.md
a build folder gets created.

PR #552, opened 2016-10-15
allow connecting to a host on a different server

dogmatic69:
Currently its not possible to configure when the psql server is not `localhost`. This fixes that

### DIFF
--- a/utils/setup.php
+++ b/utils/setup.php
@@ -80,7 +80,6 @@ if (isset($aCMDResult['osm2pgsql-cache'])) {
 
 $aDSNInfo = DB::parseDSN(CONST_Database_DSN);
 if (!isset($aDSNInfo['port']) || !$aDSNInfo['port']) $aDSNInfo['port'] = 5432;
-
 if ($aCMDResult['create-db'] || $aCMDResult['all']) {
     echo "Create DB\n";
     $bDidSomething = true;
@@ -88,7 +87,14 @@ if ($aCMDResult['create-db'] || $aCMDResult['all']) {
     if (!PEAR::isError($oDB)) {
         fail('database already exists ('.CONST_Database_DSN.')');
     }
-    passthruCheckReturn('createdb -E UTF-8 -p '.$aDSNInfo['port'].' '.$aDSNInfo['database']);
+
+    passthruCheckReturn(implode(' ', array(
+        'createdb -E UTF-8',
+        '-p ' . $aDSNInfo['port'],
+        '-h ' . $aDSNInfo['hostspec'],
+        '-U ' . $aDSNInfo['username'],
+        $aDSNInfo['database']
+    )));
 }
 
 if ($aCMDResult['setup-db'] || $aCMDResult['all']) {
@@ -96,7 +102,6 @@ if ($aCMDResult['setup-db'] || $aCMDResult['all']) {
     $bDidSomething = true;
 
     // TODO: path detection, detection memory, etc.
-    //
     $oDB =& getDB();
 
     $fPostgresVersion = getPostgresVersion($oDB);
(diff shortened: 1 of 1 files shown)


lonvia:
This breaks the standard server setting `pgsql://@/nominatim`. I'd also like to see some solution for providing the `module/nominatim.so` on the other server (for example by adding the ability to set a custom location for nominatim.so). At least there should be a sensible error message, if nominatim.so is not found.

Please see also #112 , in particular https://github.com/twain47/Nominatim/pull/112#issuecomment-41839565

dogmatic69:
Should be easy enough to make this work `pgsql://@/nominatim`

As for the .so file, I have just copied it to the location on the server. Anyone can easily `ln -s` the file to where ever it needs to be. 

Or go all out and make it a configuration option.

lonvia:
I'd lean toward a config option in settings/local.php for the path that is mandatory to be set when the host is something other than localhost but I'm open for other suggestions. The problem with nominatim.so is that the import breaks in non-obvious ways when it is missing and I'd like to avoid situations where people naively set a different host then report a bug because it doesn't work.

billsinc:
This worked for me for #566 except for the functions dependent on `nominatim.so` as RDS doesn't permit that.  Still looking for a workaround.

billsinc:
I believe this also doesn't fix the limitations around the `update.php` function not working remotely either, correct?

dsbaars:
I'm also working on a fix, couldn't find this issue because I was looking for the term "external" in the pull request title. I also improved `update.php`. Still need to run the tests but you can check it out at https://github.com/dsbaars/Nominatim/tree/external-pgsql-fix

otbutz:
Would it be possible to recreate the functionality of `nominatim.so` with pure PL/pgSQL? This would solve the issue once and for all.

lonvia:
There is a PL/pgSQL version floating somewhere around the internet. Unfortunately I can't seem to find it anymore. If you can dig it up and update it so that it works with the current version, I'd be willing to take a PR that offers it as an alternative. Be aware though that there is a performance penalty to be paid when using PL/pgSQL.

karambolishe:
Where I should specify password for connect to remote database (aws rds)?
I changed settings file (settings/settings.php) according to tips and finally I have this example:

> @define('CONST_Database_DSN', 'pgsql://nominatim:nominatim@Endpoint_to_rds_database.rds.amazonaws.com:5432/nominatim'); 

I run this command:
`./utils/setup.php --osm-file ../geofabrik/algeria-latest.osm.pbf --all --osm2pgsql-cache 28000`

and during every 'SET' to database I should enter password... but in the end I encounter the error. 
I guess that application doesn't know what password it should use.

```
WARNING: resetting threads to 1
Create DB
Password: 
Setup DB
Postgres version found: 9.6
Password for user nominatim: 
CREATE EXTENSION
Password for user nominatim: 
CREATE EXTENSION
Postgis version found: 2.3
Password for user nominatim: 
SET
SET
SET
SET
SET
SET
SET
SET
CREATE TABLE
CREATE INDEX
REVOKE
Password for user nominatim: 
SET
SET
SET
SET
SET
SET
SET
SET
SET
CREATE TABLE
CREATE INDEX
CREATE INDEX
Password for user nominatim: 
SET
SET
SET
SET
SET
SET
SET
SET
SET
CREATE TABLE
CREATE INDEX
Password for user nominatim: 
SET
SET
SET
SET
SET
SET
SET
SET
SET
CREATE TABLE
Password for user nominatim: 
SET
SET
SET
SET
SET
SET
SET
Password for user nominatim: 
SET
SET
SET
SET
SET
SET
SET
SET
SET
CREATE TABLE
Password for user nominatim: 
CREATE TABLE
Password for user nominatim: 
CREATE TYPE
Import
osm2pgsql version 0.93.0-dev (64 bit id space)

Using projection SRS 4326 (Latlong)
Connection to database failed: fe_sendauth: no password supplied
```
If some one knows how it fix, help please.
Thanks in advance

mtmail:
@karambolishe Nominatim requires a custom extension (the compiled `nominatim.so`) to exist on the remote database server. With Amazon RDS you have no way to copy/install that. See https://github.com/openstreetmap/Nominatim/pull/112 Even if the password is set (not sure why your example doesn't work) the installation would fail at a later point.

karambolishe:
Thank you for your answer, at least I'm sure that sample with rds isn't working.

PR #553, opened 2016-10-17
contribution guidelines mention code style checks

mtmail:


lonvia:
I've done a few minor commits without PR to fix a few urgent bugs on the osm.org instance and this was actually one of the things that already went in. Sorry about that. Looks like you are on an older version of Nominatim.

PR #554, opened 2016-10-17
fix PHP warnings when reverse search was called without parameters or…

mtmail:
… empty results

This fixes 7 warnings for https://github.com/twain47/Nominatim/issues/547. I also checked search.php and details.php. Remember to copy `website/*.php` into `build/website/` when testing.

### DIFF
--- a/lib/template/address-html.php
+++ b/lib/template/address-html.php
@@ -13,12 +13,12 @@
     <form class="form-inline" role="search" accept-charset="UTF-8" action="<?php echo CONST_Website_BaseURL; ?>reverse.php">
         <div class="form-group">
             <input name="format" type="hidden" value="html">
-            <input name="lat" type="text" class="form-control input-sm" placeholder="latitude"  value="<?php echo htmlspecialchars($_GET['lat']); ?>" >
-            <input name="lon" type="text" class="form-control input-sm" placeholder="longitude" value="<?php echo htmlspecialchars($_GET['lon']); ?>" >
+            <input name="lat" type="text" class="form-control input-sm" placeholder="latitude"  value="<?php echo htmlspecialchars(isset($_GET['lat']) ? $_GET['lat'] : ''); ?>" >
+            <input name="lon" type="text" class="form-control input-sm" placeholder="longitude" value="<?php echo htmlspecialchars(isset($_GET['lon']) ? $_GET['lon'] : ''); ?>" >
             max zoom
 
-            <select name="zoom" class="form-control input-sm" value="<?php echo htmlspecialchars($_GET['zoom']); ?>">
-                <option value="" <?php echo $_GET['zoom']==''?'selected':'' ?> >--</option>
+            <select name="zoom" class="form-control input-sm" value="<?php echo htmlspecialchars(isset($_GET['zoom']) ? $_GET['zoom'] : ''); ?>">
+                <option value="" <?php echo (isset($_GET['zoom']) ? $_GET['zoom'] : '') == '' ? 'selected' : '' ?> >--</option>
                 <?php
 
                     $aZoomLevels = array(
@@ -65,7 +65,7 @@
 
     <div id="content">
 
-<?php if ($aPlace) { ?>
+<?php if (isset($aPlace)) { ?>
 
         <div id="searchresults" class="sidebar">
         <?php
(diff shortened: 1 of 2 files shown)


lib/template/address-html.php
```diff
@@ -13,12 +13,12 @@
...
-            <input name="lat" type="text" class="form-control input-sm" placeholder="latitude"  value="<?php echo htmlspecialchars($_GET['lat']); ?>" >
-            <input name="lon" type="text" class="form-control input-sm" placeholder="longitude" value="<?php echo htmlspecialchars($_GET['lon']); ?>" >
+            <input name="lat" type="text" class="form-control input-sm" placeholder="latitude"  value="<?php echo htmlspecialchars(isset($_GET['lat']) ? $_GET['lat'] : ''); ?>" >
+            <input name="lon" type="text" class="form-control input-sm" placeholder="longitude" value="<?php echo htmlspecialchars(isset($_GET['lon']) ? $_GET['lon'] : ''); ?>" >
             max zoom
 
-            <select name="zoom" class="form-control input-sm" value="<?php echo htmlspecialchars($_GET['zoom']); ?>">
-                <option value="" <?php echo $_GET['zoom']==''?'selected':'' ?> >--</option>
+            <select name="zoom" class="form-control input-sm" value="<?php echo htmlspecialchars(isset($_GET['zoom']) ? $_GET['zoom'] : ''); ?>">
+                <option value="" <?php echo (isset($_GET['zoom']) ? $_GET['zoom'] : '') == '' ? 'selected' : '' ?> >--</option>
```
lonvia: Simpler:  if (isset($_GET['zoom']) && $_GET['zoom']) echo 'selected'

Similar above where the empty string is one of the choices.

lib/template/address-html.php
```diff
@@ -13,12 +13,12 @@
     <form class="form-inline" role="search" accept-charset="UTF-8" action="<?php echo CONST_Website_BaseURL; ?>reverse.php">
         <div class="form-group">
             <input name="format" type="hidden" value="html">
-            <input name="lat" type="text" class="form-control input-sm" placeholder="latitude"  value="<?php echo htmlspecialchars($_GET['lat']); ?>" >
-            <input name="lon" type="text" class="form-control input-sm" placeholder="longitude" value="<?php echo htmlspecialchars($_GET['lon']); ?>" >
+            <input name="lat" type="text" class="form-control input-sm" placeholder="latitude"  value="<?php echo htmlspecialchars(isset($_GET['lat']) ? $_GET['lat'] : ''); ?>" >
+            <input name="lon" type="text" class="form-control input-sm" placeholder="longitude" value="<?php echo htmlspecialchars(isset($_GET['lon']) ? $_GET['lon'] : ''); ?>" >
```
lonvia: Just a thought: wouldn't it be easier to use $fLat and $fLon here or will that get us into trouble with float-to-string conversion?

mtmail:
> wouldn't it be easier to use $fLat and $fLon here 

Those contain the search result which might be different from the coordinates the user searched for. When searching for specific coordinates it will be confusing why the value in the searchboxes change after submitting.

mtmail:
@lonvia Can you check if this is ready to merge?

mtmail:
Now made sure $iZoom and $aPlaces are guaranteed to set in reverse.php so the template needs less checks.

lib/template/address-html.php
```diff
@@ -13,12 +13,12 @@
...
-            <input name="lat" type="text" class="form-control input-sm" placeholder="latitude"  value="<?php echo htmlspecialchars($_GET['lat']); ?>" >
-            <input name="lon" type="text" class="form-control input-sm" placeholder="longitude" value="<?php echo htmlspecialchars($_GET['lon']); ?>" >
+            <input name="lat" type="text" class="form-control input-sm" placeholder="latitude" value="<?php echo $fLat; ?>" >
+            <input name="lon" type="text" class="form-control input-sm" placeholder="longitude" value="<?php echo $fLon; ?>" >
             max zoom
 
-            <select name="zoom" class="form-control input-sm" value="<?php echo htmlspecialchars($_GET['zoom']); ?>">
-                <option value="" <?php echo $_GET['zoom']==''?'selected':'' ?> >--</option>
+            <select name="zoom" class="form-control input-sm" value="<?php echo $iZoom; ?>">
+                <option value="" <?php echo ($iZoom === 0) && 'selected' ?> >--</option>
```
lonvia: $iZoom = 0 is a valid zoom level

Also, please change code to: <?php if ($iZoom === 0) echo 'selected'; ?>

mtmail:
Rebased with master and changed $iZoom check when $iZoom is unset.

lonvia:
Something odd happened with those commits. I've squashed and rebased everything and commited as b27926fd4d96e0bcb0471543e4f29c14752f43e2.

PR #556, opened 2016-10-17
Drop support for PHP < 5.4

lonvia:
PHP 5.3 has been out of production for two years now, so it should be safe to drop support for it. PHP 5.4 and 5.5 are actually end-of-life as well but CentOS 7 still ships with PHP 5.4. So we are going to be stuck with it for a while.


mtmail:
Looks good. Travis-CI currently fails, but looks related to changes to postgresql, not PHP.

PR #570, opened 2016-10-30
Better parent for interpolated housenumbers

mtmail:
With all the SQL cleanup it's worth running the full production test suite as well. Locally I used `git diff --word-diff=color -b` to doublecheck that no logic changed.


lonvia:
Merged with minor style fixes.

PR #583, opened 2016-11-23
pull the increment of search diameter to bottom while structure

jandersonmartins:
Move the increment to bottom of while because search diameter is set to 0.0004 but never this value is never used.


lonvia:
The first increment is already taken into account when setting the initial value, i.e. diameter is meat to start at 0.0008. Moving the instruction to the end of the loop has unintended side-effects for the max rank in effect.

PR #592, opened 2016-12-15
Tell users the website will be available at /search.php, not /

mtmail:
Not the first time it got users (including myself) confused, see https://github.com/twain47/Nominatim/issues/591


lonvia:
I would prefer to add the DefaultIndex directive. I have no idea how that translates to ngix though.

mtmail:
sure, I'll create a new PR.

PR #593, opened 2016-12-15
housenumber interpolation: force calculated housenumber into range

mtmail:
We collected 50+ unique failures due to housenumber interpolation in reverse search when using the master branch.

Examples of affected coordinates
```
http://localhost:80/nominatim/reverse?format=jsonv2&lat=32.61393&lon=-16.98925
http://localhost:80/nominatim/reverse?format=jsonv2&lat=-23.5445217&lon=-46.5939133
http://localhost:80/nominatim/reverse?format=jsonv2&lat=-27.379749&lon=153.068848
http://localhost:80/nominatim/reverse?format=jsonv2&lat=-33.53406&lon=-70.7664483
http://localhost:80/nominatim/reverse?format=jsonv2&lat=-36.626919&lon=-71.8396
http://localhost:80/nominatim/reverse?format=jsonv2&lat=-5.785943&lon=-35.1996397
http://localhost:80/nominatim/reverse?format=jsonv2&lat=23.015966&lon=120.203583
http://localhost:80/nominatim/reverse?format=jsonv2&lat=48.907551&lon=2.144232
http://localhost:80/nominatim/reverse?format=jsonv2&lat=49.5786024&lon=-123.2247476
http://localhost:80/nominatim/reverse?format=jsonv2&lat=53.6576205&lon=-113.6640007
http://localhost:80/nominatim/reverse?format=jsonv2&lat=55.15658&lon=-118.8447
http://localhost:80/nominatim/reverse?format=jsonv2&lat=55.145117&lon=-118.795183
```

Those produce HTTP error 500. They work fine on nominatim.openstreetmap.org.

`
[...]

PR #605, opened 2017-01-09
Better search radius for interpolations when reverse geocoding

### LINKED ISSUE #545: SQL error on reverse geocoding of interpolated address
Occasionally reverse geocoding fails with the following message in the postgres error logs:

```
ERROR:  line_interpolate_point: 2nd arg isn't within [0,1]
STATEMENT:  select place_id, partition, 'W' as osm_type, osm_id, 'place' as class, 'house' as type, null admin_level, housenumber, null as street, null as isin, postcode, calculated_country_code as country_code, parent_place_id, null as linked_place_id, 30 as rank_address, 30 as rank_search, (0.75-(30::float/40)) as importance, null as indexed_status, null as indexed_date, null as wikipedia, calculated_country_code,  get_address_by_language(place_id, housenumber, ARRAY['short_name','name','brand','official_name','ref','type']) as langaddress, null as placename, null as ref, ST_X(point) as lon, ST_Y(point) as lat from (select *, ST_LineInterpolatePoint(linegeo, (housenumber-startnumber::float)/(endnumber-startnumber)::float) as point from  (select *,  CASE WHEN interpolationtype='odd' THEN floor((1*(endnumber-startnumber)+startnumber)/2)::int*2+1 WHEN interpolationtype='even' THEN ((1*(endnumber-startnumber)+startnumber)/2)::int*2 WHEN interpolationtype='all' THEN (1*(endnumber-startnumber)+startnumber)::int END as housenumber from location_property_osmline where place_id = 11679134) as blub1) as blub2
```

The problematic interpolation line is starting at this node: https://www.openstreetmap.org/node/1202725003 Note that the interpolation is odd but has an even endpoint. The error happens when the computed fraction is 1.

Need to write test case.

lonvia:
Investigating #545, it turns out that the underlying cause is that reverse uses a larger radius when looking for interpolations than when looking for simple house numbers. So an interpolation may still be found even when the house numbers for the start and end points are outside the search radius. In that case, the end point of the interpolation may well be returned as an result (leading to fraction 1) while later the code assumes that this cannot be the case.

This PR fixes this by searching for interpolations only when $maxRank is larger than 26 (i.e. we are still looking for POIs).

@mtmail I've run the examples #593 and they work all fine with this PR. Would you mind quickly checking also the other 50 cases you have collected?

Fixes #545, supersedes #593.

### DIFF
--- a/lib/ReverseGeocode.php
+++ b/lib/ReverseGeocode.php
@@ -42,6 +42,29 @@ class ReverseGeocode
         $this->iMaxRank = (isset($iZoom) && isset($aZoomRank[$iZoom]))?$aZoomRank[$iZoom]:28;
     }
 
+    /**
+     * Find the closest interpolation with the given search diameter.
+     *
+     * @param string $sPointSQL   Reverse geocoding point as SQL
+     * @param float  $fSearchDiam Search diameter
+     *
+     * @return Record of the interpolation or null.
+     */
+    protected function lookupInterpolation($sPointSQL, $fSearchDiam)
+    {
+        $sSQL = 'SELECT place_id, parent_place_id, 30 as rank_search,';
+        $sSQL .= ' ST_LineLocatePoint(linegeo,'.$sPointSQL.') as fraction';
+        $sSQL .= ' , ST_Distance(linegeo,'.$sPointSQL.') as distance';
+        $sSQL .= ' FROM location_property_osmline';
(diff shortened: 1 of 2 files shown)


mtmail:
@lonvia  I still see the occasional failure, but will collect and investigate more examples first (and come up with a fix). Example
http://nominatim.openstreetmap.org/reverse.php?format=json&lat=-33.4885467&lon=-70.629193

lonvia:
Ok, thanks for the link. I've reopened #545 for tracking this.

PR #613, opened 2017-01-21
Replace lettuce with behave in tests. Use python3 Split import and (p…

mtmail:
…ython,php) tests into two travis executions

* Ubuntu16 inside Vagrant started displaying an interactive grub choice again. I found another fix and running it before 'apt-get update' seems also necessary

* Changed Ubuntu16 and Travis scripts to install python3 and behave according to https://github.com/twain47/Nominatim/pull/600. Travis uses Ubuntu 14 so it's slightly different

* Set Travis to use Python 3.6. It comes pre-install making the setup faster

* Set Travis to use Postgresql 9.6. Many attempts to use 9.3 failed because the `libpg.so` is 9.6. and cmake uses that.

* Split test runs into 'tests' (python, php) and 'monaco' (import of a small country). Travis-CI.org runs them in parallel on separate VMs and that almost halves the run time (to 7-10 minutes).
But it was also necessary: The Postgres data directory is set to a 786MB ramdisc and running 'tests' and 'monaco' on the same VM fails. The 'tests' seem to write 300MB to disc.

* My earlier attempts to get the api tests running probably failed because of the ramdisc. The Postgresql server crashes during the osm2pgsql step.

* Running against the test-db might work in the future. Travis doesn't seem to have a time limit after all (as long as a script writes something to STDOUT every 10 minutes). Limit might be RAM and the ramdisc of course.

* Running against multiple PHP versions (5.4, 5.6. 7.0) is also something to
look into later.

* Running against multiple Postgresql version will be trickier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,30 @@
 ---
-language:
-  - 'python'
 sudo: required
 dist: trusty
-before_install:
-  - git submodule update --init --recursive
+language: python
+python:
+  - "3.6"
+addons:
+  postgresql: "9.6"
+git:
+  depth: 3
+env:
+  - TEST_SUITE=tests
+  - TEST_SUITE=monaco
 install:
   - vagrant/install-on-travis-ci.sh
 before_script:
-  - cd $TRAVIS_BUILD_DIR/build
-  - wget --no-verbose --output-document=../data/monaco.osm.pbf http://download.geofabrik.de/europe/monaco-latest.osm.pbf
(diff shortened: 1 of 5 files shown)


lonvia:
> The 'tests' seem to write 300MB to disc.

Just checked that. The test template DB comes at 98MB. That one is copied for the actual test and some data is added on top. The size if mostly due to the country_osm_grid table, I don't think there is much we can do about that.

PR #622, opened 2017-02-04
Fix for Russel Way in TIGER 2016 data (v2)

mtmail:
This PR replaces https://github.com/twain47/Nominatim/pull/594.

The patch logic is no longer hidden inside the import script. Since it's a separate step and separate script it can be extended should we find more errors. Hopefully in 6 months the Tiger 2017 data has it fixed anyway.

### DIFF
--- a/utils/imports.php
+++ b/utils/imports.php
@@ -12,7 +12,8 @@ $aCMDOptions
     array('quiet', 'q', 0, 1, 0, 0, 'bool', 'Quiet output'),
     array('verbose', 'v', 0, 1, 0, 0, 'bool', 'Verbose output'),
 
-    array('parse-tiger', '', 0, 1, 1, 1, 'realpath', 'Convert tiger edge files to nominatim sql import - datafiles from 2011 or later (source: edges directory of tiger data)'),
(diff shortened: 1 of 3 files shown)


utils/imports.php
```diff
@@ -12,15 +12,16 @@
     array('quiet', 'q', 0, 1, 0, 0, 'bool', 'Quiet output'),
     array('verbose', 'v', 0, 1, 0, 0, 'bool', 'Verbose output'),
 
-    array('parse-tiger', '', 0, 1, 1, 1, 'realpath', 'Convert tiger edge files to nominatim sql import - datafiles from 2011 or later (source: edges directory of tiger data)'),
+    array('parse-tiger', '', 0, 1, 1, 1, 'realpath', 'Convert Tiger edge files to nominatim sql import - datafiles from 2011 or later (source: edges directory of tiger data)'),
+    array('patch-tiger', '', 0, 1, 0, 0, 'bool', 'Fix converted Tiger data files')
```
lonvia: Switch needs a different name.

lonvia:
Closing as this is heavily conflicting with the latest changes on master and Russel Way has been fixed in the 2017 Tiger data anyway as far as I can see.

PR #624, opened 2017-02-05
PHP tests for ParameterParser

mtmail:
Adds PHP tests for ParameterParser and Geocode.

We can't use the `userError()` logic from `init-website.php` or `init-cmd.php` because in tests they don't make sense. I opted for throwing an exception and catching that. I can also understand if user errors should produce warnings only. It tests we should be safe with exceptions.

These parsing results are questionable, but I didn't want to change any logic:
* `getInt()` returns `false` for the input `0`
* `getFloat()` returns `false` for the input `0`
* `getString()` returns `false` for the input `0`
* `getSet()` can return a default value that's not an item in the set
* `getStringList()` returns null items for each empty string. E.g. `a,,b,c,` parses to an array with five items, two of them null. Or `,` parses to an array with two empty values


test/php/Nominatim/GeocodeTest.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace Nominatim;
+
+require_once(dirname(dirname(__FILE__)).'/../../build/settings/settings.php');
```
lonvia: We can't hard-code the name of the build directory as it may be called anything. If you need to reference to the build directory, add an environment variable where it can be changed.

The question is though, if you really need to refer to specific settings. I'd expect the unit tests to run independently from the current configuration. In that case it should be enough to include settings/default and add some variables you need for the test harness.

test/php/Nominatim/GeocodeTest.php
```diff
@@ -0,0 +1,52 @@
...
+        // values are cut to min and max limits
+        $oGeocode->setViewbox([-190, -90, 190, 90]);
+        $this->assertEquals('-180,90,180,-90', $oGeocode->getViewBoxString());
+
+
+        // boxes with width or height of 0 cause exceptions, the
+        // ViewBoxString still gets set
+        try {
+            $oGeocode->setViewbox([10, 10, 10, 10]);
+            $this->assertEquals('10,10,10,10', $oGeocode->getViewBoxString());
```
lonvia: never executed because of the exception

test/php/Nominatim/GeocodeTest.php
```diff
@@ -0,0 +1,52 @@
...
+        } catch (\Exception $e) {
+            $this->assertEquals($e->getMessage(), "Bad parameter 'viewbox'. Not a box.");
```
lonvia: This will not catch when the exception is _not_ thrown. Doesn't have PHPUnit something like 'assertException'?

test/php/Nominatim/ParameterParserTest.php
```diff
@@ -0,0 +1,174 @@
...
+
+
+    protected function setUp()
+    {
+    }
+
+
+    public function testGetBool()
+    {
+        $oParams = new ParameterParser(['bool1' => '1', 'bool2' => '0', 'bool3' => 'true', 'bool4' => 'false', 'bool5' => '']);
```
lonvia: Please break lines at 80 characters.

test/php/Nominatim/GeocodeTest.php
```diff
@@ -0,0 +1,52 @@
...
+
+    protected function setUp()
+    {
+        $this->oDB =& getDB();
+    }
+
+
+    public function testViewbox()
+    {
+        $oGeocode = new Geocode($oDB);
```
lonvia: Do you mean `$this->oDB`? I think you will get away with setting a nullptr (or false) here, if not we really need a mock object, so that the tests don't create a DB connection.

lonvia:
> These parsing results are questionable, but I didn't want to change any logic:
> 
>     getInt() returns false for the input 0
>     getFloat() returns false for the input 0
>     getString() returns false for the input 0
>     getSet() can return a default value that's not an item in the set
>     getStringList() returns null items for each empty string. E.g. a,,b,c, parses to an array with five items, two of them null. Or , parses to an array with two empty values

The `getSet()` I would consider 'shooting your own foot', not an error. The rest sounds really wrong. I lean towards adding failing tests for these cases, so we don't forget that there is still something to fix.

mtmail:
In PHPUnit the `expectException`-family of methods got added in version 5.2. On Ubuntu 16 with `apt-get install phpunit` you get version 5.1.3. On Travis-CI (Ubuntu 14) it's version 3.7. I haven't checked CentOS yes.

Switching to `pear install phpunit` in our installation instructions isn't straight-forward either. That works for PHP5.6, but not PHP7. I need to do more testing.

Using a DB mock object is the best approach. Until then I'll set the null pointer.

For the logic changes I'll file a separate issue/PR.

lonvia:
if exceptions tests are not available, I lean toward setting a global variable in chksql() and do an expect on it after calling the function to be tested.

lonvia:
Closing in favour of #941.

PR #633, opened 2017-02-15
Programs and paths

melvyn-sopacua:
Package maintainers can set WWW_APP and WWW_PUBLIC by adding the following to Cmake's invocation:
```
-DWWW_PUBLIC:PATH=/usr/local/www/nominatim -DWWW_APP:PATH=/usr/local/share/nominatim
```
Provided defaults are for Linux. The above I'll use for FreeBSD's hierarchy.
Additionally, this will use osm2pgsql from system if installed and finds Osmosis and correctly updates path in settings.


settings/defaults.php
```diff
@@ -1,6 +1,6 @@
 <?php
-@define('CONST_BasePath', '@CMAKE_SOURCE_DIR@');
-@define('CONST_InstallPath', '@CMAKE_BINARY_DIR@');
+@define('CONST_BasePath', '@WWW_APP@');
+@define('CONST_InstallPath', '@WWW_PUBLIC@');
```
lonvia: I'm not sure, what you want to achieve here. BasePath and InstallPath point to the directory where to find data, php scripts and stuff. So you can't set them to some global path without installing all these files. Also note that they are currently set up so that Nominatim can be executed directly from the build directory. This is something that needs to remain possible for the development process. So we'd really need a separate 'make install' here, which creates its own settings suitable for the installed version.

CMakeLists.txt
```diff
@@ -34,7 +34,24 @@ add_definitions(-DNOMINATIM_VERSION="${NOMINATIM_VERSION}")
...
+        add_subdirectory(osm2pgsql)
+	message(STATUS "Using local osm2pgsql")
+	set(OSM2PGSQL_PATH ${CMAKE_INSTALL_PREFIX}/osm2pgsql/osm2pgsql)
+else()
+	set(OSM2PGSQL_PATH ${OSM2PGSQL})
+	message(STATUS "Using osm2pgsql at ${OSM2PGSQL}")
+endif()
+find_program(OSMOSIS_PATH osmosis)
+if (NOT EXISTS "${OSMOSIS_PATH}")
+        message(FATAL_ERROR "Osmosis is required")
```
lonvia: osmosis is optional. It is only required when running updates.

CMakeLists.txt
```diff
@@ -34,7 +34,24 @@ add_definitions(-DNOMINATIM_VERSION="${NOMINATIM_VERSION}")
 
 set(BUILD_TESTS off CACHE BOOL "Build test suite" FORCE)
 set(WITH_LUA off CACHE BOOL "Build with lua support" FORCE)
-add_subdirectory(osm2pgsql)
+set(WWW_PUBLIC /var/www/nominatim CACHE PATH "Public www root")
+set(WWW_APP ${CMAKE_INSTALL_PREFIX}/share/nominatim CACHE PATH "Web application root")
+find_program(OSM2PGSQL osm2pgsql DOC "System installed osm2pgsql")
```
lonvia: This is not a good idea. Nominatim is bound to a very specific version of osm2pgsql. Running it with any other than the one that comes with the package is not supported. I'm aware that this is a problem for packagers. How about renaming the osm2pgsql built by Nominatim into nominatim-osm2pgsql, so that it can exist in parallel with another installed version.

settings/defaults.php
melvyn-sopacua: No worries about the negative, I can take it. Sometimes the best way to get something done is to start doing it ;).

I looked at the difference between install and base and figured that Install Path is in the webroot and base path are included files (classes and so on). I may have that backwards, but maybe renaming the constants to reflect what they are semantically (not where they reside) is a good idea to begin with.
Supporting both installed and "staged" would be hard to do if you want to keep it in one file. It's probably better create two versions  of settings.php. One that works in-place as is and one that is going to be installed.

CMakeLists.txt
melvyn-sopacua: Gotcha.
And yes, a different name would work. The date seemed to coincide with latest release, but now that I'm looking at hashes, they are indeed different.

CMakeLists.txt
melvyn-sopacua: Changed.
If you could open an issue with requirements for installation I'm happy to work on it. Starting a project that will use Nominatim to feed Komoot for Netherlands region and I'd like to start off with upcoming release, rather then previous, so I have some motivation to get it in shape :).

lonvia:
Can you split off the osmosis change? That looks good to go. For the rest I've opened #634 and I would suggest to continue the discussion over there.

lonvia:
osmosis is no longer relevant. So closing now.

PR #635, opened 2017-02-17
Set osmosis if available

melvyn-sopacua:
Factored out as requested.


lonvia:
Committed with tiny syntax fix as 66d7f1a058ece50f7e1fca4c4771b164cbb76dd1.

PR #649, opened 2017-02-28
Extra headers for compatibility on IE>8

ldgeo:
I encountered some problems on IE11 with compatibility mode actived (compat with IE7), the search page does not work at all.
We need some extra meta tags to work on all IE versions.


mtmail:
Thanks!

PR #660, opened 2017-03-12
github.com/twain47 => github.com/openstreetmap

mtmail:
Github will redirect the old URL almost indefinitely so this is low priority cleanup only.


lonvia:
Awesome, thanks for that.

PR #671, opened 2017-03-19
Html view use geojson geometries

mtmail:
* removed a whole folder of styling that was used by openlayers in the past
* update leaflet.js 0.7.7 => 1.0.3, no breaking API changes but probably a year of fixes. (for example a 3 year old fix of mine that was scheduled for 0.8 release https://github.com/Leaflet/Leaflet/pull/2527)
* search and details for format=HTML return geometries in geojson instead of wkn
* leaflet 1.0 can add geojson to the map directly, no longer need the omnivore plugin


lonvia:
There is a very strange effect on the search map. When a geometry is shown and you zoom in or out of the map (for example, using the + and - buttons in the upper left corner) then the geometry moves to the upper left (zoom in) or lower right (zoom out) corner before going back to its original location when the new zoom level is reached. Do you see something similar? Does my description even make sense?

mtmail:
ok, found the issue (https://github.com/Leaflet/Leaflet/issues/5227) and updated the CSS file

lonvia:
Much better, thanks.

PR #675, opened 2017-03-20
git ignore .DS_Store files (MacOS)

mtmail:
Typical/annoying MacOS temporary file https://en.wikipedia.org/wiki/.DS_Store


lonvia:
This is not the right place for generic files to ignore. You should add it to the .gitignore file in your home directory.

PR #676, opened 2017-03-20
On /search.php the map-position box disappeared behind the map tiles

mtmail:
After the new leaflet 1.0 CSS got applied the map-position box disappeared behind the map tiles after clicking the 'show map bounds' button.

P.S. what do you think about adding a link to http://tools.geofabrik.de/osmi/?view=addresses&lon=-1.98369&lat=43.32181&zoom=18 for debugging as well?


lonvia:
An OSMI link is worth a try but it should be obvious that this is an external site with potential differences in their matching of addresses.

PR #690, opened 2017-03-26
fix table/sequence ... does not exist

ImreSamu:
This is fixing 4 SQL warning/errors: 
* `ERROR:  sequence "seq_postcodes" does not exist`
* `ERROR:  table "import_polygon_error" does not exist`
* `ERROR:  table "import_polygon_delete" does not exist`
* `ERROR:  sequence "file" does not exist`


lonvia:
Thanks!

PR #703, opened 2017-04-18
DB column alignment and size

roques:
PostgreSQL aligns column data (not just in RAM) based on the host CPU's alignment requirements / preferences. Most relevant for Nominatim is that `bigint`s are 8 byte aligned. Swapping the order of `osm_type` and `osm_id` columns consistently saves 6 bytes per row. Reducing the size of the ranks, `partition`, `index_status`, and `admin_level` from `integer` to `smallint` and re-ordering columns based on their alignment is simple and saves quite a bit of space.
To be able to change the order of columns, I had to change some SQL-statements to explicitly mention the columns INSERTed instead of relying on tables having a particular column order. There were surprisingly few such cases.

I currently need Nominatim for Germany, Switzerland, and Italy. The difference in table sizes of `place` and `placex` for freshly imported, indexed, VACUUM FULLed data before and after the changes seem worthwhile. This **should** also result in a (probably hard to meassure) speedup due to a smaller cache foot-print.

| table | original | reordered |
| ----- | -------- | --------- |
| place | 8522 MB | 8310 MB |
| placex | 16693 MB | 15519 MB |

This branch also contains several other trivial changes, which I can separate into other PRs if necessary. You can also just cherry-pick them.

Further space-savings seem possible, but will require more effort:
 * Use a `"char"` for `indexed_status` with values `'N'`, `'R'`, `'D'` and `NULL`, requiring no space for fully indexed places.
 * Use the [tinyint](https://pgxn.org/dist/tinyint/) extension instead of `smallint`.
 * Encode `osm_type` into the most significant byte of `osm_id`. — Especially interesting for indexes, which currently are still padded from 9 to 16 bytes.
 * Split the current `name HSTORE` into `name text, ref text, alt_names HSTORE`, so that the keys of `alt_names` are just the two character language code, and `alt_names` is `NULL` for most places. Don't store `alt_names` which are identical to `name`. — Will save at least 12 bytes for every named place.
 * Store geometries without SRID.
 * Determine if PostgreSQL ≥9.6 is consistently smart enough to use indices like `CREATE INDEX idx_place_classtype_shop_fish_centroid ON place USING gist (st_centroid(geometry)) WHERE class='shop' AND type='fish';` so that we could get rid of all the `place_classtype_*` tables.

### DIFF
--- a/sql/functions.sql
+++ b/sql/functions.sql
@@ -1246,7 +1246,8 @@ BEGIN
   END IF;
 
   -- Adding ourselves to the list simplifies address calculations later
-  INSERT INTO place_addressline VALUES (NEW.place_id, NEW.place_id, true, true, 0, NEW.rank_address); 
+  INSERT INTO place_addressline (place_id, address_place_id, fromarea, isaddress, distance, cached_rank_address)
+    VALUES (NEW.place_id, NEW.place_id, true, true, 0, NEW.rank_address); 
 
   -- What level are we searching from
   search_maxrank := NEW.rank_search;
@@ -1734,7 +1735,8 @@ BEGIN
       IF location.rank_search > 4 THEN
           nameaddress_vector := array_merge(nameaddress_vector, location.keywords::integer[]);
       END IF;
-      INSERT INTO place_addressline VALUES (NEW.place_id, location.place_id, true, location_isaddress, location.distance, location.rank_address);
+      INSERT INTO place_addressline (place_id, address_place_id, fromarea, isaddress, distance, cached_rank_address)
+        VALUES (NEW.place_id, location.place_id, true, location_isaddress, location.distance, location.rank_address);
 
       IF location_isaddress THEN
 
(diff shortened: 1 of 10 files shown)


roques:
Also, as osm2pgsql is in its own repo, once this has been merged, don't forget to cherry-pick the one commit on https://github.com/roques/osm2pgsql/tree/tetris to swap `osm_type` and `osm_id` in `place` and `placex`.

lonvia:
I'm not sure about the index changes. There might have been a reason that they are the way they are. In particular, the osm_id index might be used in places as an index over osm_type only.

Otherwise, looks all pretty good. I'd like to do a bit of testing though.

roques:
Re-thinking the change to the column order in `place_id_idx`: Having the much more specific column first would be nice in principle, but when the index is used with both `osm_type` and `osm_id` fixed, the order does not matter.
Switching the order unfortunately did not get rid of the padding in the index and thus did not reduce its size. So that change was not well substantiated.
I've pushed a fixup reverting the change.

I remain convinced that making `idx_placex_linked_place_id` a partial index with `WHERE linked_place_id IS NOT NULL` is the right way to go; In my currently loaded data, 99.74% of `placex` rows have `linked_place_id IS NULL` and thus it would be ludicrous for a query planner to use the index to find 99.74% of all rows.

lonvia:
I vaguely remember thinking about adding the NOT NULL restriction on that index. It might have been that I didn't do it because then the index wasn't used with that where clause. It is just as likely that I simply forgot about it before I got to implementing it.

lonvia:
I couldn't find any place where the index change place poses an issue, so merged now as is.

Many thanks for your contribution.

Issue #118, opened 2014-03-06
Make response compatible with geojson


sfkeller:
Ex ante: Nominatim is one of the most valuable and promising OSM services!

Nominatim response (parameter "polygon_geojson") suggests that the response is a valid GeoJSON format - but it's not currently.

Actually it would be very easy to make it valid with the following changes:
1. Replace "geojson" with "geometry".
2. Put "FeatureCollection" header in front.
3. For every feature, nest everything except geometry in attribute "properties".

See example below:
--S.

Example GET request: http://nominatim.openstreetmap.org/search?q=Oberseestrasse+10,+Rapperswil&format=jsonv2&polygon_geojson=1&addressdetails=1

Original Nominatim response:
    {
        "place_id": "44203959",
        "licence": "Data © OpenStreetMap contributors, ODbL 1.0. http://www.openstreetmap.org/copyright",
        "osm_type": "way",
        "osm_id": "34179470",
        "boundingbox": [
            "47.2229156494141",
            "47.2242317199707",
            "8.82080268859863",
            "8.82882499694824"
        ],
        "lat": "47.224015",
        "lon": "8.824787",
        "display_name": "Oberseestrasse, Rapperswil-Jona, Wahlkreis See-Gaster, St. Gallen, 8645, Schweiz",
        "place_rank": "26",
        "category": "highway",
        "type": "tertiary",
        "importance": 0.5,
        "address": {
            "road": "Oberseestrasse",
            "city": "Rapperswil-Jona",
            "county": "Wahlkreis See-Gaster",
            "state": "St. Gallen",
            "postcode": "8645",
            "country": "Schweiz",
            "country_code": "ch"
        },
        "geometry": {
            "type": "LineString",
            "coordinates": [
                [
                    8.8208036,
                    47.2241061
                ],

Header snippet of correct GeoJSON response:
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {
        "place_id": "44203959",
       ...

lonvia:
I'd be happy to take a pull request for another template implementing geojson but chances are small that I will find the time to implement this myself.

PR #707, opened 2017-04-20
Support to geojson compatible output #118

narigondelsiglo:
We have implemented geojson output support, but didn't write any test cases...
Don't know how to do it...
Hope it's helpful

### DIFF
--- a/lib/template/address-geojson.php
+++ b/lib/template/address-geojson.php
@@ -1,0 +1,91 @@
+<?php
+
+$aFilteredPlaces = array();
+
+if (!sizeof($aPlace))
+{
+    if (isset($sError))
+        $aFilteredPlaces['error'] = $sError;
+    else
+        $aFilteredPlaces['error'] = 'Unable to geocode';
+}
+else
+{
+    if ($aPlace['place_id']) $aFilteredPlaces['place_id'] = $aPlace['place_id'];
+    $aFilteredPlaces['licence'] = "Data © OpenStreetMap contributors, ODbL 1.0. http://www.openstreetmap.org/copyright";
+    $sOSMType = formatOSMType($aPlace['osm_type']);
+    if ($sOSMType)
+    {
+        $aFilteredPlaces['osm_type'] = $sOSMType;
+        $aFilteredPlaces['osm_id'] = $aPlace['osm_id'];
+    }
+    if (isset($aPlace['lat'])) $aFilteredPlaces['lat'] = $aPlace['lat'];
+    if (isset($aPlace['lon'])) $aFilteredPlaces['lon'] = $aPlace['lon'];
+
+    $aFilteredPlaces['place_rank'] = $aPlace['rank_search'];
+
+    $aFilteredPlaces['category'] = $aPlace['class'];
+    $aFilteredPlaces['type'] = $aPlace['type'];
+
+    $aFilteredPlaces['importance'] = $aPlace['importance'];
+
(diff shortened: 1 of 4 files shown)


lib/template/address-geojson.php
```diff
@@ -0,0 +1,70 @@
...
+    if (isset($aPlace['askml']))
+    {
+        $aFilteredPlaces['geokml'] = $aPlace['askml'];
+    }
+
+}
+
+javascript_renderData(array(
+    "type" => "Feature", "properties" => $aFilteredPlaces, "geometry" => array("type"=>"Point","coordinates"=>array($aFilteredPlaces['lon']*1,$aFilteredPlaces['lat']*1))
+));
```
lonvia: You need to render a different array here when there is an error, lat and lon will not be set.

lib/template/search-geojson.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
+$aFilteredPlaces = array();
+
+foreach($aSearchResults as $iResNum => $aPointDetails)
+{
+
+    $aPlace = array(
+            'place_id'=>$aPointDetails['place_id'],
+            'licence'=>"Data © OpenStreetMap contributors, ODbL 1.0. http://www.openstreetmap.org/copyright",
```
lonvia: I believe that GeoJSON allows to put that on the FeatureCollection? If so, that would be better than repeating it for every feature.

website/reverse.php
```diff
@@ -29,7 +29,8 @@
 $fThreshold = $oParams->getFloat('polygon_threshold', 0.0);
 
 // Format for output
-$sOutputFormat = $oParams->getSet('format', array('html', 'xml', 'json', 'jsonv2'), 'xml');
+$sOutputFormat = $oParams->getSet('format', array('html', 'xml', 'json', 'jsonv2', 'geojson'), 'xml');
+$bAsGeoJSON = $bAsGeoJSON || $sOutputFormat === 'geojson';
```
lonvia: This always downloads the polygon geometry. Can you make it like the other formats, where you just get the centroid point, if a polygon parameter is not explicitly given?

website/search.php
```diff
@@ -26,15 +26,15 @@
...
+$sOutputFormat = $oParams->getSet('format', array('html', 'xml', 'json', 'jsonv2', 'geojson'), 'html');
 
 // Show / use polygons
 if ($sOutputFormat == 'html') {
     $oGeocode->setIncludePolygonAsGeoJSON($oParams->getBool('polygon_geojson'));
     $bAsGeoJSON = false;
 } else {
     $bAsPoints = $oParams->getBool('polygon');
-    $bAsGeoJSON = $oParams->getBool('polygon_geojson');
+    $bAsGeoJSON = $oParams->getBool('polygon_geojson') || $sOutputFormat === 'geojson';
```
lonvia: same here

lib/template/address-geojson.php
narigondelsiglo: Right!

lib/template/search-geojson.php
narigondelsiglo: Totally agree

website/reverse.php
narigondelsiglo: Ok

website/search.php
narigondelsiglo: Ok

lonvia:
Closing in favour of #978.

PR #728, opened 2017-05-21
JSON pretty print option

big-o:
Allow an elasticsearch-style 'pretty' option for JSON API requests to make debugging easier.

### DIFF
--- a/lib/lib.php
+++ b/lib/lib.php
@@ -465,6 +465,10 @@ function javascript_renderData($xVal, $iOptions = 0)
 {
     if (defined('PHP_VERSION_ID') && PHP_VERSION_ID > 50400)
         $iOptions |= JSON_UNESCAPED_UNICODE;
+        if (isset($_GET['pretty']) && $_GET['pretty'] && strtolower($_GET['pretty']) != 'false') {
+            $iOptions |= JSON_PRETTY_PRINT;
+        }
+
     $jsonout = json_encode($xVal, $iOptions);
 
     if (! isset($_GET['json_callback'])) {
--- a/lib/lib.php
+++ b/lib/lib.php
@@ -465,7 +465,7 @@ function javascript_renderData($xVal, $iOptions = 0)
 {
     if (defined('PHP_VERSION_ID') && PHP_VERSION_ID > 50400)
         $iOptions |= JSON_UNESCAPED_UNICODE;
-        if (isset($_GET['pretty']) && $_GET['pretty'] && strtolower($_GET['pretty']) != 'false') {
+        if (isset($_GET['pretty']) && in_array(strtolower($_GET['pretty']), array('', '1', 'true'))) {
             $iOptions |= JSON_PRETTY_PRINT;
         }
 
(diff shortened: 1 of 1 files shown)


lib/lib.php
```diff
@@ -465,6 +465,10 @@ function javascript_renderData($xVal, $iOptions = 0)
 {
     if (defined('PHP_VERSION_ID') && PHP_VERSION_ID > 50400)
```
SrihariThalla: Enclose `JSON_PRETTY_PRINT` inside this `if`

big-o:
Fixed the rookie error on the if statement. I'm happy to add the change to 2.5.x as well but am hesitant to add anything to the docs on the wiki as they currently seem to be a messy combination of 2.5.x installation docs and master branch API docs. I think some more work is needed to fix this (sync the wiki docs with the git repo or move the docs into the git repo? Which branch is master even based off?) before adding to it.

lonvia:
I wonder, do we really need an option for this or should we simply pretty-print by default if PHP can do it? The main argument against pretty-print is probably size. Do you have an estimate how much larger the results are in the worst case because of pretty-printing?

mtmail:
I use `pretty=1` a lot when dealing with my own APIs. Very useful for posting URLs in emails. Still I'd make that an optional parameter.

Using https://stedolan.github.io/jq/ to test pretty printing on the command-line I see a worst case of about 3x.

```
## Berlin
curl -s 'http://nominatim.openstreetmap.org/search.php?q=berlin&format=jsonv2' | wc -c
    5237
curl -s 'http://nominatim.openstreetmap.org/search.php?q=berlin&format=jsonv2' | jq '.' | wc -c
    6329 (+20%)


## Berlin including geojson
curl -s 'http://nominatim.openstreetmap.org/search.php?q=berlin&polygon_geojson=1&format=jsonv2' | wc -c
  104474

curl -s 'http://nominatim.openstreetmap.org/search.php?q=berlin&polygon_geojson=1&format=jsonv2' | jq '.' | wc -c
  305160 (+190%)


## Canada including geojson
curl -s 'http://nominatim.openstreetmap.org/search.php?q=canada&polygon_geojson=1&format=jsonv2' | wc -c
  829032

curl -s 'http://nominatim.openstreetmap.org/search.php?q=canada&polygon_geojson=1&format=jsonv2' | jq '.' | wc -c
 2746900 (+230%)


## reverse, 1 result
curl -s 'http://nominatim.openstreetmap.org/reverse.php?lat=41.40139&lon=2.12870&format=jsonv2' | wc -c
     778

curl -s 'http://nominatim.openstreetmap.org/reverse.php?lat=41.40139&lon=2.12870&format=jsonv2' | jq '.' | wc -c
     929 (+20%)
```

lonvia:
Merged as bdd90e890b44b12a9d46cc3b80de56bfa12af8fa.

I've taken the liberty to remove support for 'pretty=' (without value) for sake of consistency with the other parameters, where we generally ignore that.

PR #730, opened 2017-05-23
Update the formatting in Vagrant markdown doc

SrihariThalla:
* Set proper indentations for a code block

PR #736, opened 2017-05-26
Fix indentation in Vagrant instructions

Zverik:
Because of bad indentation, half the file was formatted as code.


lonvia:
Duplicate of #730.

PR #767, opened 2017-07-06
Update VAGRANT.md

manzari:
Setup.php will not fail with below error anymore

> Error: you need to download the country_osm_grid first:
> wget -O /home/vagrant/Nominatim/data/country_osm_grid.sql.gz http://www.nominatim.org/data/country_grid.sql.gz

### DIFF
--- a/VAGRANT.md
+++ b/VAGRANT.md
@@ -47,13 +47,14 @@ is.
       # inside the virtual machine:
       mkdir data
       cd build
+      wget -O /home/vagrant/Nominatim/data/country_osm_grid.sql.gz http://www.nominatim.org/data/country_grid.sql.gz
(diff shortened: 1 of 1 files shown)


mtmail:
excellent, thanks!

lonvia:
Same question as before: the download of the country grid is done as part of the vagrant scripts [here](https://github.com/openstreetmap/Nominatim/blob/master/vagrant/Install-on-Centos-7.sh#L158) and [here](https://github.com/openstreetmap/Nominatim/blob/master/vagrant/Install-on-Ubuntu-16.sh#L145). Does this not work for you? If it doesn't, please find out why. I'd prefer to fix whatever is not working with those scripts rather than adding duplicate instructions in VAGRANT.md.

lonvia:
Ah, I think I see the problem. It's changing into the wrong directory before the download.

PR #778, opened 2017-07-21
Mention explicitly that Osmosis is not required.

woodpeck:


lonvia:
Hmm, rather might want to change the name of that option.

Issue #650, opened 2017-03-01
No results when adding postal code to search query


meijdenmedia:
[No result](http://nominatim.openstreetmap.org/search.php?q=Jacob+Stolpstraat+4%2C+1462MV+Middenbeemster%2C+NL&addressdetails=1&format=json) vs [result](http://nominatim.openstreetmap.org/search.php?q=Jacob+Stolpstraat+4%2C+Middenbeemster%2C+NL&addressdetails=1&format=json)

The weird thing is that when you open 'result' you see the postcode matches the one given in the query string in the 'No result' link. The given street is currently in building process. For all other requests it's working perfectly.

[Result](http://nominatim.openstreetmap.org/search.php?q=Kalverstraat+3%2C+1012NX+Amsterdam%2C+NL&addressdetails=1&format=json) vs [result](http://nominatim.openstreetmap.org/search.php?q=Kalverstraat+3%2C+Amsterdam%2C+NL&addressdetails=1&format=json)

I hope somebody can help me out :)

lonvia:
The postcode you see in the result comes from the building. It normally should find it's way back into the search index of the street (which is required if you want to be able to find it) but it doesn't. I suspect that when the street was updated, it didn't merge the postcodes again. Needs a test to confirm.

PhilHa:
I've got a pretty similar problem:

Nominatim finds the municipality "Meisburg" ( [meisburg](http://nominatim.openstreetmap.org/search.php?q=meisburg&addressdetails=1&format=json) ) and also it's postal code "54570" ( [54570](http://nominatim.openstreetmap.org/search.php?q=54570&addressdetails=1&format=json) ).
But combining these two values it can't find anything ( [meisburg+54570](http://nominatim.openstreetmap.org/search.php?q=54570+meisburg&addressdetails=1&format=json) ).

But there is definitely a proper postal code boundary in osm ( [postalCode](http://www.openstreetmap.org/relation/1259858) ) and when adding any street to the queries of postal code ( [54570+Kirchstraße](http://nominatim.openstreetmap.org/search.php?q=54570+kirchstra%C3%9Fe&addressdetails=1&format=json) ) or the village name ( [Meisburg+Kirchstraße](http://nominatim.openstreetmap.org/search.php?q=meisburg+kirchstra%C3%9Fe&addressdetails=1&format=json) ) there are results with full information (village + postal code).

May there be a problem with the multipoygonal postal code and may there be a bug while checking if the villiage is within that area?

[...]

Issue #541, opened 2016-10-03
Import full Ordnance Survey CodePoint database for searching


gerv:
Many postcodes in the UK still don't produce a hit on Nominatim. Examples:
http://www.openstreetmap.org/search?query=LE11%201PN
http://www.openstreetmap.org/search?query=EN2%208LA

Either the CodePoint Open full postcode database hasn't been imported, or there's something wrong with the search. Discussion on the talk-gb mailing list concluded that the problem of not being able to find postcodes is best fixed in Nominatim - see this message and following:
https://lists.openstreetmap.org/pipermail/talk-gb/2016-September/019234.html

90% of the map searches I do are postcodes, and right now I have to use Google Maps for certainty. Can this long-standing OSM sadness be resolved? :-)

Gerv

lonvia:
Nominatim contains a copy of the CodePoint database but explicit postcode search via this database has never worked because nobody has ever implemented it. The CodePoint DB is only used to pinpoint the location of a search. Note that postcode search via the OSM database is limited as well because postcode centroids are only ever updated when the database is reimported (last one was only last week as it happend).

So here is the full todo list for getting this issue solved (ordered by difficulty, easiest first):

 * implement reverse lookup for coordinate search based on CodePoint coordinates when a UK postcode is detected
 * implement #403 
 * solve [normalization issue](https://trac.openstreetmap.org/ticket/4910)

From my side, #403 is higher on the TODO list because it solves postcode issues for the entire world. But if somebody from the UK community wants to tackle the CodePoint lookup that would be welcome.

Also related: #221

lonvia:
Fixed with #782.

Issue #537, opened 2016-09-28
No result if request contains param postcode


looser:
This is working:
- https://nominatim.openstreetmap.org/search?city=Panton+Hill&country=AU&street=Rodger+Rd&format=json
- https://nominatim.openstreetmap.org/search?city=Panton+Hill&country=AU&format=json

This is not working:
 -https://nominatim.openstreetmap.org/search?city=Panton+Hill&country=AU&street=Rodger+Rd&postalcode=3759&format=json
- https://nominatim.openstreetmap.org/search?city=Panton+Hill&country=AU&postalcode=3759&format=json

But the postal code is correct and openstreetmap.org is also able to find the address
http://www.openstreetmap.org/search?query=5783%20Panton%20Hill#map=14/-37.6414/145.2390

looser:
Any opinions on that?
I consider this as a bug

lonvia:
The postcode is not in OSM. Nominatim is not able to do a fuzzy match where it excludes part of the query. The only exception are house numbers. That's why postcode still seem to be excluded sometimes: Nominatim assumes they might be a house number. (And, yes, the information that it is a postcode is lost to Nominatim even when you use the postalcode=123 parameter. That's another construction site.)

Given that the format of postcodes is rather well defined on a country level, we'll might give it some special treatment in the future and also allow it to be dropped. But that requires to solve #403 first.

looser:
> And, yes, the information that it is a postcode is lost to Nominatim even when you use the postalcode=123 parameter

What do you exactly mean with "the information is lost" ? Should i never send a request which contains a postalcode as it is lost anyways?

lonvia:
When you add postalcode=123 to your query, Nominatim does normally not retain the information that 123 is a postcode. The number will be matched against anything including house numbers.

looser:
This is still a show stopper. 
I have not figured out one combination of params which is always working. I can not omit
the postalcode. There are a ton of Springfield in USA e.g. how am i supposed to get the right one without the postalcode?

In this case, the combination of country & postalcode is not working
https://nominatim.openstreetmap.org/search?country=AU&postalcode=3759&format=json

But country, city & street is working. 
https://nominatim.openstreetmap.org/search?city=Panton+Hill&country=AU&street=Rodger+Rd9&format=json

In this case country, city & postalcode is not working
https://nominatim.openstreetmap.org/search?city=Rendsburg&country=DE&format=json&postalcode=24768

but instead country & postalcode is working
[...]

Issue #403, opened 2016-03-16
move postcode areas into separate table


mtmail:
For postal code that share the same prefix we create additional places (`placex` table) during index time. Those places represented by a centroid only, not a polygon as one would expect.

For example for the UK postcode `SW1A 2AA` the postcode area is `SW1A`, for the German postcode `10013` it's `10013`.
https://developer.here.com/rest-apis/documentation/geocoder/topics/coverage-geocoder.html lists various country specific postcode formats. Special consideration should be given to outlier detections (e.g. typos).

The centroid is not recalculated when places in OSM are modified. There's also no process, e.g. script, to start recalculation or to delete a postcode area from Nominatim.

As a first step we want to move the postcode areas into a new table. Later we can write scripts to regularly update the areas. Or if that calculations turns out to very expensive offer downloads of pre-calculated areas.

Table schema (TBD):
- area postcode
- country
- centroid
- geometry

hridayns:
Hi mtmail, 

I have setup a development environment on Vagrant with Nominatim installed and the OSM data for Monaco downloaded. I’m trying to understand what exactly needs to be done to solve the issue. 

What I understand from your post - 
Problem:  The centroid for postal code areas is not recalculated everytime a person changes the bounding box area with iD or other editors.

Solution: To create a separate table with the columns ‘area postcode’,’ country’, ‘centroid’ and ‘geometry’ in which centroid is recalculated everytime someone makes a change to the bounding box area with iD or other editors.

What I need to clarify – 
1.	By places, do you mean the bounding box of a postal area?
2.	Does the centroid get calculated based on the bounding box? Or can a particular postal area have a centroid without a bounding box?
3.	The placex table stores centroid column data in a coded form. Is there some sort of method to convert this to a pair of coordinates which is then shown on the map?
4.	Even if downloads for pre-calculated areas are offered, wouldn’t the areas still need to be updated when someone changes the bounding box of a postal area?

Regards,
Hriday

lonvia:
Postcodes in OSM are not explicitly mapped but appear as part of addresses mapped for houses or POIs. So for each postcode, there are many point coordinates available. On import, Nominatim computes a centroid of all those points and adds the resulting point as an artifical object to the placex table (see utils/setup.php, section calculate-postcodes). This ticket talks about adding these points into a separate table instead.

joh-klein:
How does this relate to #746?

lonvia:
There is no relation to #746 at all.

[...]

PR #782, opened 2017-08-19
Rework handling of artificial postcode centroids

lonvia:
This PR changes postcode handling in three significant ways:                     
                                                                                 
 * Artificially computed postcode centroids have moved into their own table `location_postcode`.
 * The `postcode` column for places is now precomputed using postcodes from the object's addr:postcode tags, the  object's address and the artificial postcodes, providing a better postcode guess for some.
 * Postcodes now also get a special category in the `word` table (class=place/type=postcode) and will be handled special during search. It is in particular now possible to ignore the postcode.
                                                                                 
This fixes a number of outstanding issues:                                       
                                                                                 
* Bad or missing postcodes can now be ignored. (#161, #537, #650)                
* The postcode table becomes updatable without a database reimport. (#403)       
* Extra postcodes for the UK are now searchable. (#221, #541)                    
                                                                                 
Open points for later PRs:                                                       
* Although it is already possible to call `./utils/setup.php --calculate-postcodes` to update the artificial     postcodes, this function deletes and recreates all postcodes. A real update function would be nicer.
* Search results in the form of <city>, <postcode> now mostly work as well but they currently return a lot of    false positives which are not always correctly ordered. Fixing that requires some kind of subranking search      results.
* Artificial postcodes still interfere with postcode boundaries when searching for postcodes.
* Recognize postcodes by their patterns instead of looking them up in the `word` table.
                                                                                 
There is a path for forward migration of existing databases described in `docs/Migration.md`. However, it is     strongly recommended to reimport a new database as only a full reimport can completely leverage the new features.
                                                                                 
If somebody would like to test this PR for their country, feedback would be highly appreciated.


utils/setup.php
```diff
@@ -493,24 +493,59 @@
...
-    $sSQL .= "address->'postcode' as pc,";
-    $sSQL .= "avg(st_x(st_centroid(geometry))) as x,avg(st_y(st_centroid(geometry))) as y ";
-    $sSQL .= "from placex where address ? 'postcode' group by country_code,pc) as x ";
-    $sSQL .= "where ST_Point(x,y) is not null";
-    if (!pg_query($oDB->connection, $sSQL)) fail(pg_last_error($oDB->connection));
+    if (!pg_query($oDB->connection, 'TRUNCATE location_postcode')) {
+        fail(pg_last_error($oDB->connection));
+    }
+
+    $sSQL  = "INSERT INTO location_postcode";
```
drzraf: isn't here-document an option?

sql/functions.sql
```diff
@@ -515,36 +628,38 @@ CREATE OR REPLACE FUNCTION add_location(
...
   AS $$
 DECLARE
   locationid INTEGER;
-  isarea BOOLEAN;
   centroid GEOMETRY;
   diameter FLOAT;
   x BOOLEAN;
   splitGeom RECORD;
   secgeo GEOMETRY;
+  postcode TEXT;
```
drzraf: why are postcode TEXT rather than VARCHAR(X)?

sql/functions.sql
```diff
@@ -236,6 +256,99 @@ END;
...
+                                      OUT rank_search SMALLINT, OUT rank_address SMALLINT)
+AS $$
+DECLARE
+  part TEXT;
+BEGIN
+    rank_search := 30;
+    rank_address := 30;
+    postcode := upper(postcode);
+
+    IF country_code = 'gb' THEN
```
drzraf: Could you comment (or put a link to a wiki page) explaining the objective of this function, so that non-nominatim expert could contribute with their country?
(is there anything needed for french postalcodes?)

sql/tables.sql
```diff
@@ -197,8 +198,24 @@ CREATE TRIGGER place_before_delete BEFORE DELETE ON place
...
+DROP TABLE IF EXISTS location_postcode;
+CREATE TABLE location_postcode (
+  place_id BIGINT,
+  parent_place_id BIGINT,
+  rank_search SMALLINT,
+  rank_address SMALLINT,
+  indexed_status SMALLINT,
+  indexed_date TIMESTAMP,
+  country_code varchar(2),
+  postcode TEXT,
```
drzraf: same as above. Are postcode that large so that they need a TEXT field rather than a (large-enough) VARCHAR?

utils/setup.php
lonvia: Yes, we have discussed that. @mtmail might remember why we didn't switch. Might have been just historical.

sql/functions.sql
lonvia: I couldn't say what the length limit is for postcodes world-wide. Given that there is no difference in performance or storage space between TEXT and VARCHAR() in Postgresql, there is not reason to set an artificial limit.

sql/functions.sql
lonvia: This function basically determines at which administration level is covered by a single postcode. This is used when guessing how far away a postcode centroid can be and still be valid and when assigning a place to a postcode (i.e when you search by postcode, the result has a town or similar).

This is actually legacy code which I only have moved out of the placex_insert function. I'm not entirely sure if it is worth extending it for other countries. It needs to be reviewed at some point. There is one thing I forgot to put on the TODO list and that is recognizing postcodes by their pattern instead of taking it from the word list. This will be something where people can (and should) contribute country-specific knowledge.

lonvia:
I don't have a server available right now for setting up a public testing instance, although I might be able to find one next week. Until then the only way to test is to install your own instance with a country excerpt, I'm afraid.

gerv:
This is great :-) Do you know when we can expect to see it in search results on osm.org?

lonvia:
This is all live already including daily updates of the new artificial postcode table.

gerv:
Great, thank you!

gerv:
Is there any way to get OSM.org to jump straight to the Nominatim answer when searching for a postcode?

PR #798, opened 2017-09-19
fix to NearPoint::extractFromQuery handling first minus sign

mtmail:
for https://github.com/openstreetmap/Nominatim/issues/797


lonvia:
That was quick. Thank you!

PR #818, opened 2017-10-20
add index.php

Piskvor:
- usual file that webservers expect
- search.php seems to be capable of this role
- not having this file makes a fresh install seem nonfunctional ("install Nominatim, go to localhost:8080, see a 403")
- not implemented as a symlink, as not all filesystems support them

### DIFF
--- a/website/index.php
+++ b/website/index.php
@@ -1,0 +1,2 @@
+<?php
+require_once __DIR__ . '/search.php';


lonvia:
The installation instructions include setup instructions for Apache and nginx which both set the directory index to search.php. So if you follow these instructions then you should see the Nominatim main page as expected. And leaving the choice to the web server means that users have more choice how to set up their service (they might not want an index page at all).

lonvia:
Poster did not answer follow-up questions. Closing.

PR #819, opened 2017-10-20
Tiger 2017 data no longer contains -divroad- field

mtmail:
for https://github.com/openstreetmap/Nominatim/issues/817 I doublechecked the data documentation 2016 and 2017 and the field is indeed missing and doesn't seem to be replaced. I ran the script again 100 files and there were no other errors.


lonvia:
Works fine. There is a tarball of preprocessed Tiger 2017 data usable for Nominatim 3.0.1 at http://nominatim.org/data/tiger2017-nominatim-preprocessed.tar.gz.

PR #825, opened 2017-10-21
PHP code style rule to enforce single quotes

mtmail:
This PR contains only the rules. About 900 lines affected.

Using `phpcbf */**.php` you can autofix them all.

Watch out (`git diff`) for lines like `"abc \"de\" fg"` where single-quoting will cause the backslash to be in the output.

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -129,6 +129,23 @@
 
 
 
+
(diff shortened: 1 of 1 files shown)


lonvia:
Replaced by #835.

PR #841, opened 2017-10-27
UI: minimap causes main map not to initialize

mtmail:
fix for https://github.com/openstreetmap/Nominatim/issues/840


lonvia:
Hmm, now the search page doesn't work anymore...

mtmail:
That works again.

PR #866, opened 2017-12-16
Add function to recalculate counts for full-word search term

lonvia:
Calculating counts for full-word search terms helps to improve performance of search queries tremendously. On nominatim.osm.org the Postgres' IO throughput was reduced by something like 60%. Problem is that the calculation is quite expensive, so I don't want to make it a mandatory part of the installation process.


lonvia:
Merged in 637c5c2936a54e207a540ef0444b96a2fc0c1a80.

PR #873, opened 2018-01-01
Update US postcode file from 2017 Tiger data

### LINKED ISSUE #179: lat and lon for zipcode 45401 is not correct
When I use nominatim to search for the zip code 45401, it returns Brunswick, but it should be Dayton, OH. It seems this zipcode returns incorrect latitude and longitude.
The lat and lon for 45401 should be (39.750471,-84.268593), not 31.1319976, -81.4859267)

The url is http://nominatim.openstreetmap.org/search?q=45401&format=json&polygon=0&addressdetails=1&limit=3&countrycodes=us,ca.

lonvia:
Replaces the precomputed US postcode table (which afaik came from ZCTA data) with ZIP code centroids computed from Tiger 2017 data. ZIP codes which would cover a larger area have been removed, assuming that these represent PO boxes and the like. One example there is [12345](https://trac.openstreetmap.org/ticket/4113). 

Fixes #179.


mtmail:
This file contains 38095 postcodes. Even postcode 17927, which according to https://en.wikipedia.org/wiki/ZIP_Code was discontinued 2002.

US post office FAQ says there are about 42000 (that includes PO boxes), always changes from year to year obviously. geonames.org knows 41437, geopostcodes.com claims 60552. The ZIP Code Tabulation Areas of 2016 (https://www.census.gov/geo/maps-data/data/cbf/cbf_zcta.html) contains 33144 but the difference is expected (see https://en.wikipedia.org/wiki/ZIP_Code_Tabulation_Area).

So a good step forward. In the future it might make sense to calculate centroids for missing postcodes from OSM data still.

lonvia:
I will remove 17927. Do you know of a (licence-wise usable) list somewhere of discontinued ZIP codes?

The OSM data is already considered via the regular postcode centroid creation. The us_postcode table only adds postcodes that are not already available. (i.e. OSM data takes precedence because it is easier to fix)

lonvia:
Merged manually without 17927.

PR #878, opened 2018-01-08
Add parameters by stdin

fominomato:
When sending parameters with stdin we don't need apache web server because just invoke in cli mode.


mtmail:
Which query parameters are you typically using? We have https://github.com/openstreetmap/Nominatim/blob/master/utils/query.php as CLI solution.

fominomato:
Hum...understand now..
Sorry, I made the mistake of changing.
I suggested this modification, because this process did not find documentation.

PR #925, opened 2018-02-15
Additional aliases for "vei"

NeilRickards:
`veg` is a common alias for `vei` in Norway


lonvia:
I'm sorry to say that I don't take additional abbreviations at the moment. The short forms tend to cause all kinds of confusions. We really need to find a different solution for that problem.

NKAmapper:
This PR would certainly be helpful. Roughly half of Norwegian municipalities use "veg" in street names, the other half use "vei" (and a few have both). Even for a native Norwegian it is almost impossible to know which form is being used. The lack of an alias in Nominatim leads to a large number of missed hits when searching. Many street names use the definite suffix, so "vegen" and "veien" should also be included.

Another confusion is "gate", "gaten" and "gata", which are also frequently used variations in street names.

lonvia:
Closing this as the `tokenstringreplacements.inc` file will hopefully go away soon. This doesn't mean it is forgotten, it will just need to be realised in a different way.

PR #933, opened 2018-02-21
upgrade vagrant setup to centos 7.4, postgres 9.5, postgis 2.4

mtmail:
related to https://github.com/openstreetmap/Nominatim/issues/932

Nominatim requires postgresql > 9.2 so I upgraded the vagrant installation files, which in turn are used to generate our install documentation.

Tested with CentOS 7.2 and 7.4. I imported Monaco and forward geocoding worked.

Using the full path in `sudo /usr/pgsql-9.5/bin/postgresql95-setup initdb` is necessary because `source /etc/profile` doesn't seem to reload sudo's environment.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -33,7 +33,7 @@ Vagrant.configure("2") do |config|
   end
 
    config.vm.define "centos" do |sub|
-      sub.vm.box = "bento/centos-7.2"
+      sub.vm.box = "bento/centos-7.4"
       sub.vm.provision :shell do |s|
         s.path = "vagrant/Install-on-Centos-7.sh"
         s.privileged = false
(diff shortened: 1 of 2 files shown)


mtmail:
We say postgresql 9.1 is fine. But `setup.php` warns `Postgresql is too old. extratags and namedetails API not available.` when `hstore_to_json` isn't available (postgresql < 9.3).

lonvia:
Updating to Centos 7.4 is fine but I would like to stay with the shipped Postgres for the standard instructions. It would be good if we can find a box with SELinux enabled, though, so that part of the instructions gets tested as well.

mtmail:
Ok. Closing now, but will keep my branch in case we decide to upgrade later or point users to it.

PR #935, opened 2018-02-21
Nominatim 2dot5 instructions from wiki

mtmail:
These are https://wiki.openstreetmap.org/wiki/Nominatim/Installation and https://wiki.openstreetmap.org/wiki/Nominatim/Installation_on_CentOS converted to markdown. I think the disclaimer on top of the pages need to change.

I couldn't access the shellbot TIGER 2015 ftp mirror (timeout) and removed that URL.


lonvia:
Thanks for doing that. I have already started archiving 'historical' installation instructions on the nominatim.org site. So I guess this is the best place for this. I've moved the stuff there and opened a PR for you to review that I converted everything correctly. If you could have a quick look that would be awesome.

Closing here.

PR #941, opened 2018-02-23
PHP tests for ParameterParser

mtmail:
In the old PR I also had tests for Geocode.php with more complicated setup (needed either environment variables or mock objects), this PR is more straight forward.

Can you review the FIXMEs https://github.com/openstreetmap/Nominatim/pull/624#issuecomment-277505651 ?

### DIFF
--- a/test/php/Nominatim/ParameterParserTest.php
+++ b/test/php/Nominatim/ParameterParserTest.php
@@ -128,7 +128,6 @@ class ParameterParserTest extends \PHPUnit_Framework_TestCase
                                        ]);
 
         $this->assertSame(false, $oParams->getSet('non-exists', ['foo', 'bar']));
-        // FIXME: unclear if the default value has to be part of the set
         $this->assertSame('default', $oParams->getSet('non-exists', ['foo', 'bar'], 'default'));
         $this->assertSame('foo', $oParams->getSet('val1', ['foo', 'bar']));
 
--- a/test/php/Nominatim/ParameterParserTest.php
+++ b/test/php/Nominatim/ParameterParserTest.php
@@ -16,11 +16,6 @@ class ParameterParserTest extends \PHPUnit_Framework_TestCase
 {
 
 
-    protected function setUp()
-    {
-    }
-
-
     public function testGetBool()
     {
         $oParams = new ParameterParser([
@@ -46,8 +41,7 @@ class ParameterParserTest extends \PHPUnit_Framework_TestCase
         $oParams = new ParameterParser([
                                         'int1' => '5',
                                         'int2' => '-1',
-                                        'int3' => '',
-                                        'int4' => 0
+                                        'int3' => 0
                                        ]);
 
         $this->assertSame(false, $oParams->getInt('non-exists'));
@@ -55,15 +49,21 @@ class ParameterParserTest extends \PHPUnit_Framework_TestCase
         $this->assertSame(5, $oParams->getInt('int1'));
 
         $this->assertSame(-1, $oParams->getInt('int2'));
-        $this->assertSame(false, $oParams->getInt('int3')); // FIXME: should be 0 instead?
-        $this->assertSame(0, $oParams->getInt('int4'));
+        $this->assertSame(0, $oParams->getInt('int3'));
+    }
+
+
+    public function testGetIntWithNonNumber()
+    {
+        $this->setExpectedException(Exception::class, "Integer number expected for parameter 'int4'");
+        (new ParameterParser(['int4' => 'a']))->getInt('int4');
     }
 
 
     public function testGetIntWithEmpytString()
     {
         $this->setExpectedException(Exception::class, "Integer number expected for parameter 'int5'");
-        (new ParameterParser(['int5' => 'a']))->getInt('int5');
+        (new ParameterParser(['int5' => '']))->getInt('int5');
     }
 
 
--- a/test/php/Nominatim/ParameterParserTest.php
+++ b/test/php/Nominatim/ParameterParserTest.php
@@ -1,0 +1,214 @@
+<?php
+
+namespace Nominatim;
+
+use Exception;
+
+require_once('../../lib/ParameterParser.php');
+
+
+function userError($sError)
+{
+    throw new Exception($sError);
+}
+
+class ParameterParserTest extends \PHPUnit_Framework_TestCase
+{
+
+
+    protected function setUp()
+    {
+    }
+
+
+    public function testGetBool()
+    {
+        $oParams = new ParameterParser([
+                                        'bool1' => '1',
+                                        'bool2' => '0',
+                                        'bool3' => 'true',
+                                        'bool4' => 'false',
+                                        'bool5' => ''
+                                       ]);
+
+        $this->assertEquals(false, $oParams->getBool('non-exists'));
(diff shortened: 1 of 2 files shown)


test/php/Nominatim/ParameterParserTest.php
```diff
@@ -0,0 +1,214 @@
...
+
+        $this->assertEquals(false, $oParams->getString('non-exists'));
+        $this->assertEquals('default', $oParams->getString('non-exists', 'default'));
+        $this->assertEquals('abc', $oParams->getString('str1'));
+        $this->assertEquals(false, $oParams->getStringList('str2'));
+        $this->assertEquals(false, $oParams->getStringList('str3')); // FIXME: should be 0 instead?
+    }
+
+
+
```
lonvia: Is phpcs happy with so many empty lines?

test/php/Nominatim/ParameterParserTest.php
```diff
@@ -0,0 +1,214 @@
...
+                             'name' => 'name',
+                             'name:de' => 'name:de',
+                             'name:en' => 'name:en',
+                             'short_name' => 'short_name',
+                             'short_name:de' => 'short_name:de',
+                             'short_name:en' => 'short_name:en',
+                             'official_name' => 'official_name',
+                             'official_name:de' => 'official_name:de',
+                             'official_name:en' => 'official_name:en',
+                            ], $oParams->getPreferredLanguages('default'));
```
lonvia: This is surprising. I would have expected name:xx and short_name:XX to come before brand and ref.

test/php/Nominatim/ParameterParserTest.php
mtmail: We disabled the `Squiz.WhiteSpace.FunctionSpacing.After` rule because it complained about having 1 empty line after function closing block (afai remember). Looks like enabling it requires only two fixes. I'll address that in a separate PR.

lonvia:
> Can you review the FIXMEs #624 (comment) ?

The problem with the exceptions is still open.

mtmail:
In the previous PR the exceptions were used in the GeocoderTest to check if the generated SQL fails in a database query. That file isn't part of this PR.

lonvia:
If I read the code right, then you are still using try-catch for checking that a bad parameter input is caught. If the function erroneously returns false instead of going into userError() then the tests will still pass.

test/php/Nominatim/ParameterParserTest.php
lonvia: It looks like assertEquals() does not take order into account for arrays but order actually matters for the results here.

mtmail:
Replaced `assertEquals` with `assertSame`. That also uncovered three `0` vs `false` return values and `1.0` vs `1` for `getFloat`.

test/php/Nominatim/ParameterParserTest.php
```diff
@@ -0,0 +1,216 @@
...
+                                        'int3' => '',
+                                        'int4' => 0
+                                       ]);
+
+        $this->assertSame(false, $oParams->getInt('non-exists'));
+        $this->assertSame(999, $oParams->getInt('non-exists', 999));
+        $this->assertSame(5, $oParams->getInt('int1'));
+
+        $this->assertSame(-1, $oParams->getInt('int2'));
+        $this->assertSame(false, $oParams->getInt('int3')); // FIXME: should be 0 instead?
```
lonvia: This should throw for being an invalid value.

test/php/Nominatim/ParameterParserTest.php
```diff
@@ -0,0 +1,216 @@
...
+                                        'float2' => '-5',
+                                        'float3' => '',
+                                        'float4' => 0
+                                       ]);
+
+        $this->assertSame(false, $oParams->getFloat('non-exists'));
+        $this->assertSame(999, $oParams->getFloat('non-exists', 999));
+        $this->assertSame(1.0, $oParams->getFloat('float1'));
+        $this->assertSame(-5.0, $oParams->getFloat('float2'));
+        $this->assertSame(false, $oParams->getFloat('float3')); // FIXME: should be 0 instead?
```
lonvia: exception

test/php/Nominatim/ParameterParserTest.php
```diff
@@ -0,0 +1,216 @@
...
+                                        'str1' => 'abc',
+                                        'str2' => '',
+                                        'str3' => '0'
+                                       ]);
+
+        $this->assertSame(false, $oParams->getString('non-exists'));
+        $this->assertSame('default', $oParams->getString('non-exists', 'default'));
+        $this->assertSame('abc', $oParams->getString('str1'));
+        $this->assertSame(false, $oParams->getStringList('str2'));
+        $this->assertSame(false, $oParams->getStringList('str3')); // FIXME: should be 0 instead?
```
lonvia: yes

test/php/Nominatim/ParameterParserTest.php
```diff
@@ -0,0 +1,216 @@
...
+        $oParams = new ParameterParser([
+                                        'list1' => ',a,b,c,,c,d',
+                                        'list2' => 'a',
+                                        'list3' => '',
+                                        'list4' => '0'
+                                       ]);
+
+        $this->assertSame(false, $oParams->getStringList('non-exists'));
+        $this->assertSame(['a', 'b'], $oParams->getStringList('non-exists', ['a', 'b']));
+        // FIXME: unclear if empty string items should be removed
```
lonvia: Either removed or throw an exception. Maybe the latter is nicer for the user.

test/php/Nominatim/ParameterParserTest.php
```diff
@@ -0,0 +1,216 @@
...
+    {
+        $this->setExpectedException(Exception::class, "Parameter 'val4' must be one of: foo, bar");
+        (new ParameterParser(['val4' => 'faz']))->getSet('val4', ['foo', 'bar']);
+    }
+
+
+    public function testGetStringList()
+    {
+        $oParams = new ParameterParser([
+                                        'list1' => ',a,b,c,,c,d',
```
lonvia: How about adding a normal list without empty spots?

test/php/Nominatim/ParameterParserTest.php
```diff
@@ -0,0 +1,216 @@
...
+    public function testGetSet()
+    {
+        $oParams = new ParameterParser([
+                                        'val1' => 'foo',
+                                        'val2' => '',
+                                        'val3' => 0
+                                       ]);
+
+        $this->assertSame(false, $oParams->getSet('non-exists', ['foo', 'bar']));
+        // FIXME: unclear if the default value has to be part of the set
```
lonvia: This would be a programming error. Just ignore it.

test/php/Nominatim/ParameterParserTest.php
mtmail: PHP does some magic where even returning a '0' gets treated as false during the comparison. I added a note. Starting PHP7 we can define the return value of functions.

test/php/Nominatim/ParameterParserTest.php
mtmail: Just wanted to make sure it's we don't have unexpected behavior. Empty strings are now ignored.

PR #949, opened 2018-02-25
/status can now output json, including data date

mtmail:
/status output defaults to text (usually "OK") but can now also output JSON. Most logic moved to a module with unit tests.

Includes the timestamp when the database was last updated. Currently `/website/(search|replace|details).php` also do the SQL query so they could use Nominatim::Status instead. More likely we remove the HTML output from those endpoints soon anyway.

With BDD tests I can only get a HTTP 500 error. When I change `@APIDB` to `@DB` the error changes to "module call failed".


lonvia:
You should rebase on master. The new version of behave (used on travis) fails on some previously undetected formatting errors. I hope I've fixed them all now.

mtmail:
Thanks. Still battling the `PHP Warning:  require_once(DB.php): failed to open stream` error. Happens on travis, but in the Vagrant Ubuntu VM.

lonvia:
The diff shows a lot of unrelated stuff. Could you please rebase against master and then force-push the result to this branch.

test/php/Nominatim/StatusTest.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
+namespace Nominatim;
+
+require_once('../../lib/Status.php');
+require_once('DB.php');
```
lonvia: Have you tried switching the order of the two requires?

mtmail: All fine locally. So far it looks like on travis-ci `phpenv` is used to switched between three versions of PHP (5.6.x, 7.0.x, 7.1.x) and the path apt-get installs the `DB.php` is in either of the load paths. I found a work-around, will clean that up tonight.
For private repositories travis allows debugging (ssh into the environment to look around). Not possible for public repositories, which makes debugging a bit more time-consuming.

website/status.php
```diff
@@ -1,37 +1,57 @@
...
+$sOutputFormat = $oParams->getSet('format', array('text', 'json'), 'text');
 
 $oDB =& DB::connect(CONST_Database_DSN, false);
-if (!$oDB || PEAR::isError($oDB)) {
-    statusError('No database');
-}
+$oStatus = new Nominatim\Status($oDB);
+
+$sStatus = $oStatus->status();
+$bGotError = ($sStatus != 'OK');
```
lonvia: Having to parse a string for an error is not particularly efficient. How about having status() return the same array as you set up in $aResponse below, i.e. with status, code and message?

lib/Status.php
```diff
@@ -0,0 +1,61 @@
...
+        $oDataDates = $this->oDB->getAll($sSQL);
+
+        if (PEAR::isError($oDataDates)) {
+            throw Exception('Data date query failed '.$oDataDates->getMessage());
+        }
+        return [
+                'epoch' => $oDataDates[0][0],
+                'formatted' => $oDataDates[0][1]
+               ];
+    }
```
lonvia: I'd only return the epoch here and leave the formatting as an exercise to the caller. They probably want to choose the format themselves (for example, they might want to localise the format according to language of the caller).

mtmail: locali[sz]ing the the output is still possible with the epoch. I was thinking about easy of debugging when somebody call the endpoint on command-line.

website/status.php
mtmail: I cleaned that up a little but didn't want to put HTTP error codes into the lib/ to separate logic from presentation. Previously I had the `status` method throw exceptions, it made tests a bit unreadable. Could add that again.

lib/Status.php
lonvia: Even that will have a thin wrapper.

mtmail: only epoch is returned now

mtmail:
I improved the BDD tests with proper JSON parsing instead of string comparison.

website/status.php
```diff
@@ -1,37 +1,57 @@
...
-}
+$oStatus = new Nominatim\Status($oDB);
+
+$aResponse = [
+              'status' => 'ok',
+              'code' => 200,
+              'message' => 'OK'
+             ];
+
+$sErrorMsg = $oStatus->status(); // can be nil
```
lonvia: Sorry but I'm still not happy with that. Using the null value like that is confusing.

Let's go back to the format of the response message. Why does the code field need to be a HTTP error at all? The HTTP error is already sent with the HTTP header. How about a generic response that contains a status field as integer with 0 status ok and a Nominatim-specific error code otherwise and the message string.

lib/Status.php
```diff
@@ -0,0 +1,58 @@
...
+        if (!$iWordID) {
+            return 'No value';
+        }
+
+        return;
+    }
+
+    public function dataDate()
+    {
+        $sSQL = "SELECT EXTRACT(EPOCH FROM lastimportdate - '2 minutes'::interval) FROM import_status LIMIT 1";
```
lonvia: We probably should get rid of the 2 min hack as well at this point. The factor is there because the data from the last update step might not have been indexed yet and therefore cannot be found. However, indexing can take arbitrarily long so there is no telling how far back the last update step was completed.

mtmail:
closing PR because github sends an email for every commit and gets too noisy. Will reopen a clean branch/PR once I figured the travis environment out.

PR #968, opened 2018-03-09
details support json output

mtmail:
This is the continuation of https://github.com/openstreetmap/Nominatim/pull/439 by @agunnarsson, now includes BDD tests.

> New parameters for details are &addressdetails=1&linkedplaces=1&childplaces=1 default are false for all parameters details, but they have been included in links on html pages.
is still in place

I added the `&group_parents` because with handlebarjs functions it's quite complex to change data structures. Should that be called `&group_child_places` instead?

### DIFF
--- a/lib/template/details-json.php
+++ b/lib/template/details-json.php
@@ -1,0 +1,41 @@
+<?php
+
+$aPlaceDetails = $aPointDetails;
+
+$aPlaceDetails['geojson'] = json_decode($aPointDetails['asgeojson']);
+unset($aPlaceDetails['asgeojson']);
+
+if ($aAddressLines) {
+    $aPlaceDetails['address_lines'] = $aAddressLines;
+}
+
+if ($aLinkedLines) {
+    $aPlaceDetails['linked_lines'] = $aLinkedLines;
+}
+
+if ($aPlaceSearchNameKeywords) {
+    $aPlaceDetails['place_search_name_keywords'] = $aPlaceSearchNameKeywords;
+}
+
+if ($aPlaceSearchAddressKeywords) {
+    $aPlaceDetails['place_search_address_keywords'] = $aPlaceSearchAddressKeywords;
+}
+
+if ($aParentOfLines) {
+    $aPlaceDetails['parentof_lines'] = $aParentOfLines;
+
+    if ($bGroupParents) {
+        $aGroupedAddressLines = [];
+        foreach ($aParentOfLines as $aAddressLine) {
+            if ($aAddressLine['type'] == 'yes') $sType = $aAddressLine['class'];
+            else $sType = $aAddressLine['type'];
+
+            if (!isset($aGroupedAddressLines[$sType]))
+                $aGroupedAddressLines[$sType] = [];
+            $aGroupedAddressLines[$sType][] = $aAddressLine;
+        }
+        $aPlaceDetails['parentof_lines'] = $aGroupedAddressLines;
+    }
+}
+
+javascript_renderData($aPlaceDetails);
--- a/website/details.php
+++ b/website/details.php
@@ -9,13 +9,20 @@ ini_set('memory_limit', '200M');
 
 $oParams = new Nominatim\ParameterParser();
 
-$sOutputFormat = 'html';
+$sOutputFormat = $oParams->getSet('format', array('html', 'json'), 'html');
+
 $aLangPrefOrder = $oParams->getPreferredLanguages();
 $sLanguagePrefArraySQL = 'ARRAY['.join(',', array_map('getDBQuoted', $aLangPrefOrder)).']';
 
 $sPlaceId = $oParams->getString('place_id');
 $sOsmType = $oParams->getSet('osmtype', array('N', 'W', 'R'));
 $iOsmId = $oParams->getInt('osmid', -1);
+
+
+$bIncludeAddressDetails = ($sOutputFormat == 'html') ? 1 : $oParams->getBool('addressdetails');
+$bIncludeLinkedPlaces = ($sOutputFormat == 'html') ? 1 : $oParams->getBool('linkedplaces');
+$bIncludeChildPlaces = ($sOutputFormat == 'html') ? 1 : $oParams->getBool('childplaces');
+$bGroupParents = $oParams->getBool('group_parents', 0);
 
 $oDB =& getDB();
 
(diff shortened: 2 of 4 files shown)


website/details.php
```diff
@@ -9,14 +9,21 @@
...
+
 $aLangPrefOrder = $oParams->getPreferredLanguages();
 $sLanguagePrefArraySQL = 'ARRAY['.join(',', array_map('getDBQuoted', $aLangPrefOrder)).']';
 
 $sPlaceId = $oParams->getString('place_id');
 $sOsmType = $oParams->getSet('osmtype', array('N', 'W', 'R'));
 $iOsmId = $oParams->getInt('osmid', -1);
 
+
+$bIncludeAddressDetails = ($sOutputFormat == 'html') ? 1 : $oParams->getBool('addressdetails');
```
lonvia: ... = $oParams->getBool('addressdetails', $sOutputFormat == 'html');

dto below

website/details.php
```diff
@@ -9,14 +9,21 @@
...
+$bIncludeLinkedPlaces = ($sOutputFormat == 'html') ? 1 : $oParams->getBool('linkedplaces');
+$bIncludeChildPlaces = ($sOutputFormat == 'html') ? 1 : $oParams->getBool('childplaces');
+$bGroupParents = $oParams->getBool('group_parents', 0);
```
lonvia: On a general note: use true/false instead of 0/1 for bools.

website/details.php
```diff
@@ -148,62 +155,71 @@
 }
 
 // Address
-$aAddressLines = getAddressDetails($oDB, $sLanguagePrefArraySQL, $iPlaceID, $aPointDetails['country_code'], -1, true);
+$aAddressLines = false;
+if ($bIncludeAddressDetails) {
+    $aAddressLines = getAddressDetails($oDB, $sLanguagePrefArraySQL, $iPlaceID, $aPointDetails['country_code'], -1, true);
```
lonvia: Please keep line length below 100 for new/changed code.

lib/template/details-json.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+$aPlaceDetails = $aPointDetails;
+
+$aPlaceDetails['geojson'] = json_decode($aPointDetails['asgeojson']);
+unset($aPlaceDetails['asgeojson']);
```
lonvia: Can we make the geometry optional as well for json, i.e. a URL parameter which is off by default? Preferably use the usual polygon_geojson parameter.

lib/template/details-json.php
```diff
@@ -0,0 +1,41 @@
...
+            if ($aAddressLine['type'] == 'yes') $sType = $aAddressLine['class'];
+            else $sType = $aAddressLine['type'];
+
+            if (!isset($aGroupedAddressLines[$sType]))
+                $aGroupedAddressLines[$sType] = [];
+            $aGroupedAddressLines[$sType][] = $aAddressLine;
+        }
+        $aPlaceDetails['parentof_lines'] = $aGroupedAddressLines;
+    }
+}
```
lonvia: May I suggest a slightly simplified naming here:
* geojson -> geometry
* address_lines -> address
* linked_lines -> linked_places
* place_search_name_keywords -> name_keywords  or possibly: keywords->name
* place_search_name_address -> address_keywords or possibly: keywords->address
* parentof_lines -> parent_of

lonvia:
Some other observations after looking at the output:

* there is an odd 'name' field which contains a string-encoded 'aNames' field. I'd rename the latter to 'names' and remove the first
* aAddressTags -> address_tags
* aExtraTags -> extra_tags
* what about empty fields, should they be removed in json?
* there are a lot of int and bool types encoded as string
* keywords contains all raw data from word now, I wonder if we should format that nicely somehow
* addressdetails: name needs to be a hash
* childplaces: what is namelength?

mtmail:
Updated based on feedback:
* fieldnames snake_cased, use integer/boolean types whenever possible
* removed fields which aren't used in the HTML (nominatim-ui) output
* if `&polygon_geojson=1` is set then the `geometry` field contains a polygon, otherwise centrepoint
* split BDD tests into `simple.feature` and `params.feature`

Here's a request for Berlin (state) with all possible parameters enabled http://localhost:8089/nominatim/details.php?format=json&osmtype=R&osmid=62422&addressdetails=1&keywords=1&linkedplaces=1&childplaces=1&group_childplaces=1&polygon_geojson=1

Other notes:
* we call the parameter `childplaces` to enable the field `parentof`, should both be children?
* I renamed the field `wikipedia` to `calculated_wikipedia` because that's what the HTML labels it. Should we do the same with `postcode`?
* In details output the `icon` contains the filename, in search and reverse output the `icon` contains a pull path plus filename. Should they be the same?
* the `indexed_date` field contains a string, e.g. `2018-02-01 08:32`. Should that be an integer so the frontend can format it freely?

lib/template/details-json.php
```diff
@@ -0,0 +1,102 @@
...
+        'localname' => $aFull['localname'],
+        'place_id' => (int) $aFull['place_id'],
+        'osm_id' => (int) $aFull['osm_id'],
+        'osm_type' => formatOSMType($aFull['osm_type']),
+        'class' => $aFull['class'],
+        'type' => $aFull['type'],
+        'admin_level' => (int) $aFull['admin_level'],
+        'rank_address' => (int) $aFull['rank_address'],
+        'distance' => (float) $aFull['distance']
+    ];
```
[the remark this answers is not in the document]
mtmail: The short form `[]` got introduced in PHP 5.4. We use 655x long form, 67x short form. Happy to switch to long and add a PHPCS rule (separate PR)

mtmail:
* use `array()` instead of `[]` for array initialization for new code. PR to check/fix in other files will come separately
* renamed `childplaces` parameter and `parentsof` return key to `hierarchy`
* `linkedplaces` parameter defaults to `true`
* renamed `postcode` key to `calculated_postcode`
* better JSON types (integer) for `place_id`, `osm_id`
* output `null` when address values not set. E.g. for `place:country_code` or `place:postcode` there's no `place_id`, `osm_id`, `admin_level`
* skip unused hierarchy field, e.g. `namelength`
* no longer map `osm_type` (e.g. R => relation)
* `indexed_date` is a RFC822 date (e.g. `Sat, 10 Feb 18 00:35:49 +0000`) now to be consistent with our XML output. ISO8601/RFC3339 would be `2018-02-10T00:35:39+00:00` (fyi PHP skips the last ":", Postgresql skips the "T" by default).

New Test URL
http://localhost:8089/nominatim/details.php?format=json&osmtype=R&osmid=62422&addressdetails=1&keywords=1&linkedplaces=1&hierarchy=1&group_hierarchy=1&polygon_geojson=1

lib/template/details-json.php
```diff
@@ -0,0 +1,102 @@
...
+
+    return $aMapped;
+};
+
+$funcMapKeyword = function ($aFull) {
+    $aMapped = [
+        'id' => (int) $aFull['word_id'],
+        'token' => $aFull['word_token']
+    ];
```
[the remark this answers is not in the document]
mtmail: Changed, further occurrences will be caught/fixed by https://github.com/openstreetmap/Nominatim/issues/1008

mtmail:
* changed human readable date output to ISO8601/RFC3339. Will use the same format for /status
* changed lat/lon fields to a `centroid` field containing a geojson point
* removed `rank_search_label` and will put a mapping into the frontend instead https://github.com/openstreetmap/Nominatim/blob/28ee59dd643e4582ade1aa8fd6a788f0a4daf69e/sql/functions.sql#L2577

mtmail:
OK, should be cleaner now.

Should I prepare a PR with a documentation page (markdown file)? Or wait until it's used in production?

PR #970, opened 2018-03-09
converted wiki API documentation to markdown

mtmail:
This is 80-90% of https://wiki.openstreetmap.org/wiki/Nominatim into one markdown file. Should it be split by endpoint for readability, especially since Details will have an API endpoint soon?  Should the filename contain "API"?

### DIFF
--- a/sql/partition-functions.src.sql
+++ b/sql/partition-functions.src.sql
@@ -129,7 +129,7 @@ BEGIN
           ST_Distance(centroid, point) as distance, null as isguess
           FROM search_name_-partition-
           WHERE name_vector @> isin_token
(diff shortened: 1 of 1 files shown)


lonvia:
One page per endpoint sounds good. If you can add a subheader 'API reference' and then put the pages under that, that would be ideal.

If you add a new subdirectory in docs, you need to add symlink creation in `docs/CMakeList.txt`. Then go to the build directory, and run cmake and `make doc`. Now you can run `mkdocs serve` and watch the documentation live in the browser. The detour via the build directory is necessary because of the vagrant instructions being built. I'll try to add instructions about that somewhere.

PR #971, opened 2018-03-10
merge json and jsonv2 templates, they were very similar

mtmail:
Looking at the geojson PR https://github.com/openstreetmap/Nominatim/pull/707 I noticed the json and jsonv2 templates are very similar. 

[json-vs-jsonv2-diffs.txt](https://github.com/openstreetmap/Nominatim/files/1798874/json-vs-jsonv2-diffs.txt)

The geojson templates (after phpcs cleanup) will also be only a couple of lines different, so I think it's best to avoid code duplication now.

I ran PHP unit tests, phpcs and BDD tests (against a planet database). A couple of failures which are not template related.
```
behave -DBUILDDIR=/home/vagrant/build api/search/
634 steps passed, 17 failed, 9 skipped, 0 undefined
behave -DBUILDDIR=/home/vagrant/build api/reverse/
251 steps passed, 7 failed, 2 skipped, 0 undefined
```


lonvia:
Merged as 88beeb79164d0fec40778678e5d93ae44b8f07d8 (with a tiny formatting fix).

PR #972, opened 2018-03-11
HTML error output when asking for json format

rumpletweezer:
Still needs work to get behave tests working properly.

### DIFF
--- a/test/bdd/steps/queries.py
+++ b/test/bdd/steps/queries.py
@@ -533,6 +533,5 @@ def check_for_duplicates(context, neg):
 
 @then(u'the result page will include (?P<error>.*).')
 def error_check(context,error):
-    if not error in getattr(context.response,'page'):
-        assert not error, "Incorrect error displayed. Expected page to contain \"%s\" but got %s" % (error, vars(context.response))
+    assert_in(error,context.response.page,'thisIsTheError')
 
--- a/test/bdd/steps/queries.py
+++ b/test/bdd/steps/queries.py
@@ -530,3 +530,9 @@ def check_for_duplicates(context, neg):
         assert not has_dupe, "Found duplicate for %s" % (dup, )
     else:
         assert has_dupe, "No duplicates found"
+
+@then(u'the result page will include (?P<error>.*).')
+def error_check(context,error):
+    if not error in getattr(context.response,'page'):
+        assert not error, "Incorrect error displayed. Expected page to contain \"%s\" but got %s" % (error, vars(context.response))
+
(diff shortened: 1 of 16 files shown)


test/bdd/steps/queries.py
```diff
@@ -530,3 +530,9 @@ def check_for_duplicates(context, neg):
         assert not has_dupe, "Found duplicate for %s" % (dup, )
     else:
         assert has_dupe, "No duplicates found"
+
+@then(u'the result page will include (?P<error>.*).')
+def error_check(context,error):
+    if not error in getattr(context.response,'page'):
+        assert not error, "Incorrect error displayed. Expected page to contain \"%s\" but got %s" % (error, vars(context.response))
+
```
lonvia: Just for the syntax, it should be something like that:

def error_check(..):
    assert_in(error, context.response.page, "some error")

Have a look at https://docs.python.org/3/library/unittest.html#assert-methods for the different asserts. The underscore notation comes from the nose library.

On a more general note, you should rather extend the existing 'a HTTP xxx is returned' to 'a HTTP xxx is returned with error "<msg>"'. (Just saying in case you hadn't planed that anyways.)

website/reverse.php
```diff
@@ -46,7 +46,7 @@
         }
     }
 } elseif ($sOutputFormat != 'html') {
-    userError('Need coordinates or OSM object to lookup.');
+    userError('Need coordinates or OSM object to lookup.', $oParams->getString('format'));
```
lonvia: You must not get the format from the parameters directly like that because it disregards the default setting when the parameter is not given. Use $sOutputFormat instead. This applies to all the appearances in this PR. This might be the issue with your failing test.

lonvia:
Looking at what you have got so far, I can't help thinking that the changes would be a lot simpler, if we switched the userError() functions to exceptions first. There are still quite a few questionable uses of oParam->getString() and they would all just go away if the exception was just caught and converted in the main website/* scripts where the output format is readily available.

gopi-ar:
A better approach to ensure correct format might be to add error_handlers() for errors / uncaught exceptions and shutdown handlers for fatal errors right at the script's initialisation. We can then output the error in json / xml / geojson.

Even though the master branch is  compatible with PHP7, it throws a number of Pear::DB deprecation errors which we catch and ignore with a separate error handler we inject before loading nominatim. We deal with fatal `out of memory` errors for long search results in a similar fashion.

This might also push the effort required to add try/catches across the board off for a few months.

lonvia:
Superseded by #1181.

PR #973, opened 2018-03-14
get apt-get php-db package running on travis-ci

mtmail:
Separated from https://github.com/openstreetmap/Nominatim/pull/949 where I caused dozens of commits to find this solution.

I think once we have PHP composer the hack will no longer be needed.


lonvia:
I think it would be better to force travis to use the system phpunit installation. Doesn't look like you can do that with travis env variables but the phpenv manual says, you can force the use of system php with: `phpenv global system`.

mtmail:
I tried the `phpenv global system`. On travis-ci's Ubuntu trusty that enables PHP 5.5.9 and PHPUnit 3.7. PHPUnit fails because it can't mock objects. I was hoping for PHP 5.6 at least.

Installing PHPUnit 5.x (support ended last month) fails because it requires PHP 5.6.

mtmail:
`pear` installs DB into `/usr/share/php/DB.php`. The load path of php inside the test suite is `include_path='.:/home/travis/.phpenv/versions/5.6.32/share/pear`, so it's pretty similar.

lonvia:
pear seems to be using /usr/bin/php by default instead whatever is in the path. This might do the trick:

    PHP_PEAR_PHP_BIN=`which php` pear ...

PR #974, opened 2018-03-14
remove the 2-minute offset hack from data date in HTML output

mtmail:
As noted in https://github.com/openstreetmap/Nominatim/pull/949

    We probably should get rid of the 2 min hack as well at this point. The factor is there because the 
    data from the last update step might not have been indexed yet and therefore cannot be found. 
    However, indexing can take arbitrarily long so there is no telling how far back the last update step 
    was completed.

### DIFF
--- a/website/details.php
+++ b/website/details.php
@@ -231,7 +231,7 @@ if ($oParams->getBool('keywords')) {
 logEnd($oDB, $hLog, 1);
 
 if ($sOutputFormat=='html') {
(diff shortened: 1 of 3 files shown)


lonvia:
I think it would be better to force travis to use the system phpunit installation. Doesn't look like you can do that with travis env variables but the `phpenv` manual says, you can force the use of system php with: `php global system`.

mtmail:
@lonvia Was your last comment meant for php-db and https://github.com/openstreetmap/Nominatim/pull/973 ?

lonvia:
oops, yes

PR #978, opened 2018-03-18
search/reverse/lookup with geojson output

mtmail:
Replacement for https://github.com/openstreetmap/Nominatim/pull/707, fixes https://github.com/openstreetmap/Nominatim/issues/118

Follows the JSONv2 output fields (category vs class). For reverse returns a FeatureCollection with one result. Unless no result was found, then it's the same `{"error":"Unable to geocode"}` the other json formats return.

PHP, PHPcs, BDD api tests work. Can't see anything in the error log.

### DIFF
--- a/lib/template/address-json.php
+++ b/lib/template/address-json.php
@@ -17,7 +17,7 @@ if (empty($aPlace)) {
     if (isset($aPlace['lat'])) $aFilteredPlaces['lat'] = $aPlace['lat'];
     if (isset($aPlace['lon'])) $aFilteredPlaces['lon'] = $aPlace['lon'];
 
-    if ($sOutputFormat == 'jsonv2') {
+    if ($sOutputFormat == 'jsonv2' || $sOutputFormat == 'geojson') {
         $aFilteredPlaces['place_rank'] = $aPlace['rank_search'];
 
         $aFilteredPlaces['category'] = $aPlace['class'];
--- a/lib/Geocode.php
+++ b/lib/Geocode.php
@@ -249,6 +249,11 @@ class Geocode
         $this->oPlaceLookup->loadParamArray($oParams, $sForceGeometryType);
         $this->oPlaceLookup->setIncludeAddressDetails(false);
         $this->oPlaceLookup->setIncludePolygonAsPoints($oParams->getBool('polygon'));
+
+        if ($oParams->getString('format', '') == 'geocodejson') {
+            $this->oPlaceLookup->setAddressDetails(true);
+            $this->oPlaceLookup->setAddressAdminLevels(true);
+        }
     }
 
     public function setQueryFromParams($oParams)
--- a/lib/PlaceLookup.php
+++ b/lib/PlaceLookup.php
@@ -11,6 +11,7 @@ class PlaceLookup
     protected $aLangPrefOrderSql = "''";
 
     protected $bAddressDetails = false;
+    protected $bAddressAdminLevels = false;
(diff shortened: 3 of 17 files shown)


lonvia:
There has been https://github.com/geocoders/geocodejson-spec around for a while. I wonder if we should follow that spec for geojson and just add fields where there is additional information. One advantage would be that https://github.com/geocoders/geocoder-tester could then be run against Nominatim.

mtmail:
I vote for creating a thin wrapper/proxy instead like those who exist for mapbox, navitia, osmnames already https://github.com/geocoders

My worry is that future versions of the spec change fields (e.g. https://github.com/geocoders/geocodejson-spec/issues/15) and we wouldn't be able to support those without breaking backward compatibility (or supporting a `&version` parameter). There haven't been many changes to the spec but Nominatim will always move slower.

lib/template/address-json.php
```diff
@@ -57,4 +57,8 @@
     }
 }
 
-javascript_renderData($aFilteredPlaces);
+if ($sOutputFormat == 'geojson' && !isset($aFilteredPlaces['error'])) {
+    javascript_renderData(jsonResultsToGeoJson([$aFilteredPlaces]));
```
lonvia: This is a very clever idea but I'm a bit concerned that now the data gets copied around yet another time.

lib/lib.php
```diff
@@ -403,6 +403,32 @@ function getResultDiameter($aResult)
...
+function jsonResultsToGeoJson($aResults)
+{
+    $funcConvert = function ($aResult) {
+        $aGeoJsonFeature = [
+            'type' => 'Feature',
+            'properties' => $aResult,
+            'geometry' => [
+                'type' => 'Point',
+                'coordinates' => [(float) $aResult['lon'], (float) $aResult['lat']]
+            ]
```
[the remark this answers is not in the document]
mtmail: fixed

lib/lib.php
```diff
@@ -403,6 +403,32 @@ function getResultDiameter($aResult)
...
+        ];
+        if (isset($aResult['asgeojson'])) {
```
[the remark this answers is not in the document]
mtmail: Absolutely. Fixed.

lib/template/address-json.php
mtmail: I started with having everything in the `search-json.php` template but the various conditions became a mess and a lot harder to test. Later I moved the logic into `Lib.php`

mtmail:
Updated:
* `licence` key moved from Features to FeatureCollection
* renamed `boundingbox` to `bbox` and fixed the coordinate ordering (and variable type)
* removed `lat` and `lon` keys

mtmail:
Cleanup, squashed commits, forced pushed

lib/template/address-json.php
lonvia: What I meant is that I would prefer an extra template instead of transforming the json template yet again.

mtmail:
@lonvia I added a format=geocoderjson as well in separate templates. Not sure how to fill the accuracy for forward geocoding.

lonvia:
Looks good but there is a couple of failing BDD tests now.

mtmail:
Rebased. BDD tests work.

lonvia:
I've merged this now with the tiny change that `addressdetails` is required for `geocodejson` as well to get extended infos like about street, city, etc.  There is still API documentation missing for the new formats.

PR #993, opened 2018-03-27
Immenbusch BDD failed after ranking returns a place in France first

mtmail:
Changed two API BDD tests because they fail against my planet installation. Checked against nominatim.osm.org output.

https://nominatim.openstreetmap.org/search.php?q=immenbusch returns a place in France on first position.

The Steinwald query indeed returns more components https://nominatim.openstreetmap.org/search.php?q=2+Steinwald%2C+Austria&format=jsonv2&addressdetails=1&pretty=1

### DIFF
--- a/test/bdd/api/search/queries.feature
+++ b/test/bdd/api/search/queries.feature
@@ -11,6 +11,8 @@ Feature: Search queries
           | house_number | 2 |
           | hamlet       | Steinwald |
           | postcode     | 6811 |
+          | county       | Feldkirch |
+          | state        | Vorarlberg |
           | country      | Austria |
           | country_code | at |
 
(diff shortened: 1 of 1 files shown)


lonvia:
The tests must pass against a database containing https://www.nominatim.org/data/test/nominatim-api-testdata.pbf. Do you feel like creating a newer version of that file (i.e. extract from current planet), so that it is closer to your planet installation? Instructions are in test/README.

mtmail:
Excellent. I'll also check if it's possible to load that *pbf on travis-ci to run the API tests. Their time limit is 30 minutes.

mtmail:
@lonvia Quick feedback: The database needs 4GB while travis puts the postgresql data directory in a speed-optimized 768MB ramfs. Over that hurdle the import takes longer than 45min and travis terminates public repository builds after 50min.

PR #1002, opened 2018-04-06
when looking for keywords on detail page SQL bracket error was possible

mtmail:
fixes https://github.com/openstreetmap/Nominatim/issues/999


lonvia:
That was quick. Thank you.

PR #1013, opened 2018-04-12
add JSON format to /status endpoint

mtmail:
Replaces https://github.com/openstreetmap/Nominatim/pull/949

* The 2 minute hack is also gone, just like in master
* The date gets returned in ISO format instead of epoch timestamp (matches /details)
* json format will always return HTTP 200. Any errors will be in a status!=0 with a message string

Feel free to suggest another name for the `status` field or different status codes.

### DIFF
--- a/website/status.php
+++ b/website/status.php
@@ -9,53 +9,47 @@ require_once(CONST_BasePath.'/lib/Status.php');
 
 $oParams = new Nominatim\ParameterParser();
 $sOutputFormat = $oParams->getSet('format', array('text', 'json'), 'text');
-$bForceError = $oParams->getBool('force_error', false);
+$bForceError = $oParams->getInt('force_error', false);
 
 $oDB =& DB::connect(CONST_Database_DSN, false);
 $oStatus = new Nominatim\Status($oDB);
 
 
-$sErrorMsg = $oStatus->status(); // can be nil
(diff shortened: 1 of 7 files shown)


website/status.php
```diff
@@ -1,37 +1,55 @@
...
-
-function statusError($sMsg)
-{
-    header('HTTP/1.0 500 Internal Server Error');
-    echo 'ERROR: '.$sMsg;
-    exit;
-}
+$oParams = new Nominatim\ParameterParser();
+$sOutputFormat = $oParams->getSet('format', array('text', 'json'), 'text');
+$bForceError = $oParams->getInt('force_error', false);
```
lonvia: Was this for debugging? Can this go away now?

mtmail: The `force_error` parameter is for BDD tests (test/bdd/api/status/simple.feature) because the /status endpoint for text output returns HTTP 500, but for JSON returns HTTP 200.

lonvia: I'm reluctant to accept test feature like that in a production API. If you want to test the error condition, how about sneaking a bad database name into the test? You could add a context tag that forces a different DB name. Have a look at before_scenario() in tests/bdd/environment.py. That might be a good place to start.

mtmail:
Updated nominatim-ui as well.

Issue #1058, opened 2018-06-12
improve local search


dieterdreist:
Right now, if I search for "Viale Carnaro" with a viewbox in zoom15 that is slightly not containing the road in Rome, Nominatim will point me to a "Via Carnaro" (i.e. different name) 200 km away and the precise match is not even on the first page of results (only "Via Carnaro"s are found).
e.g.
https://nominatim.openstreetmap.org/search.php?q=viale+carnaro&polygon_geojson=1&viewbox=12.51491%2C41.93361%2C12.53014%2C41.92962
IMHO it would lead to better results if:

1.  increase the viewbox according to the zoom level for "high" and maybe "medium" zoomlevels (high zoom values will increase to "city scale", e.g. z11). When I write "zoomlevel" and increase zoomlevel, I actually mean bounding box extents (should be generously buffered, especially for zoomed in maps).
2. prefer accurate correlation (viaLE) over fuzzy matching (VIA)
3. maybe prefer higher road ranks over lower ones (here we find a residential street while in Rome the exact match is tertiary). Not sure about this, if it should be done, it would have to be less important than the spatial criterion (1.)

lonvia:
1. The street is not slightly outside the viewbox, it is more than twice away from it. Nominatim does give a bit of leeway but not that much.  There is no concept of zoomlevels. The idea is that if you need more space covered, you need to give a larger viewbox. (There is already ticket #824 for reconsidering how the viewbox is considered when it is relatively small.)
2. That's currently not possible without reworking the way abbreviations are handled.
3. For most cases you don't want that. Even in the example given, you'd still have complained if residential Via Peralba hadn't been found.

Please don't cramp tons of suggestions into one ticket, this makes it impossible to do have anything actionable. Closing because of that.

dieterdreist:
2. there are no abbreviations involved in this search. "Viale" is a word, "Via" is a different word. The abbreviations are "V.le" and "V." You cannot abbreviate "Viale" as "Via" or "Via."

3. Yes, was just an idea, but I agree you might not want it.

PR #1062, opened 2018-06-14
Display viewbox on map

mtmail:
fixes https://github.com/openstreetmap/Nominatim/issues/1061

URLSearchParams is not available (yet) in all browsers https://caniuse.com/#feat=urlsearchparams so I added a polyfill.

image


lonvia:
Thanks!

lonvia:
The example in #1058 still doesn't work because the order in viewbox is not as Leaflet expects it. Nominatim does not care about order.

mtmail:
I see the viewbox. Adding ",roma" to the URL because currently a road in Naples is returned

https://nominatim.openstreetmap.org/search.php?q=viale+carnaro,roma&polygon_geojson=1&viewbox=12.51491%2C41.93361%2C12.53014%2C41.92962

lonvia:
Indeed, I was just too far zoomed out so it got too small. All good.

PR #1077, opened 2018-06-27
Export script

### LINKED ISSUE #377: Finding street names of a given county
Is there a way to filter out the cities,street names, buildings and houses of a given (country,state,county).

thanks in advance

gemo1011:
this fixes #377 

lonvia wrote the script a long time ago and I made it work with the current master.


lonvia:
Nice, thank you.

PR #1112, opened 2018-07-22
Remove country and state nodes from address computation

lonvia:
OSM has by now almost complete coverage of admin
boundaries up to state level. Place nodes will do more
harm than good in this case.


lonvia:
Manually merged after rebase.

PR #1114, opened 2018-07-23
PHP code style: use long array syntax

mtmail:
Fixes PHPCS error `Short array syntax is not allowed`


lonvia:
I wonder why Travis didn't report those.

PR #1146, opened 2018-08-20
Allow custom locations for PHP binary

ThomasBarris:
**Issue**
https://github.com/openstreetmap/Nominatim/issues/1109

Adding the detection of the PHP binary to CMakelists.txt and adding the variable to settings/default.php i.e. with value in /settings/settings.php after cmake run.

Replaced hard coded shebangs `#!/usr/bin/php -Cq` in all utils/*.php by `#!@PHP_BIN@ -Cq` that will be replaced by detected PHP_BIN during cmake during copy to build/utils/

Setup.php splitted into setup_functions.php with most of the program logic. The remaining part of setup.php is somehow controlling the flow. The update.php is not using a passthru anymore but includes the setup_functions.php and uses the function directly.

This is my first real pull request ever (except README updates). Please let me know if something ist wrong.

**Test Results**

`behave -DBUILDDIR=/srv/gnominatim/Nominatim/build db osm2pgsql`
......
```
20 features passed, 0 failed, 0 skipped
350 scenarios passed, 0 failed, 3 skipped
1203 steps passed, 0 failed, 27 skipped, 0 undefined
Took 2m23.080s
```

`phpunit .`
.....
`OK (42 tests, 197 assertions)`


ThomasBarris:
Log of an import with errors during the Wikipedia part 
[setup.log](https://github.com/openstreetmap/Nominatim/files/2311798/setup.log)

utils/setup_functions.php
```diff
@@ -0,0 +1,931 @@
...
+        echo "\n";
+        info('Setup finished.');
+    }
+}
+
+// *********************************
+
+function pgsqlRunScriptFile($sFilename)
+{
+    global $aCMDResult;
```
lonvia: You shouldn't be using globals anymore when moving this to a module. Just hand it in as a parameter like in the other functions although I admit that this is a bit of a pain. Another option would be to move the whole code in this file in a class and hand in $aCmdResult in the constructor, then you don't have to worry any more at all handing it around.

ThomasBarris:
I have made some more changes since the update did not work. It works now. I will revert the large change and create a new pull request once the cmake changes are done

ThomasBarris:
The easy part is done. 

Once you've merged, I will work on the other part i.e. avoiding the passthru by splitting/restructuring the setup.php

lonvia:
I've merged this now. For the record, please put unrelated changes into separate PRs. I know that it can be a bit annoying if it is a one liner but it does make changes infinitely easier to follow.

PR #1151, opened 2018-08-21
documentation for /details endpoint

mtmail:


docs/api/Details.md
```diff
@@ -0,0 +1,147 @@
...
+## Parameters
+
+The details API has the following two formats:
+
+```
+  https://nominatim.openstreetmap.org/details?osmtype=[N|W|R]&osmid=<value>
+```
+
+Both parameters are mandatory, the type is one of node(N), way(W) or relation(R). Up to 50 ids
+can be queried at the same time.
```
lonvia: C&P error? I think it's only one object per request.

docs/api/Details.md
```diff
@@ -0,0 +1,147 @@
...
+
+Or
+
+```
+  https://nominatim.openstreetmap.org/details?placeid=<value>
+```
+
+The placeid created assigned sequentially during Nominatim data import. It is different between servers and changes when data gets reimported. Therefore it can't be used as permanent id and shouldn't be used in bug reports.
```
lonvia: 'created assigned' -> 'assigned'

mtmail:
@lonvia I refined the wording. I left it as second commit so you can spot the changes easier.

PR #1152, opened 2018-08-21
in /details JSON output add check if place (type) has an icon set

mtmail:
Similar to https://github.com/openstreetmap/Nominatim/pull/1113 a place's icon can be undefined.


lib/template/details-json.php
```diff
@@ -26,7 +26,9 @@
 
 $aPlaceDetails['extratags'] = $aPointDetails['aExtraTags'];
 $aPlaceDetails['calculated_wikipedia'] = $aPointDetails['wikipedia'];
-$aPlaceDetails['icon'] = CONST_Website_BaseURL.'images/mapicons/'.$aPointDetails['icon'].'.n.32.png';
+if (isset($aPointDetails['icon'])) {
```
lonvia: This gets set to false in website/details.php:139, so you need to check for `if ($aPointDetails['icon'])`.

mtmail: fixed, squashed and force pushed

PR #1153, opened 2018-08-23
Fix example links

JamesKingdom:
These examples didn't actually link to the page.


lonvia:
Unfortunately, this still doesn't produce the correct link. It looks like the only way to get that link and text is by repeating the URL. I've done this now in c80c80200c0c2e78811d0b96e225406471e0ad48. Thanks for noticing and attempting to fix.

PR #1154, opened 2018-08-24
moving functions from setup.php to a lib file .....

ThomasBarris:
...... in lib/setup_functions.php and change a passthru in update.php by calling the function with this new lib.

### **TESTS**
**PHPUNIT**
PHPUnit 5.1.3 by Sebastian Bergmann and contributors.

..........................................                        42 / 42 (100%)

Time: 299 ms, Memory: 44.01Mb

OK (42 tests, 197 assertions)

**BEHAVE**
20 features passed, 0 failed, 0 skipped
350 scenarios passed, 0 failed, 3 skipped
1203 steps passed, 0 failed, 27 skipped, 0 undefined
Took 2m31.913s

**IMPORT WITH UPDATE**
imported a small country with Wikipedia, UK postcodes, TIGER 46* and successfully applied updates to it.


lonvia:
I'm sorry to say that the current division between the main setup.php and the library isn't very clean. You are mixing different levels of abstractions and functionality, which creates a functional interdependency between the two files. We need to clearly define what the tasks of each part are.

One option: setup_functions.php is only there to provide support functions. Then it shouldn't have to know anything about $aCmdResult and its internal structure. The functions should just receive the parameters they need one by one, e,g, `import_data($aCMDResult['osm-file'], $aCMDResult['ignore-errors'])`. Parameter reading and checking would remain in setup.php.

Second option: setup_function.php contains a SetupDatabase class that keeps track of all the internal state necessary for the setup and provides the functions for each step of the setup. Provide a function that initialises the internal state from a $aCMDResult-like array. That function would do most of what prepSystem() is doing and some additional sanity checking. It should not read any of the parameters that decide which functions are executed (--all, --create-functions etc.) Control over the execution would remain with setup.php.

Third option: like option 2 but move additional functionality to setup_function.php, so that setup.php only has:
```
o = new SetupDatabase()
o.configureFromArgs($_SERVER['argv'])
o.run()
```
Of course, we still want to keep the separate public functions as they are now, so SetupDatabase() can still be used from update.php.


The second option is almost what your doing right now, just with a more tight encapsulation of the parameters. The third option has some appeal given that we might want to have all utils as only thin php wrappers at some point. But that might be left for later. I'm happy with any of the three options really.

lib/setup_functions.php
```diff
@@ -0,0 +1,965 @@
...
+        $aProcEnv = null;
+        if (isset($aDSNInfo['password']) && $aDSNInfo['password']) {
+            $aProcEnv = array_merge(array('PGPASSWORD' => $aDSNInfo['password']), $_ENV);
+        }
+    
+        $result = runWithEnv($sCreateDBCmd, $aProcEnv);
+        if ($result != 0) fail('Error executing external command: '.$sCreateDBCmd);
+    }
+    
+    if ($aCMDResult['setup-db'] || $aCMDResult['all']) {
```
lonvia: create-db and setup-db should still be provided in separate functions. It makes the code easier to read.

lib/setup_functions.php
```diff
@@ -0,0 +1,965 @@
...
+function createFunctions($aCMDResult)
+{
+    info('Create Functions');
+
+    if (!checkModulePresence()) {
+        fail('error loading nominatim.so module');
+    }
+
+    createSqlFunctions($aCMDResult);
+}
```
lonvia: I'd leave that in the part that controls the flow of the setup and have onlc createSqlFunctions() and checkModulePresence() here.

lib/setup_functions.php
```diff
@@ -0,0 +1,965 @@
...
+    $sTemplate = replaceTablespace(
+        '{ts:aux-index}',
+        CONST_Tablespace_Aux_Index,
+        $sTemplate
+    );
+    pgsqlRunScript($aCMDResult, $sTemplate, false);
+
+    // re-run the functions
+    info('Recreate Functions');
+    createSqlFunctions($aCMDResult);
```
lonvia: The rerunning should also be part of the control flow and this function should create tables only as promised.

lib/setup_functions.php
ThomasBarris: I am not sure I get what you mean with "that". I can move the info(.....) to the part that controls the flow. Then I would do it in all other functions as well. Or I could move the if to the checkModulePresenceparent function i.e. 
if !checkModulePresence()
   raise error
else
   createSqlFunctions

ThomasBarris:
Many thanks for your comments, much appreciated. I think I will go for 2. With regards to the functions that are used by update.php too, I am thinking to add the dual used functions to a master class so the setup.php (and later update.php class) can get the function by inheritance and add their own functions to their specific class. Keeping only the dual used function outside looks strange in my eyes.
However, before I do that I would like to double check with you.

ThomasBarris:
can you please let me know if the attached meet your expectations? It is the concept and needs to be continued and bug fixed etc. I just wanted to double check before I spend a lot of time to finish it. 

[Flow setup_raw.txt](https://github.com/openstreetmap/Nominatim/files/2327620/Flow.setup_raw.txt)
[Class SetupFunction_raw.txt](https://github.com/openstreetmap/Nominatim/files/2327621/Class.SetupFunction_raw.txt)

lonvia:
That looks quite good, yes. I would also pick up options like 'verbose' and  'ignore-errors' during construct() and save that in member variables. Rule of thumb: if they are used in many functions throughout the class, they should go into a member.

Regarding dual-use functions: you don't need that. Just instantiate a temporary SetupFunction() object in update.php and call create_functions() on it.

lib/setup_functions.php
lonvia: Yes, I meant the whole thing with checkModulePresence() and createSqlFunctions(). It basically means removing this createFunctions() completely from this file.

ThomasBarris:
first draft of an setup class. I have successfully imported a small extract but it requires a lot of further testings. Can you please let me know if this meets your expectations (of cause finishing it, which it isn't yet).
I have just seen that createTables and recreateFunction is not yet divided as per your request. Will do it.

utils/setupClass.php
```diff
@@ -0,0 +1,844 @@
+<?php
+
```
lonvia: A namespace would be good. t least Nominatim (see the other classes), maybe even Nominatim\Setup.

utils/update.php
```diff
@@ -99,12 +101,7 @@
         fail('pyosmium-get-changes not found or not usable');
     }
     if (!$aResult['no-update-functions']) {
-        $sSetup ='@PHP_BIN@ '. CONST_InstallPath.'/utils/setup.php';
-        $iRet = -1;
-        passthru($sSetup.' --create-functions --enable-diff-updates', $iRet);
-        if ($iRet != 0) {
-            fail('Error running setup script');
-        }
+        createFunctions($aCMDResult);
```
lonvia: This should be (new SetupFunctions(array('enable-diff-updates' => true, ...)))->createFunctions(); but I'm sure that's the part you have not gotten around yet.

ThomasBarris:
- I could not provide the two command line arguments directly in the class instantiation in the update.php required for --init-updates as per your comment since the class instantiation requires further arguments. We are creating the large command line argument array anyway for the update.php itself, I modified it a little bit and used it for the instantiation of the setup class. The array now contains "ignore-errors, enable-debug-statements, no-partitions" as these are used in the constructor but not used anywhere for the update.php. I could use a second parameter when instantiate the class and decide which member variables to set. But I am not sure this is better as it requires more code in the constructor.
- I found an extension that ensures my code is PSR-2 compatible. This made some more cosmetic edits, mainly one-line if clauses and " . " -> ".". At least the pedantic Mr. Travis is fine with my code now.
- with regards to your comment on an extra directory I am personally happy with everything that improves the structure and makes it easier to find something :)
- importing a small country with Wikipedia data, UK postcodes, reduced Tiger data, special phrases and minutely updates works
- phpunit and behave are fine

If you think I should improve something, please do not hesitate to let me know. And many thanks for your help so far.

lonvia:
> I could not provide the two command line arguments directly in the class instantiation in the update.php required for --init-updates as per your comment since the class instantiation requires further arguments. 

Then the constructor of the setup class should just provide defaults to missing arguments. We cannot just add bogus arguments to update.php as they will be visible to the user and confuse them.

> I found an extension that ensures my code is PSR-2 compatible. This made some more cosmetic edits, mainly one-line if clauses and " . " -> ".". At least the pedantic Mr. Travis is fine with my code now.

Have you tried phpcs as recommended in CONTRIBUTING.md? That is also what is executed on Travis. Please do not modify the style of code that you are not touching.

> with regards to your comment on an extra directory I am personally happy with everything that improves the structure and makes it easier to find something :)

So go ahead. ;)

ThomasBarris:
I have made the requested changes and made myself familiar with phpcs. The Setup class was moved to lib/classes. The classification into functions might be more difficult as some files serve multiple purposes. I'd like to point out that the function pgsqlRunRestoreData https://github.com/openstreetmap/Nominatim/blob/master/utils/setup.php#L857 was not used anywhere and is not included.

lonvia:
You have not changed the update argument list. Adding the create-functions and enable-diff-update arguments to update is unfortunately a no-no. And I really would prefer if you left the setup of $aCMDOptions where they are for now. It makes them a lot easier to find if they are in setup.php and update.php respectively. 

Also, `lib/classes` is not a good name, sorry. There are lots of classes under lib/ already that are not in that directory. Call it `lib/setup`.

lonvia:
I've merged this now but with some further changes to the SetupFunctions constructor, see 3afd12f977e8793f0f63a76e133d5e7ca11b0a3d.

PR #1160, opened 2018-08-29
allow nginx to serve files without php extensions

ganeshkrishnan1:
The apache config allows api calls without extension for eg /search?q=query string.
This does not work on nginx and we need to enable this via this patch


lonvia:
This looks okay but I don't really have the setup to test this. A vagrant setup with nginx might be useful.

lonvia:
It took a bit longer but eventually we get to all PRs. ;) Thanks!

PR #1181, opened 2018-09-16
Set exception handler by request format

mtmail:
For https://github.com/openstreetmap/Nominatim/issues/947

- catch any unhandled exceptions in `init-website.php`, convert existing exit() logic to use exceptions
- allow setting exception handler format (JSON, XML). Only works after an endpoint has parsed the format from the URL. Alternative would be to examine `$_GET` but we have some legacy default formats, e.g. `/reverse` it's `json`
- move exception content into template files
- new `Nominatim::DatabaseError`, a subclass which can store a SQL result object. I didn't want to write a generic `Nominatim::Error` (yet)
- depends on https://github.com/openstreetmap/Nominatim/pull/1180

For the future
- The exception handling in the status endpoint uses custom error codes (`7xx`) and should be adapted, e.g. a new exception subclass
- BDD tests could parse the error message from HTML/JSON/XML. I started but it soon got too complex
- Unclear how much extra content should go into the JSON/XML output. The HTML output has the most details, e.g. link to github repository
- While running `phpunit .` finished fine, running individual files require more module imports (`require_once(...`). Once I add that the `phpunit .` complains about `Cannot redeclare chksql()`. More OO and classes in the future can solve this

### DIFF
--- a/lib/init-website.php
+++ b/lib/init-website.php
@@ -2,6 +2,7 @@
 
 require_once('init.php');
 require_once('ParameterParser.php');
+require_once('DatabaseError.php');
 require_once(CONST_Debug ? 'DebugHtml.php' : 'DebugNone.php');
 
 /***************************************************************************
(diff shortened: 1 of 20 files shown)


lib/init-website.php
```diff
@@ -15,74 +16,51 @@ function chksql($oSql, $sMsg = 'Database request failed')
...
+
+function set_exception_handler_by_format($sFormat = 'html')
+{
+    if ($sFormat == 'json') {
+        set_exception_handler('exception_handler_json');
+    } elseif ($sFormat == 'xml') {
+        set_exception_handler('exception_handler_xml');
+    } else {
+        set_exception_handler('exception_handler_html');
+    }
```
lonvia: We have more json formats. So I'd rather do: if (html) {} elseif (xml) {} else {}.

lib/template/error-html.php
```diff
@@ -0,0 +1,59 @@
...
+
+    <h3>Stack trace</h3>
+    <pre><?php echo $exception->getTraceAsString() ?></pre>
+
+    <p>
+        If you feel this error is incorrect feel file an issue on
+        <a href="https://github.com/openstreetmap/Nominatim/issues">Github</a>.
+
+        Please include the error message above and the URL you used.
+    </p>
```
lonvia: I really wouldn't want to show details like stack trace or file positions on a user error unless debug=1 is given. It is normally simply a result of bad request parameters which the user has to correct. Giving debug details would just give the impression of a software error that needs to be reported.

lib/template/error-json.php
```diff
@@ -0,0 +1,8 @@
+<?php
+    $error = array(
+              'code' => $exception->getCode(),
+              'message' => $exception->getMessage(),
+              'details' => $exception->getFile() . '('. $exception->getLine() . ')'
```
lonvia: dto

lib/template/error-xml.php
```diff
@@ -0,0 +1,5 @@
+<error>
+    <code><?php echo $exception->getCode() ?></code>
+    <message><?php echo $exception->getMessage() ?></message>
+    <details><?php echo $exception->getFile() . '('. $exception->getLine() . ')' ?></details>
```
lonvia: dto

lib/init-website.php
mtmail: Plus the `eq_(context.response.format, fmt)` didn't catch the mismatch.

mtmail:
Rebased. Filenames, line numbers and stack only returned if `CONST_DEBUG` set. `javascript_renderData` throws error instead of custom header setting (it used to return empty content).

lonvia:
Merging #1154 has messed up this PR, I fear. Sorry about that. Can I ask you to rebase once again and resolve the conflicts?

mtmail:
No problem. Merged and rebased now.

lonvia:
I've changed the HTML error message some more. I found the word 'exception' still too technical for the non-debug view. Feel free to open another PR if you don't agree with the outcome.

PR #1188, opened 2018-09-20
copy of the Ubuntu18 vagrant setup but with nginx as webserver

mtmail:
Test environment for https://github.com/openstreetmap/Nominatim/pull/1160 Let me know if I should try to merge it somehow with the Apache one.

### DIFF
--- a/vagrant/Install-on-Ubuntu-18-nginx.sh
+++ b/vagrant/Install-on-Ubuntu-18-nginx.sh
@@ -1,0 +1,120 @@
+#!/bin/bash
+
+#
+# This is variation of Install-on-Ubuntu.sh showcasing how to use the
+# nginx webserver instead of Apache2. We might eventually merge both
+# files. Right now expect this file to become outdated/unmaintained
+# over time.
+#
+# This file lacks many comments found in Install-on-Ubuntu.sh, you
+# should check that file first to get a basic understanding.
+#
+
+# hacks for broken vagrant box
+sudo rm -f /var/lib/dpkg/lock
+sudo update-locale LANG=en_US.UTF-8
+export APT_LISTCHANGES_FRONTEND=none
+export DEBIAN_FRONTEND=noninteractive
+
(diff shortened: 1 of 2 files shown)


vagrant/Install-on-Ubuntu-18-nginx.sh
```diff
@@ -0,0 +1,217 @@
...
+#
+# hacks for broken vagrant box      #DOCS:
+sudo rm -f /var/lib/dpkg/lock       #DOCS:
+sudo update-locale LANG=en_US.UTF-8 #DOCS:
+export APT_LISTCHANGES_FRONTEND=none #DOCS:
+export DEBIAN_FRONTEND=noninteractive #DOCS:
+
+#
+# *Note:* these installation instructions are also available in executable
+#         form for use with vagrant under vagrant/Install-on-Ubuntu-18.sh.
```
lonvia: We need to decide, if we want this as part of the documentation or not. If yes, this needs to be fixed and the CMake file needs to be adapted to build it. If no, this note should be removed.

I'm leaning towards 'no'.

mtmail: We can document it in the wiki, or a gist or similar or even point to this open PR. But I see `php-fpm` as optimization for scaling which the majority, especially the average administrator, doesn't need. It adds complexity starting with the reader of documentation wondering which of two options to choose, on top of Apache vs nginx.

lonvia: As indicated, I'm fine with that but then at least the comment (line 10/11) needs to be deleted before this PR can be merged. Possibly shorten the rest of the comments as well as it will only end up going stale.

mtmail: I removed all `#DOCS` and some sections which are the same with the base installation, added an intro text saying the document is unmaintained for now.

PR #1192, opened 2018-09-26
Phrase splitting: treat multiple spaces like one

mtmail:
`new Phrase('a   b', '');` creates 8 word sets (`(a   b),(a|  b),(a|| b),(a|||b),(a||b),(a | b),(a ||b),(a  |b)`), `new Phrase('a     b', '');` even 32. With this change it's only two (`(a b),(a|b)`).

While I think more word sets won't make a different further down the search logic (either being trimmed, ignored or database returning no results for the word set) more wordsets consume unnecessary memory.

### DIFF
--- a/lib/Phrase.php
+++ b/lib/Phrase.php
@@ -23,7 +23,7 @@ class Phrase
 
     public function __construct($sPhrase, $sPhraseType)
     {
-        $this->sPhrase = trim($sPhrase);
+        $this->sPhrase = preg_replace('/  +/', ' ', trim($sPhrase));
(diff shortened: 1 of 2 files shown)


lonvia:
Every phrase goes through 'SELECT make_standard_name()' before calling `new Phrase()` on it. The former already takes care of removing duplicate spaces. So removing duplicate spaces is not necessary in the constructor (even the trimming could likely be removed but I'd need to recheck the code to be sure). If you like you can note it as a precondition that the string must be correctly trimmed/deduplicated.

mtmail:
You're right, not change needed.

PR #1194, opened 2018-09-28
adjust BDD api test cases to 2018 test database

mtmail:
I updated the testdb PDF extract, you can fetch it from https://downloads.opencagedata.com/public/nominatim-testdb-201809.pbf

Nice to see tests needed only few changes.


lonvia:
I cannot download the test data. It says 403.

Could you say something about growth in size and import time?

mtmail:
Sorry, permissions are fixed. I think it was 150MB vs now 180MB, so not much difference.

PR #1195, opened 2018-09-28
New 2018 TIGER data and moved conversion scripts

mtmail:
Downloading the TIGER data from their FTP server took me 12h, converting another several hours. It's unlikely the average server admin will choose the extra work when a preprocessed file is mentioned in the install instructions.

This PR documents how the preprocessed file got created in a new `data-sources/us-tiger/` directory and also moves the related scripts there. I'm open for suggestions on naming the directories and files.

`utils/imports.php` was only a small wrapper around the TIGER conversion script and is no longer needed. One can argue it also had the wrong name as it didn't import anything (that logic is in `utils/setup.php --import-tiger-data`).

You can fetch an updated preprocessed file on https://downloads.opencagedata.com/public/

I imported the 46* files in the nominatim api test database and updated one test case.


docs/admin/Import-and-Update.md
```diff
@@ -101,52 +101,34 @@ Note that this command downloads the phrases from the wiki link above.
...
          @define('CONST_Use_US_Tiger_Data', true);
 
-  5. Apply the new settings:
+  4. Apply the new settings:
 
 ```sh
     ./utils/setup.php --create-functions --enable-diff-updates --create-partition-functions
 ```
 
-The entire US adds about 10GB to your database.
```
lonvia: Don't remove this sentence.

data-sources/us-tiger/convert.sh
```diff
@@ -0,0 +1,50 @@
...
+
+INPATH="/tmp/tiger2018raw"
+INREGEX='_([0-9]{5})_edges.zip'
+NUM_EXPECTED=3233
+
+OUTPATH="/tmp/tiger2018parsed"
+mkdir -p "$OUTPATH"
+
+WORKPATH="/tmp/tiger2018work"
+mkdir -p "$WORKPATH"
```
lonvia: I'd like to avoid hardcoded paths. How about changing the script to take two mandatory parameters 'inpath' and 'outpath' and then create a temporary work path under outpath.

data-sources/us-tiger/convert.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
+set +x
```
lonvia: Debug option that should be removed.

data-sources/us-tiger/convert.sh
```diff
@@ -0,0 +1,50 @@
...
+WORKPATH="/tmp/tiger2018work"
+mkdir -p "$WORKPATH"
+
+
+
+INFILES=($INPATH/*.zip)
+if [[ ${#INFILES[*]} != $NUM_EXPECTED ]]; then
+    echo "Expected $NUM_EXPECTED files. Found ${#INFILES[*]}."
+    exit 1
+fi
```
lonvia: I don't think that's a good idea because a) there is no guarantee that the number does not change over the years and then the script becomes incompatible and b) sometimes one only wants to convert a couple of files for testing.

lonvia:
While you are moving stuff around for Tiger, do you mind looking into #1199? I would suggest to add a new utility for any imports and updates of external data.

docs/admin/Import-and-Update.md
mtmail: Wasn't removed, just moved a paragraph higher up.

mtmail:
Added your feedback to the script. Moved (symlink) documentation into /docs/data-sources I think deserves its own section since we don't expect admins to run this scripts often if at all.

data-sources/us-tiger/convert.sh
mtmail: removed

lonvia:
Merged manually.

PR #1216, opened 2018-10-28
enabling Universe repository

matsebas:
on previous versions of Bionic, the Universe repository isn't enabled by default.
see [Only "main" component enabled after install](https://bugs.launchpad.net/subiquity/+bug/1783129)


mtmail:
Looks good to me, @lonvia ?

lonvia:
If I read that bug report correctly, then this is a bug in bionic. So it should be (and was) fixed there, not in our repo. I don't see how that can scale otherwise.

matsebas:
Yes, your appreciation is correct, it is not a bug in Nominatim, it is a bug in bionic. Thanks.

PR #1220, opened 2018-10-31
new Phrase::MAX_SETS to limit memory use

mtmail:
Queries containing many words can exhaust the process memory which causes HTTP 500 responses. 

`
[Wed Oct 31 17:10:25.516006 2018] [php7:error] [pid 3320] [client ::1:34704] PHP Fatal error:  Allowed memory size of 209715200 bytes exhausted (tried to allocate 20480 bytes) in /home/vagrant/Nominatim/lib/Phrase.php on line 103
`

I found that to be the case after about 18 words, though of course it would also be depended on how long each word is. 

Test query, which is likely not even an address `中国人民大学 (2016年) 文学院 清华大学 (2008年) 新闻与传播学院` which gets normalized by `make_standard_word` to `herezhong guo ren min da xu 2016 nian wen xu yuan qing hua da xu 2008 nian xin wen yu chuan bo xu yuan` (24 words). Here memory is exhausted between 100.000 and 150.000 word sets.

It might be worth considering putting that into configuration for those installations that give PHP processes a higher memory limit. Or instead adjusting the MAX_DEPTH by number of initial words.

### DIFF
--- a/lib/Phrase.php
+++ b/lib/Phrase.php
@@ -10,6 +10,7 @@ namespace Nominatim;
 class Phrase
 {
     const MAX_DEPTH = 7;
+    const MAX_SETS = 100000;
 
     // Complete phrase as a string.
     private $sPhrase;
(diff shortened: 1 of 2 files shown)


lonvia:
How does the new variable achieve something different than the already existing MAX_DEPTH?

PR #1225, opened 2018-11-04
update and document GB postcode data

mtmail:
Conversion script and documentation how the `gb_postcode_data.sql` file can be recreated. I used August data and you can download it from https://downloads.opencagedata.com/public/

The old file is a couple of year old. 92,639 postcodes are new. 83,512 no longer exist.

The documentation should be symlinked into /docs I'm waiting for https://github.com/openstreetmap/Nominatim/pull/1195 to be merged first.

### DIFF
--- a/data-sources/gb-postcodes/README.md
+++ b/data-sources/gb-postcodes/README.md
@@ -1,0 +1,52 @@
+# GB Postcodes
+
+
+The server [importing instructions](https://www.nominatim.org/release-docs/latest/admin/Import-and-Update/) allow optionally download [`gb_postcode_data.sql.gz`](https://www.nominatim.org/data/gb_postcode_data.sql.gz). This document explains how the file got created.
+
+## GB vs UK
+
+GB (Great Britain) is more correct as the Ordnance Survey dataset doesn't contain postcodes from Northern Ireland.
+
+## Importing separately after the initial import
+
+If you forgot to download the file, or have a new version, you can import it separately:
+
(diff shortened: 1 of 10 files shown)


data-sources/gb-postcodes/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+GB (Great Britain) is more correct as the Ordnance Survey dataset doesn't contain postcodes from Northern Ireland.
+
+## Importing separately after the initial import
+
+If you forgot to download the file, or have a new version, you can import it separately:
+
+1. Table schema is in `/data/gb_postcode_table.sql`. 
+
+2. Run `GRANT SELECT ON TABLE gb_postcode TO "{www-user}";` (found in file `sql/tables.sql`). `{www-user}` needs to be replaced with the user the webserver (e.g. Apache2) runs under, e.g. on Ubuntu that's `www-data`.
```
lonvia: This shouldn't be needed. Nominatim always creates an empty gb_postcode table and runs the grant on it. So if you import gb_postcpde_data.sql.gz later, that shouldn't affect the permissions.

mtmail:
I tested around and there's a PHP package allowing me to convert to WGS84 lat,lon. But the table schema currently has a `geometry` column and one can't directly INSERT FROM FILE into that. I'll try adding lat,lon columns and as last step have Postgresql fill the geometry column.

lonvia:
Superseded by #1252.

PR #1237, opened 2018-11-17
match languages such as ja_rm (or any other with underscore) properly

ckquentvp:
ja_rm does nothing:
https://nominatim.openstreetmap.org/reverse.php?format=jsonv2&lat=35.6828387&lon=139.7594549&zoom=12&namedetails=1&accept-language=ja_rm

ko-Latn works as expected, for example
https://nominatim.openstreetmap.org/reverse.php?format=jsonv2&lat=35.1799528&lon=129.0752365&zoom=8&namedetails=1&accept-language=ko-Latn

### DIFF
--- a/lib/ParameterParser.php
+++ b/lib/ParameterParser.php
@@ -91,7 +91,7 @@ class ParameterParser
         $sLangString = $this->getString('accept-language', $sFallback);
 
         if ($sLangString) {
-            if (preg_match_all('/(([a-z]{1,8})(-[a-z]{1,8})?)\s*(;\s*q\s*=\s*(1|0\.[0-9]+))?/i', $sLangString, $aLanguagesParse, PREG_SET_ORDER)) {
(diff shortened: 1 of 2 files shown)


lonvia:
I really wish that would be fixed in OSM. An underscore is not a valid character in a language code. I see that the wiki recommends ja-Latn these days.

ckquentvp:
I did not realize, that this was directly an OSM issue, but I do agree, that this would be better to be fixed on that side.

Dirty legacy fix versus doing things the right way. I'm honestly not sure which is better. I'll leave it up to you.

lonvia:
There are still 90k objects in the database with name:ja_rm (vs. 15k with name:ja-Latn). It might take a long time until this tag has disappeared. And during that time we have to live with both. I don't like dirty fixes much but there is a good argument for this one.

Do you mind adding a test case for the underscore case in `test/php/Nominatim/ParameterParserTest.php`? Then this would be okay to merge from my side.

ckquentvp:
There we go.

PR #1252, opened 2018-11-28
GB postcode: new conversion script, documentation

mtmail:
Improvement/Alternative to https://github.com/openstreetmap/Nominatim/pull/1225

* `proj4php` dependency not needed
* use a temporary database. Using an existing "nominatim" database also works
* take licence text from source data
* created dump is backward compatible to what is currently used

### DIFF
--- a/data-sources/gb-postcodes/README.md
+++ b/data-sources/gb-postcodes/README.md
@@ -1,0 +1,63 @@
+# GB Postcodes
+
+
+The server [importing instructions](https://www.nominatim.org/release-docs/latest/admin/Import-and-Update/) allow optionally download [`gb_postcode_data.sql.gz`](https://www.nominatim.org/data/gb_postcode_data.sql.gz). This document explains how the file got created.
+
+## GB vs UK
+
+GB (Great Britain) is more correct as the Ordnance Survey dataset doesn't contain postcodes from Northern Ireland.
+
+## Importing separately after the initial import
+
+If you forgot to download the file, or have a new version, you can import it separately:
+
+1. Import the downloaded `gb_postcode_data.sql.gz` file.
+
+2. Run `utils/setup.php --calculate-postcodes` from the build directory. This will copy data form the `gb_postcode` table to the `location_postcodes` table.
+
+
+
+## Converting Code-Point Open data
+
+1. Download from [Code-Point® Open](https://www.ordnancesurvey.co.uk/business-and-government/products/code-point-open.html). It requires an email address where a download link will be send to.
+
+2. `unzip codepo_gb.zip`
+
+   Unpacked you'll see a directory of CSV files.
+
+   ```
+   $ more codepo_gb/Data/CSV/n.csv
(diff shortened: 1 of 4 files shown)


lonvia:
Do you have the dump made with this available on your server?

mtmail:
https://downloads.opencagedata.com/public/gb_postcode_data.201811.sql.gz

data-sources/gb-postcodes/README.md
```diff
@@ -0,0 +1,63 @@
...
+   
+   createdb $DBNAME
+   echo 'CREATE EXTENSION postgis' | psql $DBNAME
+   
+   cat data/gb_postcode_table.sql | psql $DBNAME
+   
+   cat codepo_gb/Data/CSV/*.csv | ./data-sources/gb-postcodes/convert_codepoint.php | psql $DBNAME
+   
+   cat codepo_gb/Doc/licence.txt | iconv -f iso-8859-1 -t utf-8 | sed 's/^/-- /g' > gb_postcode_data.sql
+   pg_dump -a -t gb_postcode $DBNAME | grep -v '^--' | sed 's/public.gb_postcode/gb_postcode/g' >> gb_postcode_data.sql
```
lonvia: You can't just remove the 'public' namespace here. The file won't work without it because public is actively removed from the search path a couple of lines before with

    SELECT pg_catalog.set_config('search_path', '', false);

Maybe you can convince pg_dump to leave out all these `set` commands and stuff and only issue the COPY?

data-sources/gb-postcodes/README.md
```diff
@@ -0,0 +1,63 @@
...
+   DBNAME=create_gb_postcode_file
+   
+   createdb $DBNAME
+   echo 'CREATE EXTENSION postgis' | psql $DBNAME
+   
+   cat data/gb_postcode_table.sql | psql $DBNAME
+   
+   cat codepo_gb/Data/CSV/*.csv | ./data-sources/gb-postcodes/convert_codepoint.php | psql $DBNAME
+   
+   cat codepo_gb/Doc/licence.txt | iconv -f iso-8859-1 -t utf-8 | sed 's/^/-- /g' > gb_postcode_data.sql
```
lonvia: Also pipe the license through `dos2unix` to remove the MS-DOS line endings.

data-sources/gb-postcodes/README.md
mtmail: Postgresql 11 introduced `--no-comment` which would be a bit cleaner. I was only able to get the `SET/SELECT` removed with a `grep -v -e '^SET '` (not commited).

data-sources/gb-postcodes/README.md
mtmail: Added

mtmail:
Rebased, force-pushed and uploaded a new datafile.

lonvia:
Data file updated on nominatim.org. Let's see if anybody complains about the public namespace.

PR #1255, opened 2018-11-28
new FAQ entry about place_id values

mtmail:
Common misconception to treat the `place_id` as never changing identifier of a single place, e.g. in URLs or saving for later use. With Nominatim the ids are not even the same between servers.

https://developers.google.com/places/web-service/place-id on the other hand are afaik somewhat permanent. Though I see "Place IDs may change due to large-scale updates on the Google Maps database" so they don't 100% guarantee id. Wikidata IDs are permanent afaik.

Having something like a permanent id in Nominatim would need to be solved in OSM itself, see discussion in https://wiki.openstreetmap.org/wiki/Permanent_ID

### DIFF
--- a/docs/api/Output.md
+++ b/docs/api/Output.md
@@ -46,7 +46,7 @@ a single place (for reverse) of the following format:
 
 The possible fields are:
 
- * `place_id` - reference to the Nominatim internal database ID
+ * `place_id` - reference to the Nominatim internal database ID (see notes below)
  * `osm_type`, `osm_id` - reference to the OSM object
  * `boundingbox` - area of corner coordinates
  * `lat`, `lon` - latitude and longitude of the centroid of the object
@@ -75,7 +75,7 @@ a bounding box (`bbox`).
 
 The feature list has the following fields:
 
- * `place_id` - reference to the Nominatim internal database ID
+ * `place_id` - reference to the Nominatim internal database ID (see notes below)
  * `osm_type`, `osm_id` - reference to the OSM object
  * `category`, `type` - key and value of the main OSM tag
  * `display_name` - full comma-separated address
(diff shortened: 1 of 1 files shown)


mtmail:
Makes sense. Moved to end of `Output.md` in a section so it's easier to link to.

docs/api/Output.md
```diff
@@ -220,3 +220,24 @@ as subelements with the type of the address part.
 Additional information requested with `extratags=1` and `namedetails=1` can
 be found in extra elements as sub-element of each place.
 
+
+## Notes on field values
+
+### place_id is not a persistent id
+
+The `place_id` is when a Nominatim database gets installed. A single
```
lonvia: is what? There is a verb missing in the sentence.

docs/api/Output.md
```diff
@@ -220,3 +220,24 @@ as subelements with the type of the address part.
...
+place will have a different value on another server or even when the
+same data gets re-imported. It's thus not useful to treat it as
+permanent for later use.
+
+The combination `osm_type`+`osm_id` is slighly better but remember in
```
lonvia: To be precise, the exact unique combination in Nominatim is osm_type+osm_id+class. Should we mention that here?

docs/api/Output.md
```diff
@@ -220,3 +220,24 @@ as subelements with the type of the address part.
...
+OpenStreetMap mappers can delete, split, recreate places (and those
+get a new `osm_id`), there is no link between those old and new id.
+Places can only change their meaning without changing their `osm_id`,
```
lonvia: only -> also?

docs/api/Output.md
mtmail: Added in a third paragraph.

PR #1263, opened 2018-12-11
when updating GB postcodes, also run SQL getorcreate_postcode_id

mtmail:
Without this query new postcodes won't be found.


lonvia:
Please don't skip the CI, not even on documentation changes.

PR #1293, opened 2019-01-27
Update Import-and-Update.md

jeremy-rutman:
this would have been very useful for me in my own install

### DIFF
--- a/docs/admin/Import-and-Update.md
+++ b/docs/admin/Import-and-Update.md
@@ -274,7 +274,7 @@ up-to-date, have a look at the script in
 
 #### checking that all is well 
 
-To make sure your server is serving your pages, you can do a local test lookup as follows:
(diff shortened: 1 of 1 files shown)


lonvia:
The location of the website rather depends on how the webserver is set up. The import section doesn't make any assumptions about that. It should be mentioned in the 'Setting up the Webserver' sections in the installation instructions for specific OSses though. We kind of assume there that it is obvious from the config, which is not good.

jeremy-rutman:
imho this can only help noobs such as myself .
The site http://localhost/nominatim/lookup.php   would work no matter the webserver setup, no?

lonvia:
No. The example webserver setup deliberately creates an Alias for /nominatim so that the URL above works.

lonvia:
The link is already mentioned in the Apache setup section, so closing this.

PR #1297, opened 2019-01-30
Allow to hand over additional Osm2pgsql command line arguments

Nakaner:
This enables users to add whatever command line argument for the invocation of Osm2pgsql they want, e.g. `--extra-attributes`.


lonvia:
I have now updated osm2pgsql to the current master in 3bc4b4bf9f16dc4a25f6deeff497f80ed6d2bce1. This change is no longer needed then. So closing.

If you feel like adding a test and documentation for the new feature, then that would be appreciated.

PR #1302, opened 2019-02-08
new TokenList::addAdditionalPostcodeTokens

mtmail:
Logic to find more postcode formats. Core matching logic could probably move to (subclassable?) separate postcode module.

### DIFF
--- a/lib/TokenList.php
+++ b/lib/TokenList.php
@@ -162,6 +162,114 @@ class TokenList
     }
 
     /**
+     * Inspecting unmatched tokens check if another variation of postcod exist. A
+     * variation might be another normalization (e.g. needs space) or looking for
+     * the broader/shortened postcode.
+     *
+     * @param string[] $aTokens       List of tokens to look up in the database.
+     * @param string[] $aCountryCodes List of country restrictions.
+     *
+     * @return void
+     */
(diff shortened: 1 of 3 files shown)


mtmail:
Found a bug, rebased with master.

Tested against worldwide database and I believe it fixes https://github.com/openstreetmap/Nominatim/issues/1286 and https://github.com/openstreetmap/Nominatim/issues/927

lib/TokenList.php
```diff
@@ -149,6 +149,115 @@ public function addTokensFromDB(&$oDB, &$aTokens, &$aCountryCodes, $sNormQuery,
...
+     *
+     * @param string[] $aTokens       List of tokens to look up in the database.
+     * @param string[] $aCountryCodes List of country restrictions.
+     *
+     * @return void
+     */
+    public function addAdditionalPostcodeTokens(&$aTokens, &$aCountryCodes)
+    {
+        foreach ($aTokens as $sToken) {
+            if ($sToken[0] == ' ') {
```
lonvia: You have left out the check for the token being unmatched: && !$oValidTokens->contains($sToken)

mtmail: added and added a test

lonvia:
Finally closing this. We have a list of postcode regexps now. We just need to integrate them in search somewhat differently.

PR #1304, opened 2019-02-09
pear.php.net offline. Use composer instead of pear install

mtmail:
Switch to `composer global require` in the travis-ci environment and make sure the include paths are set accordingly.


lonvia:
Jay, CI is back. Do you mind rebasing your changes on master so that we get passing CI for them? Just force push on the existing branches.

PR #1316, opened 2019-02-26
add logEnd to reverse.php, just like search.php

mtmail:
Add file logging of reverse queries, just like `search.php` and `details.php` already do.

PHP's `count(null)` returns `0`.


lonvia:
Merged with slight correction. Just count() is not quite correct here as $aPlace has exactly one result and count() returns the number of fields within the result.

PR #1318, opened 2019-03-01
replace database abstraction DB with PDO

mtmail:
Travis-ci tests pass. I see API unrelated test failures with my test database

```
  api/search/params.feature:48  coordinate search with addressdetails
  => 'Plei Ya Rê, Kon Tum province, Vietnam' != 'Plei Ya Rê, Vietnam'
  api/search/queries.feature:17  House number interpolation even
  api/search/queries.feature:32  House number interpolation odd
  => Additional address parts found: {'state': 'Hamburg'}
```

## CHANGES

* `PEAR` and `PEAR::DB` dependencies no longer required, Debian `php-pgsql`, `libpq-dev` still are. https://secure.php.net/manual/en/pdo.installation.php talks about enabling (removing the `;`) from `;extension=pdo_pgsql` in `php.ini` files, it works for me unchanged though.

* new connection string (DSN) format. The minimal string is 'pgsql:dbname=nominatim', a longer strings looks like `pgsql:dbname=db1;host=machine1;port=1234;user=john;password=secret`. Without host set the connection will be over UNIX socket.

* new `Nominatim::DB` class replacing global namespace. It throws `Nominatim::DatabaseError` on connection or query errors. Wrappers so `$connection` can be private (it isn't yet) and method names to match those of the old DB library.

* `Nominatim::DB` doesn't open a database connection on initialization. It could though if desired. So far - apart from error related tests - any call to `new()` was immediately followed by `connect()`.

* PDO handles empty result sets slightly different. For example `getCol()` used to return false when no row was found. Now it returns an empty array.

* PDO is better to at treating returned numeric values. Where currently we return sometimes return strings to users, now integers are returned.

    1. forward search
https://nominatim.openstreetmap.org/search.php?q=Monaco&format=json&limit=1 vs http://localhost/nominatim/search.php?q=Monaco&format=json&limit=1&addressdetails=1

        - `place_id`, `osm_id` no longer strings
        - bounding box coordinates, `lat`, `lon`, `importance` still float

    2. reverse search comparison
https://nominatim.openstreetmap.org/reverse.php?lat=43.73&lon=7.42&format=jsonv2&addressdetails=1 vs http://localhost/nominatim/reverse.php?lat=43.73&lon=7.42&format=jsonv2&addressdetails=1

        - `place_id`, `osm_id`, `place_rank` no longer strings
        - bounding box coordinates, `lat`, `lon`, `importance` still float

    3. detail page comparison
https://nominatim.openstreetmap.org/details.php?osmtype=R&osmid=393226&format=jsonv2 vs http://localhost/nominatim/details.php?osmtype=R&osmid=393226&format=jsonv2

        - `admin_level` no longer string

    4. With both old and new importance can be "0" (string). The database query returns an integer 0, but the abstraction libraries turn it into a string.

    5. HTML and XML output not affected

## TODO

Possible followups, to be addressed individually (separate PRs)

* Connections are persistent by default, but I didn't know how to implement the `$bNew`. Should be reviewed if we can use a singleton pattern by default.

* We can now use prepared statement handlers. We should add a new second parameter to `Nominatim::DB` query functions to allow setting values.

* `chksql()` functions are empty and should be removed. In 16 calls we set useful error messages, e.g. `chksql($oDB->getAll($sSQL), 'Could not get word tokens.');`
We should add a third parameter to the `Nominatim::DB` query functions to set an error message for exceptions.

* Class naming. `db.php` should be `DB.php`, `DatabaseError` probably `DBError`.

* `Nominatim::DB` doesn't have enough unit tests. Tests for `Nominatim::DB` and `DatabaseError` could be merged.

* Review naming of the `Nominatim::DB` query methods. Currently they follow whatever the old abstraction library used.

* add `Nominatim::DB` method to check if database exists. Currently in `SetupClass::createDB` looks for an exception, a method returning a boolean would be more readable and testable.

* cleanup `Nominatim::DB::parseDSN`. The returned field names might also be outdated. It's only called twice so easy to change the return values.

* There's two methods in `SetupClass` which do parallel (async) processing of SQL files using non-PDO postgresql methods (https://secure.php.net/manual/en/ref.pgsql.php), e.g. `pg_query, pg_send_query, pg_get_result, pg_result_status, PGSQL_COMMAND_OK, ...` Would be better to move that logic into `Nominatim::DB`. Once done we can make `$connection` a private variable.

* `StatusClass::pgExec` is no longer needed

* remove all calls to `$oDB->getLastError()`, it returns an array. Should all be handled via exceptions.

### DIFF
--- a/lib/setup/SetupClass.php
+++ b/lib/setup/SetupClass.php
@@ -40,7 +40,7 @@ class SetupFunctions
         info('module path: ' . $this->sModulePath);
 
         // parse database string
-        $this->aDSNInfo = array_filter(\DB::parseDSN(CONST_Database_DSN));
+        $this->aDSNInfo = \Nominatim\DB::parseDSN(CONST_Database_DSN);
         if (!isset($this->aDSNInfo['port'])) {
             $this->aDSNInfo['port'] = 5432;
         }
@@ -74,8 +74,15 @@ class SetupFunctions
     public function createDB()
     {
         info('Create DB');
-        $sDB = \DB::connect(CONST_Database_DSN, false);
-        if (!\PEAR::isError($sDB)) {
+        $bExists = true;
+        try {
+            $oDB = new \Nominatim\DB;
+            $oDB->connect();
+        } catch (\Nominatim\DatabaseError $e) {
+            $bExists = false;
+        }
+
+        if ($bExists) {
             fail('database already exists ('.CONST_Database_DSN.')');
         }
 
@@ -94,14 +101,15 @@ class SetupFunctions
 
     public function connect()
     {
-        $this->oDB =& getDB();
+        $this->oDB = new \Nominatim\DB();
+        $this->oDB->connect();
     }
 
     public function setupDB()
     {
         info('Setup DB');
 
-        $fPostgresVersion = getPostgresVersion($this->oDB);
+        $fPostgresVersion = $this->oDB->getPostgresVersion();
         echo 'Postgres version found: '.$fPostgresVersion."\n";
 
         if ($fPostgresVersion < 9.01) {
@@ -122,7 +130,7 @@ class SetupFunctions
         }
 
 
-        $fPostgisVersion = getPostgisVersion($this->oDB);
+        $fPostgisVersion = $this->oDB->getPostgisVersion();
         echo 'Postgis version found: '.$fPostgisVersion."\n";
 
         if ($fPostgisVersion < 2.1) {
@@ -379,6 +387,7 @@ class SetupFunctions
 
         $sSQL = 'select distinct partition from country_name';
         $aPartitions = chksql($this->oDB->getCol($sSQL));
+
         if (!$this->bNoPartitions) $aPartitions[] = 0;
         foreach ($aPartitions as $sPartition) {
             $this->pgExec('TRUNCATE location_road_'.$sPartition);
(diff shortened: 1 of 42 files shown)


lib/setup/SetupClass.php
```diff
@@ -711,7 +746,7 @@ public function drop()
         }
         foreach ($aDropTables as $sDrop) {
             if ($this->bVerbose) echo "Dropping table $sDrop\n";
-            @pg_query($this->oDB->connection, "DROP TABLE $sDrop CASCADE");
+            $this->oDB->exec('DROP TABLE $sDrop CASCADE');
```
lonvia: You can't replace " with ' here. There is a variable in the string.

lib/setup/SetupClass.php
```diff
@@ -883,8 +918,8 @@ private function runWithPgEnv($sCmd)
      */
     private function pgExec($sSQL)
     {
-        if (!pg_query($this->oDB->connection, $sSQL)) {
-            fail(pg_last_error($this->oDB->connection));
+        if ($this->oDB->exec($sSQL) === false) {
+            fail($this->oDB->getLastError());
```
lonvia: Is that even necessary? I would expect that exec() throws and which leads to fail() anyways.

lib/setup_functions.php
```diff
@@ -24,14 +24,15 @@ function checkModulePresence()
...
-    $oDB = &getDB();
-    $oResult = $oDB->query($sSQL);
+    $oDB = new \Nominatim\DB();
+    $oDB->connect();
+    $oResult = $oDB->exec($sSQL);
 
     $bResult = true;
 
-    if (PEAR::isError($oResult)) {
+    if ($oResult === false) {
```
lonvia: Here too. I'd expect exec() to throw.

settings/defaults.php
```diff
@@ -7,7 +7,8 @@
 
 // General settings
 @define('CONST_Debug', false);
-@define('CONST_Database_DSN', 'pgsql://@/nominatim'); // <driver>://<username>:<password>@<host>:<port>/<database>
+@define('CONST_Database_DSN', 'pgsql:dbname=nominatim'); // or add ;host=...;port=...;user=...;password=...
+@define('CONST_Database_DSN', 'pgsql://@/nominatim');
```
lonvia: Debugging artifact?

test/php/Nominatim/DatabaseErrorTest.php
```diff
@@ -32,13 +29,12 @@ public function testSqlObjectDump()
...
+    // {
+    //     $this->expectException(DatabaseError::class);
+    //     $this->expectExceptionMessage('My custom error message');
+    //     $this->expectExceptionCode(500);
 
-        $oDB = new \DB_Error;
-        $this->assertEquals(false, chksql($oDB, 'My custom error message'));
-    }
+    //     $this->assertEquals(false, chksql(false, 'My custom error message'));
+    // }
```
lonvia: Just remove.

lib/setup/SetupClass.php
mtmail: good catch

mtmail:
Thanks a lot @lonvia , I implemented all your feedback. `exec()` indeed throws an error, just the return values can be both false or numeric depending on the SQL query and is better checked by the caller.

I also added error handling tests in DBTest.php and the `tableExists` method. Working on some of the other TODOs would cause too many line changes and makes any PR review harder.

mtmail:
I updated vagrant installation and test environment documentation files as well.

lib/setup_functions.php
lonvia: My point was that the whole of the if statement is never reached. You need to try/catch the exec.

I've confirmed that in the meantime. Delete module/nominatim.so and watch an uncaught exception being thrown. (BTW, the exception is not very informative as the stack trace is missing for some reason, so it just reports the rethrow as a source. We might need exception handlers in init-cmd.php too.)

lib/setup/SetupClass.php
```diff
@@ -477,7 +504,11 @@ public function importTigerData()
 
         $aDBInstances = array();
         for ($i = 0; $i < $this->iInstances; $i++) {
-            $aDBInstances[$i] =& getDB(true);
+            // https://secure.php.net/manual/en/function.pg-connect.php
+            $DSN = CONST_Database_DSN;
+            $DSN = preg_replace('/^pgsql:/', '', $DSN);
+            $DSN = preg_replace('/;/', '', $DSN);
+            $aDBInstances[$i] = pg_connect($DSN, PGSQL_CONNECT_FORCE_NEW | PGSQL_CONNECT_ASYNC);
```
lonvia: This doesn't work for me. I get:
```
2019-03-06 22:01:08 == Import Tiger data
Nominatim/build-testdb/tiger/46003.sql: ERROR: no connection to the server
Nominatim/lib/lib.php:7:
string(27) "no connection to the server"
```

lib/setup/SetupClass.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Nominatim\Setup;
 
+// require_once(CONST_BasePath.'/lib/db.php');
```
lonvia: Why is this commented out?

website/details.php
```diff
@@ -141,23 +143,23 @@
 $sSQL = 'SELECT (each(name)).key,(each(name)).value FROM placex ';
 $sSQL .= "WHERE place_id = $iPlaceID ORDER BY (each(name)).key";
 $aPointDetails['aNames'] = $oDB->getAssoc($sSQL);
-if (PEAR::isError($aPointDetails['aNames'])) { // possible timeout
+if ($aPointDetails && $aPointDetails['aNames']) { // possible timeout
```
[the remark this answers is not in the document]
mtmail: Nice, less code.

utils/export.php
```diff
@@ -160,7 +159,7 @@
                 $sSQL .= 'and pa.place_id in (select place_id from place_addressline where address_place_id in ('.substr($aRow['place_ids'], 1, -1).')) ';
                 $sSQL .= 'group by postcode order by count(*) desc limit 1';
                 $sRes = $oDB->getOne($sSQL);
-                if (PEAR::IsError($sRes)) fail(pg_last_error($oDB->connection));
+                if ($sRes == false) fail($oDB->getLastError());
```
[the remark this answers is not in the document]
mtmail: All removed. Adding a note to remove all getLastError() calls because it returns an array, well currently there is only one call.

lib/setup/SetupClass.php
mtmail: First `;` had to be replaced with space instead of empty string to make it a valid DSN. Then a `pg_ping` after connection did the trick, not sure why. Looking forward to moving the logic into tested methods in db.php

lib/setup/SetupClass.php
mtmail: simply forgot

PR #1324, opened 2019-03-05
remove polygon=1 (polypoints) feature

mtmail:
Removes the `&polygon=1` output feature.

Documentation on https://wiki.openstreetmap.org/wiki/Nominatim#Parameters says 'deprecated' since 6 August 2013‎.

### DIFF
--- a/lib/Geocode.php
+++ b/lib/Geocode.php
@@ -245,7 +245,6 @@ class Geocode
         }
 
         $this->oPlaceLookup->loadParamArray($oParams, $sForceGeometryType);
-        $this->oPlaceLookup->setIncludePolygonAsPoints($oParams->getBool('polygon'));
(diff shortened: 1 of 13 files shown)


mtmail:
Now also returns HTTP 400 (with tests) when `polygon=1` parameter got set

lonvia:
I think it's time to go forward with this, although I still expect quite a few complains once it gets deployed on the OSMF server. There are still quite a few requests with this parameter. Any suggestion where we could announce the change? I don't think there are many people on the mailing lists.

mtmail:
Not really and I haven't seen other tools announce breaking changes on mailing lists either.

lonvia:
I have now merged that as 7a94872413bbf305386322364b66bdc41c22ff4e without the check for the parameter. That means that search and reverse still work fine with the parameter, there just won't be the expected output. I suspect that a lot of consumers which still use the parameter, don't parse for the geometry anyway..

PR #1328, opened 2019-03-10
Nominatim::DB support input variables, custom error messages

mtmail:
Following up on https://github.com/openstreetmap/Nominatim/pull/1318

* renamed file `lib/db.php` => `DB.php`
* Reuse of existing connection if DB->connect is called multiple times
* remove `chksql` methods which were empty wrappers by now
* Extend DB query methods to take input values with prepared statements. I added that to a couple of calls but many large SQL strings use variables containing SQL where quoting wouldn't work. Tablenames cannot be input values.
* Extend DB query methods to take custom error messages
* Add documentation to Nominatim::DB
* More use of `tableExists()` and new `databaseExists()`


lonvia:
> * Reuse of existing connection if DB->connect is called multiple times

Can you elaborate? Persistent connections was already enabled.

mtmail:
@lonvia Previously when you ran `$oDB->connect(); $oDB->connect();` it opened a new connection (sets $this->connection again). Now it reuses the existing one.

PR #1344, opened 2019-03-26
if nameaddress_vector was {} the database queries failed

mtmail:
fixes https://github.com/openstreetmap/Nominatim/issues/1343


lonvia:
That was quick. Thank you. I've also added a BDD test for it.

PR #1368, opened 2019-04-28
Readme for GB postcodes had markdown syntax error

mtmail:
code blocks were not formatted correctly on http://nominatim.org/release-docs/develop/data-sources/GB-Postcodes/


lonvia:
This should be working according to [the mkdocs documentation](https://www.mkdocs.org/user-guide/writing-your-docs/#fenced-code-blocks). There might be something wrongly configured on the server that produces the documentation.

lonvia:
I see, the problem was the combination with lists. Still didn't quite fix the syntax because the indent for the paragraphs was wrong. Fixed that as well.

PR #1381, opened 2019-05-14
FAQ entry about managed database services

mtmail:
Related to https://github.com/openstreetmap/Nominatim/issues/1377 Has been asked twice so far.


lonvia:
I actually wonder if we should put this into the main documentation (i.e. a section about using remote databases).

PR #1387, opened 2019-05-21
Fix some minor issues in docs.

joto:


docs/admin/Import-and-Update.md
```diff
@@ -189,19 +189,25 @@ entire US adds about 10GB to your database.
...
-        ./utils/setup.php --import-tiger-data
+```sh
+    ./utils/setup.php --import-tiger-data
+```
 
   3. Enable use of the Tiger data in your `settings/local.php` by adding:
 
+```
          @define('CONST_Use_US_Tiger_Data', true);
+```
```
lonvia: Intermixing list items and code blocks does not really work as expected. We probably should get rid of the list completely to get this right. Something for another change tough. Do you mind removing those for now? The rest looks good to merge.

joto: I have reverted those changes for now.

PR #1388, opened 2019-05-21
register shutdown function to handle out-of-memory errors

mtmail:
fixes https://github.com/openstreetmap/Nominatim/issues/1385

Throwing errors inside the callback isn't allowed, calling our exception handler works.

To avoid code duplication I also tried callbacks using `call_user_func('exception_handler_'.$sOutFormat, ...` but overall it became less readable.

With the XML format generating the output can cause a second out-of-memory error, leading to HTTP 500 response with empty body. It only happens sometimes.


reddoxster:
To prevent a 2nd memory error from happening inside the shutdown handler, block a few KBs of memory using a string when the request begins and release this in shutdown handler.

PR #1398, opened 2019-06-02
Fix path for Nginx-to-php sock configuration

rlytleatrel8edto:
missing directory in second config statement.  Fix works locally


mtmail:
You're right, the service places the socket file in that path. Thanks a lot!

PR #1408, opened 2019-06-23
Command line option to specify osm2pgsql process count

Shkarface:
This command line options is useful when importing full updates with --import-file
It allows setting the --number-processes command line of osm2pgsql to use more CPU for the import process


lonvia:
The number of processes was set to 1 intentionally. The osm2pgsql import itself is not multithreaded (or rather it is but not in any way that is influenced by this parameter). The only phase that is parallel is indexing. In this phase a high number of threads tends to be counter-productive. In my experience it is better to increase the maintenance memory to make creating a single index faster than increasing the number of parallel threads.

Shkarface:
@lonvia thank you for the explanation, in my case the update was really slow and i thought this might be the issue

curiousercreative:
@lonvia issue should be closed?

lonvia:
I hadn't initially seen that we are talking about updates here and meant to look into if the comment above is true for that too.

lonvia:
So, same thing for updates. Closing.

Issue #1404, opened 2019-06-17
Internal Server Error when memory quota is exhausted


Guillaumevk:
Hello and thank you so much for your excellent work.

I get a 500 error when making this request:
- URL: https://nominatim.openstreetmap.org/search.php?q=Divyasree+Trinity+Building+Plot+No.+6+Hi+Tech+City+Layout+Survey+Number+64+Madhapur+Village+Serilingampally+Ranga+Reddy+District+Hyderabad+Telangana+500081+India
- Details: Allowed memory size of 209715200 bytes exhausted (tried to allocate 20480 bytes)

lonvia:
Indeed but #1388 is live, so I guess it didn't really fix the issue.

mtmail:
During creation of the out-of-memory response we try to allocate memory for the HTML/XML reponse and that fails. I've read one can pre-allocate during startup and release in the error handler but the release doesn't happen fast enough. I'll see if there's other work-arounds.

lonvia:
Ah, I assumed #1388 did handle that situation. I'm actually okay with sending back a 500 as well in that case.

lonvia:
Indirectly resolved in #1412 by avoiding out-of-memory conditions.

Issue #654, opened 2017-03-07
Can't handle more than 20 Chinese characters


leesiongchan:
20 Chinese characters (Success)
https://nominatim.openstreetmap.org/search?q=%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94&accept-language=en%2C*%3Bq%3D0.5&format=json

21 Chinese characters (500)
https://nominatim.openstreetmap.org/search?q=%E6%88%91%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94%E5%96%94&accept-language=en%2C*%3Bq%3D0.5&format=json

lonvia:
The number of words that are allowed in a query is intentionally limited. Chinese characters are considered one word per character. Does this limit provide a practical problem?

leesiongchan:
We faced this problem when we used `formatted_address` given by Google Maps, turns out the address is longer than expected.

lonvia:
I'd need a more concrete example where that many characters are needed in search. The word limit exists precisely to avoid that people dump random output from other programs into the openstreetmap.org instance.

alexander-kucheryavenko:
Hi @lonvia, 

Maybe here is the case where we need to handle more then 20 characters, but not 100% sure.

`中国 安徽省 宣城市 郎溪县 郎溪县人民医院 九楼 护士站 黎琪 收`
Looks like this is and address of some hospital in China, please check this case if its possible. 

Thanks.

Best regards.

mtmail:
Source seems to be https://community.travelchinaguide.com/forum2.asp?i=79271

lonvia:
Resolved in #1412.

PR #1412, opened 2019-06-29
Rework word set computation

### LINKED ISSUE #1403: No results when adding postal code to search query
"No results when adding postal code to search query" from https://github.com/openstreetmap/Nominatim/issues/650

I think I have the same problem today.

Sample: No resut with "Calle Periodista Francisco de Paula Terrón. Granada 18015" but with ""Calle Periodista Francisco de Paula Terrón. Granada " I have results:
...
"Vía peatonal Calle Periodista Francisco de Paula Terrón, Cerrillo de Maracena, Chana, Granada, Comarca de la Vega de Granada, Granada, Andalucía, 18014, España"

"Vía peatonal Calle Periodista Francisco de Paula Terrón, Cerrillo de Maracena, Chana, Granada, Comarca de la Vega de Granada, Granada, Andalucía, 18015, España"
...

Thanks

lonvia:
Switch from an recursive algorithm for computing the word sets
to an iterative one that benefits from caching intermediate
results. This considerably reduces the amount of memory needed,
so that the depth restriction can be dropped. To ensure that
the number of word sets remains manageable, only sets up to
a certain length are accepted and only a certain number of
total word sets. If word sets need to be dropped, we drop
the ones with more words per word set first.

To further reduce the number of potential word sets, the valid
tokens are looked up first and then only word sets containing
valid tokens are computed.

Fixes #1403, #1404 and #654.

### DIFF
--- a/lib/Geocode.php
+++ b/lib/Geocode.php
@@ -348,10 +348,7 @@ class Geocode
             $aNewPhraseSearches = array();
             $sPhraseType = $bIsStructured ? $oPhrase->getPhraseType() : '';
 
-            foreach ($oPhrase->getWordSets() as $iWordSet => $aWordset) {
-                // Too many permutations - too expensive
-                if ($iWordSet > 120) break;
-
+            foreach ($oPhrase->getWordSets() as $aWordset) {
                 $aWordsetSearches = $aSearches;
 
                 // Add all words from this wordset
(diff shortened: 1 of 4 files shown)


ckquentvp:
FYI: Class constant visibility modifiers are only available as of PHP 7.1.0. Docs (Installation.md) mention 7.0 to be the lowest supported.

lonvia:
Thanks for noticing. I'll fix that. I'd love to raise the version requirement to 7.1 but we can't do that before the new Debian release is out.

lonvia:
Fixed in cdc7d0fe0eb17ffbcd01d8bd80d2fd0e0b3d0c6d.

PR #1414, opened 2019-06-30
Remove more places from address ranking

lonvia:
Gets rid of the hard-coded expection for place nodes and sets the address rank generally via the address level config instead. That means only administrative boundaries are now used at that level in addresses.

For Germany also remove place=county (counties are complete at boundary level) and Regierungsbezirke (admin_level=5) because nobody would ever use them in an address or search.

### DIFF
--- a/settings/address-levels.json
+++ b/settings/address-levels.json
@@ -83,6 +83,16 @@
           "neighbourhood" : [30, 0]
       }
   }
+},
+{ "countries" : [ "de" ],
+  "tags" : {
+      "place" : {
+          "county" : [12, 0]
+      },
+      "boundary" : {
+          "administrative5" : [10, 0]
+      }
+  }
 }
 ]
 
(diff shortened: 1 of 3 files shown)


Megachip:
Seems that change thrown out whole communities (gemeinden) and Cities out of the address ranking. 
Example:
https://nominatim.openstreetmap.org/reverse?format=geocodejson&lat=51.934952&lon=12.819672&zoom=42&accept-language=en&addressdetails=1 (Here the City Zahna-Elster is missing and only the "suburb" Zahna is delivered

lonvia:
Nothing to do with this change. The municipality is there, it's just that the place node wins: https://nominatim.openstreetmap.org/details.php?place_id=24351665

lonvia:
I've opened #1643 for you where you can collect all the examples you find to be wrong and I'll have a look in due time.

PR #1421, opened 2019-07-10
Minor change on lookup endpoint doc

asantoz:
Fix documentation about lookup endpoint on output formats available on filter `format`

### DIFF
--- a/docs/api/Lookup.md
+++ b/docs/api/Lookup.md
@@ -19,7 +19,7 @@ Additional optional parameters are explained below.
 
 ### Output format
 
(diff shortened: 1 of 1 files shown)


lonvia:
Hmm, there is no reason why we shouldn't allow jsonv2 and geocodejson. @mtmail any idea why they are not enabled?

mtmail:
No reason. I'll add them this week with tests.

lonvia:
Great. I've merged this anyways to reflect the current state of affairs. Just add the formats back when they are enabled. 

@asantoz, thanks, nice find.

PR #1427, opened 2019-07-19
New readme file on how to build the documentation

mtmail:
It didn't fit into `CONTRIBUTING.md` and I wasn't sure if it should be a subpage on http://nominatim.org/release-docs/develop/ somewhere.

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -22,6 +22,7 @@ pages:
         - 'Overview' : 'develop/overview.md'
         - 'OSM Data Import' : 'develop/Import.md'
         - 'Place Ranking' : 'develop/Ranking.md'
+        - 'Documentation' : 'develop/Documentation.md'
     - 'External Data Sources':
         - 'Overview' : 'data-sources/overview.md'
         - 'US Census (Tiger)': 'data-sources/US-Tiger.md'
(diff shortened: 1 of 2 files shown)


docs/How-to-build-and-preview-documentation.md
```diff
@@ -0,0 +1,34 @@
...
+   ```
+
+   If you develop inside a Vagrant virtual machine:
+   * add port forwarding to your Vagrantfile, e.g. `config.vm.network "forwarded_port", guest: 8000, host: 8000`
+   * use `mkdocs serve --dev-addr 0.0.0.0:8000` because the default localhost
+      IP does not get forwarded.
+
+4. Push changes to the nominatim.org webserver
+
+   This step is manual and requires ssh access to the server.
```
lonvia: No it's not. The master branch is automatically deployed every night on under https://nominatim.org/release-docs/develop/ I suggest to remove point 4 completely and just mention the fact.

mtmail: removed

docs/mkdocs.yml
```diff
@@ -3,7 +3,7 @@ theme: readthedocs
 docs_dir: ${CMAKE_CURRENT_BINARY_DIR}
 site_url: http://nominatim.org
 repo_url: https://github.com/openstreetmap/Nominatim
-pages:
+nav:
```
lonvia: I know that newer versions of mkdocs are complaining but I'd like to leave that for a while to remain backwards compatible.

mtmail: sure, removed now/again

PR #1428, opened 2019-07-19
In install instruction make clear behave,phpunit,phpcs are optional

mtmail:
Fixes https://github.com/openstreetmap/Nominatim/issues/1425

Make it more clear the test suite software packages are optional. Add a link how to install PHP composer as not Debian (apt-get) package exists.

The comment inside the code block renders fine in mkdocs
image

### DIFF
--- a/vagrant/Install-on-Ubuntu-18.sh
+++ b/vagrant/Install-on-Ubuntu-18.sh
@@ -32,14 +32,18 @@ export DEBIAN_FRONTEND=noninteractive #DOCS:
                             apache2 php php-pgsql libapache2-mod-php \
                             php-intl git
 
-# If you want to run the test suite, you need to install the following
-# additional packages:
+# Optional Software
+# --------------------------------
+#
+# If you want to help with the development of Nominatim and run the
+# test suite, you need to install the following additional packages:
 
     sudo apt-get install -y python3-setuptools python3-dev python3-pip \
                             python3-psycopg2 python3-tidylib phpunit php-cgi
 
     pip3 install --user behave nose
 
+    # Requires Composer package manager https://getcomposer.org/download/
     composer global require "squizlabs/php_codesniffer=*"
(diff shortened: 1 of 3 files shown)


vagrant/Install-on-Ubuntu-18.sh
```diff
@@ -32,14 +32,18 @@ export DEBIAN_FRONTEND=noninteractive #DOCS:
...
+# test suite, you need to install the following additional packages:
 
     sudo apt-get install -y python3-setuptools python3-dev python3-pip \
                             python3-psycopg2 python3-tidylib phpunit php-cgi
 
     pip3 install --user behave nose
 
+    # Requires Composer package manager https://getcomposer.org/download/
     composer global require "squizlabs/php_codesniffer=*"
     sudo ln -s ~/.config/composer/vendor/bin/phpcs /usr/bin/
```
lonvia: Do we really need phpcs from the composer on ubuntu 18? I've been running the packaged version on Debian without any issues.

mtmail: `apt-get install php-codesniffer` installs `PHP_CodeSniffer version 3.2.3` for me. That's about    Feb 2018 https://github.com/squizlabs/PHP_CodeSniffer/releases Sometimes rules change. With composer it's easier to fix a version number, ideally using a composer configuration file https://github.com/openstreetmap/Nominatim/issues/944

lonvia: It did work fine even with the 2.7 from old Debian. At least I've never used anything else than the packaged version (and never will). I don't like forcing people into manually installed stuff when it is readily available through their distribution.

I also didn't realise during last review that we are recommending adding links into `/usr/bin`. That's a no-no as well. Add `~/.config/composer/vendor/bin/` to the PATH or, if you must, link into `/usr/local/bin`, but `/usr/bin` must not be touched.

PR #1433, opened 2019-07-21
make US postcode data to an optional download

mtmail:
Followup to https://github.com/openstreetmap/Nominatim/pull/1418#issuecomment-510181522

I placed a copy of the 2018 file to https://downloads.opencagedata.com/public/us_postcode_data.sql.gz

OSM coverage got better but this data file still suppliments 7000 missing US postcodes.


settings/defaults.php
```diff
@@ -23,12 +23,6 @@
 // term. Spaces are kept but collapsed to one standard space.
 @define('CONST_Term_Normalization_Rules', ":: NFD (); [[:Nonspacing Mark:] [:Cf:]] >;  :: lower (); [[:Punctuation:][:Space:]]+ > ' '; :: NFC ();");
 
-// Set to false to avoid importing extra postcodes for the US.
-@define('CONST_Use_Extra_US_Postcodes', true);
-/* Set to true after importing Tiger house number data for the US.
-   Note: The tables must already exist or queries will throw errors.
-         After changing this setting run ./utils/setup --create-functions
-         again. */
```
lonvia: That was a bit too much deletion.

mtmail: Indeed. Added again.

PR #1453, opened 2019-08-05
lookup endpoint returns boundingbox

mtmail:
fixes https://github.com/openstreetmap/Nominatim/issues/1449

[geocodejson specification](https://github.com/geocoders/geocodejson-spec/tree/master/draft)  doesn't list bounding box fields.


lonvia:
Many thanks.

PR #1478, opened 2019-08-27
Remove administrative arrondissements from Belgian addresses

tbertels:
"administrative7" -> [14, 0]

### DIFF
--- a/settings/address-levels.json
+++ b/settings/address-levels.json
@@ -93,6 +93,13 @@
           "administrative5" : [10, 0]
       }
   }
+},
+{ "countries" : [ "be" ],
+  "tags" : {
+      "boundary" : {
+          "administrative7" : [14, 0]
+      }
+  }
 }
 ]
 


lonvia:
Could you add a little bit of background information about these level 7 boundaries? What they are and why they shouldn't be part of addresses.

Please be aware that removing them from rank_address means not only that they disappear from the display but they are also no longer searchable as part of an address. You would still be able to find the boundary proper, say "arrondissement foo", but not in conjunction with other places, e.g. "Bar street, arrondissement foo". Is that what you intend to achieve?

tbertels:
The only "interaction" Belgians have with level 7 boundaries (administrative arrondissements) is during elections, when the voting process depends on the district (a subdivision of administrative arrondissements not mapped on osm - municipalities (a subdivision of districts) are mapped instead).
Administrative arrondissements are never part of an address. Actually not even provinces or regions, but those are useful to identify the municipality especially when there is not postcode. And people can say "I live in this province or this region" but they'll never say "I live in this arrondissement".

Yes, they shouldn't be visible at all unless specifically searched for. They currently [tend to replace the municipality in Nominatim](https://nominatim.openstreetmap.org/details.php?place_id=14170903) (admin level 8) which is very bad since administrative arrondissements names are taken from the main municipality of each arrondissement. Hence in this example [Charneux](https://fr.wikipedia.org/wiki/Charneux_(Jalhay)) isn't in Verviers but in Jalhay, that's why the postcode is 4845 and not 4800.
As you can notice on Wiki, the intro says "Charneux est un village belge situé en province de Liège (admin_level=6) dans la commune de Jalhay (admin_level=8)." The only place where the Verviers arrondissement is mentioned is the infobox.

lonvia:
Thanks for the explanation. Sounds indeed like a good candidate for removing it from the address. I've merge this now. I will also run a database update at some point but I can't promise when exactly that will happen.

tbertels:
> I've merge this now.

Thanks!
> I will also run a database update at some point but I can't promise when exactly that will happen.

That's great news!

PR #1479, opened 2019-08-27
Increase postcode address rank for Belgium

tbertels:
rank_address -> 17

### DIFF
--- a/sql/functions.sql
+++ b/sql/functions.sql
@@ -313,6 +313,12 @@ BEGIN
             rank_address := 11;
         END IF;
 
+    ELSEIF country_code = 'be' THEN
+        IF postcode ~ '^([0-9]{4})$' THEN
+            rank_search := 21;
+            rank_address := 17;
+        END IF;
+
     ELSE
         -- Guess at the postcode format and coverage (!)
         IF postcode ~ '^[A-Z0-9]{1,5}$' THEN -- Probably too short to be very local


lonvia:
Could you elaborate here as well a bit what your goal is with that change?

tbertels:
The goal is to put the postcode before the city name.
In Belgian addresses, the postcode is always after the street name and before the city name, like so:

image

UK and US addresses however put it after respectively the city and the state/province/region (current default behavior).
But the Belgian position is also in use in most (if not all) other European countries (see https://en.wikipedia.org/wiki/Address for the position in the whole world).

mtmail:
The returned `display_name` is simply the comma separated list of address components from small to large (usually country). So https://nominatim.openstreetmap.org/details.php?osmtype=W&osmid=715872770&class=building returns "13, Rue Courtejoie, Ciney, Dinant, Namur, Wallonia, 5590, Belgium".

To create an postal address you can use a separate project https://github.com/opencagedata/address-formatting It has tests for all(?) countries, observes the position of postcode vs city and housenumber vs street and does some filtering of components, .e.g. Namur or Wallonia are probably not part of what people usually write.

tbertels:
The goal here isn't to create a postal address (otherwise, province and region names wouldn't be shown) but to show useful information in order of usefulness, from the most precise to the least precise, which is the same order used for Belgian postal addresses.
The current default behavior seems to mimic US postal addresses.

mtmail:
Roughly speaking
Belgium uses: street house-number, postcode city
France uses: house-number street, postcode city
Great Britain/USA uses: house-number street, city postcode

The `display_name` is not meant to represent any country. Improved/localised ordering of components in the returned `display_name` field is better done outside of Nominatim.

Postcodes ranks can be all-over depending on country. Postcodes can be both larger than a city and smaller than a city in the same country, some postcodes represent only a single building (e.g. government building), others can be non-geographic (PO boxes, also so called postal towns).

Starting Nominatim 3.1 (https://github.com/openstreetmap/Nominatim/releases/tag/v3.1.0) postcodes are largely ignored in the address search. That comes with its own new set of issues/bugs but overall we saw an improvement in search.

tbertels:
> Great Britain/USA uses: house-number street, city postcode

[USA is actually](https://en.wikipedia.org/wiki/Address#United_States): house-number street, city state abbreviation postcode (the current default behavior).

> Starting Nominatim 3.1 (https://github.com/openstreetmap/Nominatim/releases/tag/v3.1.0) postcodes are largely ignored in the address search. That comes with its own new set of issues/bugs but overall we saw an improvement in search.

This change is not about changing the search rank but the address rank.
Unless postcodes are to be hidden from search results, the fact that they are largely ignored in address search ranking isn't relevant. The aim here is to show them in a more useful place in each search result.

> The `display_name` is not meant to represent any country. Improved/localised ordering of components in the returned `display_name` field is better done outside of Nominatim.

Some ordering has to be used anyway. Do you suggest that instead of keeping the US one and adding improved/localised ordering for other countries we should use the continental Europe one?

mtmail:
> instead of keeping the US one

Again, it's not meant to represent any country. If it does it's a unintended.

> adding improved/localised ordering for other countries

For output that's what https://github.com/opencagedata/address-formatting or a similar library can do. For a more complete solution we need a new Nominatim feature first, tracked in https://github.com/openstreetmap/Nominatim/issues/213 with some progress in https://github.com/openstreetmap/Nominatim/pull/1245

lonvia:
We can't change the address rank of postcodes right now. There are a lot of hard-coded assumptions about the rank all over the place. They need to be removed first. 

In any case, address formatting is a touchy subject. It is not yet quite clear to me if the formatting is a property of the country the address is in or the country the human in front of the computer is from. I tend to lean toward the latter. As long as we don't have a way to choose the formatting, it is better to stick to the current one-size-fits-all solution. And that it to display admin levels in descending order with the postcode right before the country.

PR #1484, opened 2019-09-01
on --drop warn on non-existing tables, dont croak

mtmail:
Closes https://github.com/openstreetmap/Nominatim/issues/1483

Tested on a local copy. Before
```
nominatim=# SELECT count(*) FROM pg_tables;
 count
-------
  1187
```
`./utils/setup.php --drop` runs and prints no errors.
After
```
nominatim=# SELECT count(*) FROM pg_tables;
 count
-------
   418
```

### DIFF
--- a/lib/setup/SetupClass.php
+++ b/lib/setup/SetupClass.php
@@ -725,9 +725,7 @@ class SetupFunctions
         }
         foreach ($aDropTables as $sDrop) {
             if ($this->bVerbose) echo "Dropping table $sDrop\n";
-            $this->oDB->exec("DROP TABLE $sDrop CASCADE");
-            // ignore warnings/errors as they might be caused by a table having
-            // been deleted already by CASCADE
+            $this->oDB->exec("DROP TABLE IF EXISTS $sDrop CASCADE");
         }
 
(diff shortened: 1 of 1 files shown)


lib/setup/SetupClass.php
```diff
@@ -724,10 +724,16 @@ public function drop()
...
-            // been deleted already by CASCADE
+            if ($this->oDB->table_exists($sDrop)) {
+                if ($this->bVerbose) echo "Dropping table $sDrop\n";
+                try {
+                    $this->oDB->exec("DROP TABLE $sDrop CASCADE");
+                } catch (\Nominatim\DatabaseError $e) {
+                    // ignore warnings/errors as they might be caused by a table having
+                    // been deleted already by CASCADE
+                    if ($this->bVerbose) warn($e->getMessage());
+                }
```
lonvia: As you are doing it right now and check if the table exists before dropping. Shouldn't there also be no errors now?

otbutz: Why not

```sql
DROP TABLE IF EXISTS $sDrop CASCADE
```

And let Postgres handle this?

mtmail: Yes, seems, easier

Issue #1436, opened 2019-07-22
Reranking needs to preserve order


geoneutrino:
uk postcodes in combination with town/street names sometimes result in results worse than postcode alone
E.g. the address of openstreetmap foundation :)
entering only postcode
CB4 0WS
results in correct location 
https://www.openstreetmap.org/search?query=CB4%200WS#map=16/52.2324/0.1584

Entering the complete address
https://www.openstreetmap.org/search?query=Cowley%20Road%20CB4%200WS%20Cambridge#map=19/52.22795/0.15669
jumps to a segment of cowley road several hundred meters away with postcode CB4 0WU
although search loop 1 finds a result with the correct code
https://nominatim.openstreetmap.org/search.php?format=xml&addressdetails=1&q=Cowley%20Road%20CB4%200WS%20Cambridge&limit=10&debug=1

Is it possible to "boost" the priority of postcodes if given in uk addresses ?

tbertels:
It's currently hard coded here: https://github.com/openstreetmap/Nominatim/blob/b0e6fb73c6ccabbc28464e5a9910ef7bd3293e0e/sql/functions.sql#L282

tbertels:
Though I don't understand why [this German street](https://nominatim.openstreetmap.org/details.php?place_id=115956975) postcode gets an address rank of 5 instead of 11.

tbertels:
> 
> 
> Though I don't understand why [this German street](https://nominatim.openstreetmap.org/details.php?place_id=115956975) postcode gets an address rank of 5 instead of 11.

It's actually because the database needs to be updated (see https://github.com/openstreetmap/Nominatim/commit/b8f7b3cc8dfe4da04bf3d4c26cf152dc0886665d#commitcomment-34839750).

About the issue itself, you want to give more weight to UK postcodes than to street names.
According to https://en.wikipedia.org/wiki/Postcodes_in_the_United_Kingdom a UK postcode + a house number is enough to deliver the mail, meaning that the postcode is at least as precise as the street name (and is actually more precise).
So it seems indeed a good idea. A rank_search of 28 for postcodes matching ^([A-Z][A-Z]?[0-9][0-9A-Z]? [0-9][A-Z][A-Z])$ seems reasonable. For the record, buildings numbers have a search rank of 28 and names rank 29, whereas streets rank as high as 27.

geoneutrino:
good input, thanks
[...]

PR #1485, opened 2019-09-04
Increase UK postcodes search rank

tbertels:
See issue #1436


mtmail:
Have you tested the change on a local installed UK (or smaller data extract) database?

tbertels:
Unfortunately not. If someone can, it would be useful. A search rank of 27 may be enough, though I don't know what gets the higher rank if both street name (with a different postcode) and building postcode have the same rank and we search for "street name postcode".

lonvia:
This is absolutely not going to work. Ranks >= 26 are reserved for streets and POIS and must not be assigned to anything else. That is a fundamental assumption that's made all over the code.

PR #1514, opened 2019-10-03
documentation: viewbox parameter as lon,lat instead of x,y

mtmail:
Following https://github.com/openstreetmap/Nominatim/issues/1512#issuecomment-537575270 and we don't use `x`, `y` for any other parameter. Of course user might still mix up what latitude and what longitude is.


lonvia:
I would have left the x/y description and just added a note, simply because now the choice of order looks really odd.

mtmail:
closing in favor of https://github.com/openstreetmap/Nominatim/pull/1519

PR #1531, opened 2019-10-22
Separate compilation

eyusupov:
It makes it easier to install Nominatim server and PostgresSQL with module on separate servers or use separate docker images for them.


CMakeLists.txt
```diff
@@ -35,23 +35,21 @@ add_definitions(-DNOMINATIM_VERSION="${NOMINATIM_VERSION}")
 
 set(BUILD_TESTS off CACHE BOOL "Build test suite" FORCE)
```
lonvia: The situation with BUILD_TESTS is a bit difficult. The variable here is forced to off to avoid building the tests for osm2pgsql. So it actually should be 'off' in a non-configurable way. If you want the Nominatim tests to be configurable they should get a different variable.

lonvia:
I've implemented the suggested changes in #1770. Thanks getting this started, this was very helpful.

PR #1532, opened 2019-10-23
Use ExtraDataPath for country grid

eyusupov:
This directory is already printed in the warning when country grid is missing, but
the file itself was being loaded from a different dir.


lonvia:
Thanks. Now I wonder if not all the paths for downloaded files should go into extradata.

PR #1536, opened 2019-10-26
Fix partitioning handling in setup

eyusupov:
`no-partitions` flag was handled incorrectly - no partitions were created with it and only partition zero was created without it.


eyusupov:
I've read the code wrong, looks like I'm having issues due to something else

PR #1537, opened 2019-10-26
Handle zero in getCol()

eyusupov:
getCol() method did not distinguish between value 0 being returned and no more values in the result.
When running with no-parititions flag, only partition 0 is used, so a query `select distinct partition from placex` returns 0,
which getCol() mishandles and returns an empty array, resulting in no partition tables created and incorrect partition functions.


lonvia:
Good catch!

PR #1545, opened 2019-10-31
HTML page with search form when /details.php called without params

mtmail:
fixes https://github.com/openstreetmap/Nominatim/issues/1540

image


lonvia:
Perfect. Thank you.

PR #1555, opened 2019-11-05
setup: escape arguments when executing shell commands (psql, createdb)

mtmail:
Add quotes around shell parameters. Fixes https://github.com/openstreetmap/Nominatim/issues/1553


lonvia:
Not the prettiest change, mostly because the code wasn't pretty before. Looks like we could do with a small CmdLineBuilder helper class.

PR #1578, opened 2019-11-24
Update installation documentation wrt memory usage

lonvia:
* Increases minimum required RAM for planet installations to 64GB. Those who know what they are doing should still be able to import with less but they have been warned.
* Add a section with background info about time and memory usage, so people can better understand what is going on with their import.
* By default disable node cache in osm2pgsql when a flatnode file is used. This should decrease the probability that the installation process runs into out-of-memory situations.

The PR also adds a clear warning to the README that the github issue tracker is the wrong place to ask for installation support. The increase in support tickets here is simply taking too much time off the maintainers that should be better spent improving the software. This has to stop. I suggest, we create an issue template that clearly states the minimum information a ticket must contain (similar to what is already in the CONTRIBUTING file) and then close any tickets immediately that are not up to those standards.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,14 @@ https://nominatim.org/release-docs/develop/ .
 Installation
 ============
 
+**Nominatim is a complex piece of software and runs in a complex environment.
+Installing and running Nominatim is something for experienced system
+administrators only who can do some trouble-shooting themselves. We are sorry,
+but we can not provide installation support. We are all doing this in our free
+time and there is just so much of that time to go around. Do not open issues in
+our bug tracker if you need help. You can ask questions on the mailing list
+(see below) or on [help.openstreetmap.org](https://help.openstreetmap.org/).**
+
 The latest stable release can be downloaded from https://nominatim.org.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -19,6 +19,14 @@ https://nominatim.org/release-docs/develop/ .
...
 ============
 
+**Nominatim is a complex piece of software and runs in a complex environment.
+Installing and running Nominatim is something for experienced system
+administrators only who can do some trouble-shooting themselves. We are sorry,
+but we can not provide installation support. We are all doing this in our free
+time and there is just so much of that time to go around. Do not open issues in
+our bug tracker if you need help. You can ask questions on the mailing list
+(see below) or on [help.openstreetmap.org](https://help.openstreetmap.org/).**
+
```
mmd-osm: As the FAQ is a bit buried in the install section, maybe add an explicit hint to it here:

"Also, be sure to thoroughly review the FAQ section in the troubleshooting guide (https://nominatim.org/release-docs/latest/admin/Faq/)."

PR #1589, opened 2019-12-01
remove old wikidata script. See data-sources/wikipedia-wikidata/ for new process

mtmail:
Scripts from 6 years ago, last touched 2 years ago.


lonvia:
The build failure was due to the osm2pgsql update. I've reverted it for now. You can rebase and force-push if you like.

The new wikipedia script has still an issue where it computes the link counts wrong. I wonder if we should keep the old scripts around until this is fixed.

PR #1595, opened 2019-12-03
new script utils/check_import_finished.php

mtmail:
For https://github.com/openstreetmap/Nominatim/issues/1574

Both number of checks and output in case of failures can still be improved.

### DIFF
--- a/utils/check_import_finished.php
+++ b/utils/check_import_finished.php
@@ -1,0 +1,151 @@
+<?php
+
+require_once(CONST_BasePath.'/lib/init-cmd.php');
+
+$term_colors = array(
+                'green' => "\033[92m",
+                'red' => "\x1B[31m",
+                'normal' => "\033[0m"
+);
+
+$print_success = function ($message = 'OK') use ($term_colors) {
+    echo $term_colors['green'].$message.$term_colors['normal']."\n";
+};
+
+$print_fail = function ($message = 'Failed') use ($term_colors) {
+    echo $term_colors['red'].$message.$term_colors['normal']."\n";
+};
+
+
+$oDB = new Nominatim\DB;
+
+
+echo 'Checking database got created ... ';
+if ($oDB->databaseExists()) {
+    $print_success();
+} else {
+    $print_fail();
+    echo <<< END
+    Hints:
+    * Is the database server started?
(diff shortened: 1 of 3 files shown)


utils/check_import_finished.php
```diff
@@ -0,0 +1,144 @@
...
+    // sql/indices.src.sql
+    'idx_word_word_id',
+    'idx_place_addressline_address_place_id',
+    'idx_placex_rank_search',
+    'idx_placex_rank_address',
+    'idx_placex_pendingsector',
+    'idx_placex_parent_place_id',
+    'idx_placex_geometry_reverse_lookupPoint',
+    'idx_placex_geometry_reverse_lookupPolygon',
+    'idx_placex_geometry_reverse_placeNode',
```
lonvia: These three indexes have lower case for the last word. ..._lookuppoint, ..._lookuppolygon etc.

utils/check_import_finished.php
```diff
@@ -0,0 +1,144 @@
...
+        echo <<< END
+        Hints:
+        * Rerun the setup.php --create-search-indices step
+END;
+        exit(1);
+    }
+}
+
+echo 'Checking indexing status ... ';
+if ($oDB->tableExists('import_status')) {
```
lonvia: Better thing to check here: SELECT count(*) FROM placex where indexed_status > 0
That should be 0 when indexing is finished.
Also: I'd move that before the index checks as this step is before indexing during import.

mtmail: changed and moved

utils/check_import_finished.php
```diff
@@ -0,0 +1,144 @@
...
+    $print_fail();
+    echo <<< END
+    * The import didn't finish.
+    Hints:
+    * Check the output of the utils/setup.php you ran.
+    Usually the osm2pgsql step failed. Check for errors related to
+    * the file you imported not containing any places
+    * harddrive full
+    * out of memory (RAM)
+    * osm2pgsql killed by other scripts, the for consuming to much memory
```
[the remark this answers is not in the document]
mtmail: removed

lonvia:
Merged as 7db0da40ad02694b82fe9a387f277a58ca7c5fce where I've removed the accidental addition of nearby.php to the CMake script.

PR #1596, opened 2019-12-03
remove old wikidata script. See data-sources/wikipedia-wikidata/ for new process

mtmail:
Scripts from 6 years ago, last touched 2 years ago.


mtmail:
Copying @lonvia's comment from the previous PR "The new wikipedia script has still an issue where it computes the link counts wrong. I wonder if we should keep the old scripts around until this is fixed."

Do we have what's wrong documented anywhere, a new github issue maybe?

lonvia:
I think that #1604 fixed the open issue.

PR #1604, opened 2019-12-14
wikipedia: create all language pagelinkcount tables before querying them

mtmail:
The table `${language}pagelinkcount` gets created in the outer loop, but the inner loop already assumes the table for all languages exist.


lonvia:
Yep, that looks like it solves the bad statistics.

PR #1608, opened 2019-12-17
fix typo - see openstreetmap#1606

liKe2k1:
Fixed other *.style files see https://github.com/openstreetmap/Nominatim/issues/1606


liKe2k1:
see https://github.com/openstreetmap/Nominatim/issues/1606

PR #1612, opened 2019-12-18
refactor import_wiki* script for readability

mtmail:
- code reformatted for readability
- much less verbose output, downloading progress and postgresql debug/info statements were >90% of output and errors were easily missed
- automatic retry of file downloads
- move language list into shared file
- removed language `vo` (https://en.wikipedia.org/wiki/Volap%C3%BCk) It's a rarely old language with only 9000 places in the output


data-sources/wikipedia-wikidata/import_wikidata.sh
```diff
@@ -1,91 +1,270 @@
...
+        item          text,
+        instance_of   text,
+        lat           numeric(11,8),
+        lon           numeric(11,8),
+        ips_site_page text,
+        language      text
+      );" | psqlcmd
 
-# create derived tables
+for i in "${language[@]}"
```
lonvia: Shouldn't that be LANGUAGES?

mtmail: fixed now (twice even :| )

PR #1617, opened 2019-12-19
Centos Vagrant: Add CLANG=true to make

mtmail:
Apparently this fixes https://github.com/openstreetmap/Nominatim/issues/1611

Only relevant reference I found is https://github.com/powa-team/pg_stat_kcache/issues/15#issuecomment-422683516 

I recreated the vagrant VM and saw no difference (no new warnings or such).

### DIFF
--- a/vagrant/Install-on-Centos-7.sh
+++ b/vagrant/Install-on-Centos-7.sh
@@ -172,7 +172,7 @@ fi                                 #DOCS:
     mkdir build
     cd build
     cmake $USERHOME/Nominatim
-    make
+    make CLANG=true
 
 #
 # Adding SELinux Security Settings


lonvia:
I'm not convinced that this is the right solution. Shouldn't we rather be adding the missing packages, i.e. the clang stuff?

mtmail:
What we have in our vagrant file works fine though. I can't tell what our users are missing. If it's really just a `CLANG=true` that would suggest the clang stuff is already installed, just not used by default. Happy to close this, or convert into a FAQ/troubleshooting item.

lonvia:
Yes, FAQ item sounds like a better idea.

PR #1638, opened 2020-01-14
check_import_finished: check for invalid indices

mtmail:
for https://github.com/openstreetmap/Nominatim/issues/1635

I have a server were this script found a broken index so I had a test case :/


lonvia:
I added a hint that the indices must be deleted manually for now. We can remove this again when --create-index cleans up the indexes.

PR #1660, opened 2020-02-04
Vagrant install instructions for CentOS 8

mtmail:
Postgresql 12, PostGIS 3, PHP 7.2, Python 3.6

Imported Monaco and website returned results.

`dnf` replaces `yum` in RedHat Enterprise Linux 8 / Fedora 22 / CentOS 8. https://en.wikipedia.org/wiki/DNF_(software)

Postgresql gets installed into `/usr/pgsql-12/`, that's not too different from our CentOS 7 instructions (`/usr/pgsql-11/`) but I had to set paths explicitly for Nominatim CMake and `psycopg2` to find the lib and include paths.

Two lines require PHP `composer` to be installed and thus fail. That's the same in our other Ubuntu/CentOS instructions. Few developers want to run tests and for those who do it's easy to install `composer`. There's no system packages for `composer`, the recommendation is downloading a file and piping into the shell.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -61,6 +61,18 @@ Vagrant.configure("2") do |config|
       sub.vm.synced_folder ".", "/vagrant", disabled: true
   end
 
+  config.vm.define "centos8" do |sub|
+      sub.vm.box = "generic/centos8"
+      sub.vm.provision :shell do |s|
+        s.path = "vagrant/Install-on-Centos-8.sh"
+        s.privileged = false
+        s.args = "yes"
+      end
+      sub.vm.synced_folder ".", "/home/vagrant/Nominatim", disabled: true
+      sub.vm.synced_folder ".", "/vagrant", disabled: true
+  end
+
+
   config.vm.provider "virtualbox" do |vb|
     vb.gui = false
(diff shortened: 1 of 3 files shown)


lonvia:
CentOS 8 comes packaged with Postgresql 10.6. That's recent enough. We should really use this as it is bound to cause less trouble.

mtmail:
I can't find a postgis in the default package repos though. Once you use `download.postgresql.org/pub/repos/` it's the same effort to install postgresql 9.4/9.5/9.6/10x/11s/12x, postgis 2.4/2.5/3.0 combination.
Should we use postgresql 10, postgis 2.4 like in the Ubuntu install files?

lonvia:
Postgis used to be in the EPEL repository. Doesn't look like it is available in version 8. Too bad.

I'm a bit reluctant about recommending Postgis 3.0 right now as I haven't tested it at all yet. Even the OSM production servers are running on 2.5. If you are already using 3.0 in production, then it's fine to keep it. Otherwise, I'd say, we recommend 2.5 for the time being.

mtmail:
will create a new PR once ready for review

PR #1664, opened 2020-02-06
check_import_finished.php - reverse_only mode has less indices

mtmail:
Having a `REVERSE_ONLY_SETUP` configuration variable might make sense. Right now we have to guess post-install if the admin used `--reverse-only`.


lonvia:
> Right now we have to guess post-install if the admin used --reverse-only.

Checking for search_name is good. The reason why `--reverse-only` is a command line options is because it can really only be changed at import time.

PR #1666, opened 2020-02-07
Vagrant Ubuntu: psycopg2 is required

mtmail:
- `psycopg2` is now a requirement, no longer optional
- Ubuntu16 defaults to PHP 7.0, that's too old for phpunit 7, but 6 seems to still work


lonvia:
We should retire Ubuntu 16 as soon as 20 is out. It's starting to be a hassle to maintain.

PR #1672, opened 2020-02-10
FAQ entry about psycopg2 error message

mtmail:
I saw this error on Ubuntu 16 when running indexing. The stacktrace makes it looks like `nominatim.py` sets something wrong, but the better solution is upgrading the library. Background: Python 3.7 will introduce a `async` reversed keyword and the psycopg2 authors renamed the options appending an underscore.


lonvia:
#1675 fixes the nominatim script to not use async_, which makes a new FAQ entry obsolete. Thanks, nonetheless.

PR #1682, opened 2020-02-15
nominatim.py does not support -q/quiet output mode

mtmail:


lonvia:
Missed the ! in front of quiet.

PR #1702, opened 2020-03-01
Make admin boundaries inherit address rank from place nodes

### LINKED ISSUE #774: Structured Query returns wrong result: Nürnberg
This query: https://nominatim.openstreetmap.org/search.php?city=N%C3%BCrnberg&country=Deutschland 

delivers the wrong location: Nürnberg, Tann (VGem), Landkreis Rottal-Inn, Lower Bavaria, Bavaria, Germany (Hamlet)

Otherwise by using query string, the web service return the correct city:
https://nominatim.openstreetmap.org/search.php?q=N%C3%BCrnberg&country=Deutschland

Any explanation or suggestion (if the case) how to fix this?

Thanks

lonvia:
When a place node gets linked to a boundary and has an address rank defined, then the boundary inherits this address rank. This allows for such constructs as "kreisfreie Städte" where the administrative level is a county but the address classification would be whatever the place designates (i.e. city or town).

The change also changes that place tags on boundaries are taken into account for linking. This should catch those boundaries where a place tag exists but a label link is still missing. To ensure that we can recompute the correct linking, the place designation of a linked place now gets saved in 'linked_place`, so that the original place tag does not get overwritten.

Finally the algorithm for typed searches has changed slightly to not only look for places where the search rank matches the type but also the address rank. This way we will catch these reranked places.

Fixes #774.

### DIFF
--- a/lib/Geocode.php
+++ b/lib/Geocode.php
@@ -808,9 +808,7 @@ class Geocode
                         $sSQL .= 'WHERE place_id in ('.$sPlaceIds.') ';
                         $sSQL .= '  AND (';
                         $sSQL .= "         placex.rank_address between $this->iMinAddressRank and $this->iMaxAddressRank ";
-                        if (14 >= $this->iMinAddressRank && 14 <= $this->iMaxAddressRank) {
-                            $sSQL .= "     OR (extratags->'place') = 'city'";
-                        }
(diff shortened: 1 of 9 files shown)


mtmail:
Follow-up issue to create: Display `linked_place` on the details HTML page.

PR #1722, opened 2020-03-25
Add simple/structured query selector to HTML homepages

galewis2:
Required some CSS changes to be consistent. My first time in PHP-land and I'm not great at CSS/HTML, so be gentle!

### DIFF
--- a/lib/template/address-html.php
+++ b/lib/template/address-html.php
@@ -9,7 +9,7 @@
 <body id="reverse-page">
 
     <?php include(CONST_BasePath.'/lib/template/includes/html-top-navigation.php'); ?>
-<div class="top-bar">
+    <div class="top-bar">
     <form class="form-inline" role="search" accept-charset="UTF-8" action="<?php echo CONST_Website_BaseURL; ?>reverse.php">
         <div class="form-group">
             <input name="format" type="hidden" value="html">
--- a/lib/template/address-html.php
+++ b/lib/template/address-html.php
@@ -9,7 +9,7 @@
 <body id="reverse-page">
 
     <?php include(CONST_BasePath.'/lib/template/includes/html-top-navigation.php'); ?>
-
+<div class="top-bar">
     <form class="form-inline" role="search" accept-charset="UTF-8" action="<?php echo CONST_Website_BaseURL; ?>reverse.php">
         <div class="form-group">
             <input name="format" type="hidden" value="html">
@@ -64,7 +64,7 @@
             <a href="<?php echo CONST_Website_BaseURL; ?>search.php">forward search</a>
         </div>
     </form>
-
+</div>
 
     <div id="content">
 
--- a/lib/template/search-html.php
+++ b/lib/template/search-html.php
@@ -10,10 +10,60 @@
 
     <?php include(CONST_BasePath.'/lib/template/includes/html-top-navigation.php'); ?>
 
-    <form class="form-inline" role="search" accept-charset="UTF-8" action="<?php echo CONST_Website_BaseURL; ?>search.php">
-        <div class="form-group">
-            <input id="q" name="q" type="text" class="form-control input-sm" placeholder="Search" value="<?php echo htmlspecialchars($sQuery); ?>" >
+    <div class="top-bar" id="structured-query-selector">
+        <div class="search-type-link">
+            <a id="switch-to-reverse" href="<?php echo CONST_Website_BaseURL; ?>reverse.php?format=html">reverse search</a>
         </div>
+        <?php
+        if (empty($aMoreParams['q']) and empty($aMoreParams['street']) and empty($aMoreParams['city']) and empty($aMoreParams['county']) and empty($aMoreParams['state']) and empty($aMoreParams['country']) and empty($aMoreParams['postalcode'])) {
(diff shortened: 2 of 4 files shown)


galewis2:
Eg: 
image

galewis2:
image

lib/template/address-html.php
```diff
@@ -9,7 +9,7 @@
 <body id="reverse-page">
 
     <?php include(CONST_BasePath.'/lib/template/includes/html-top-navigation.php'); ?>
-
+<div class="top-bar">
```
mtmail: further indend the section to the right for consistency

lib/template/search-html.php
```diff
@@ -10,10 +10,60 @@
 
     <?php include(CONST_BasePath.'/lib/template/includes/html-top-navigation.php'); ?>
 
-    <form class="form-inline" role="search" accept-charset="UTF-8" action="<?php echo CONST_Website_BaseURL; ?>search.php">
-        <div class="form-group">
-            <input id="q" name="q" type="text" class="form-control input-sm" placeholder="Search" value="<?php echo htmlspecialchars($sQuery); ?>" >
+    <div class="top-bar" id="structured-query-selector">
```
mtmail: further indend the section to the right for consistency

lib/template/search-html.php
```diff
@@ -10,10 +10,60 @@
...
+            echo '<div class="radio-inline"><input type="radio" name="query-selector" id="structured" value="structured" checked="checked">';
+            echo '<label for="structured">structured</label></div>';
+        }
+        ?>
+
+    <form role="search" accept-charset="UTF-8" action="<?php echo CONST_Website_BaseURL; ?>search.php">
+        <div class="form-group-simple"
+        <?php
+        if (empty($aMoreParams['q']) and empty($aMoreParams['street']) and empty($aMoreParams['city']) and empty($aMoreParams['county']) and empty($aMoreParams['state']) and empty($aMoreParams['country']) and empty($aMoreParams['postalcode'])) {
+            echo "style='display:block;'";
```
mtmail: using CSS class is preferred, but I also see it would be the same amount of words so it wouldn't be much more readable or shorter

lib/template/search-html.php
```diff
@@ -10,10 +10,60 @@
...
+        }
+        elseif (empty($aMoreParams['q'])) {
+            echo "style='display:none;'";
+        }
+        ?>>
+            <input id="q" name="q" type="text" class="form-control input-sm" placeholder="Search" value="<?php echo $aMoreParams['q']; ?>" >
```
mtmail: Whenever printing (echo) a value it should go through `htmlspecialchars()` function to avoid cross-side-scripting (somebody might search for something with `<` and break the HTML otherwise).

website/js/nominatim-ui.js
```diff
@@ -20,11 +20,27 @@ function parse_and_normalize_geojson_string(raw_string){
...
+                    $("div.form-group-structured").hide();
+                    $("div.form-group-simple").show();
+                    $('.form-group-structured').find('input:text').val('');
+                }
+                else if (query_val == "structured") {
+                    $("div.form-group-simple").hide();
+                    $("div.form-group-structured").show();
+                    $('.form-group-simple').find('input:text').val('');
+                }
+           });
```
mtmail: Correct indentation, it's not immediately clear where the closing brackets belong to.

mtmail:
PHP/CSS/Javascript looks great. Not escaping output (https://www.php.net/manual/en/function.htmlspecialchars.php) is something most web development frameworks do automatically these days so it's easy to miss when data is printed (echo) directly.

otbutz:
My two cents:

* would it be possible use tab indicators instead of the radio group?
* can we switch the order of the fields? I know there is no international standard for this(at least AFAIK) but postalcode at the end seems a bit strange. Maybe order it from highest to lowest detail level (housenumber+street, postalcode, city, county, state, country)

galewis2:
> My two cents:
> 
>     * would it be possible use tab indicators instead of the radio group?
> 
>     * can we switch the order of the fields? I know there is no international standard for this(at least AFAIK) but postalcode at the end seems a bit strange. Maybe order it from highest to lowest detail level (housenumber+street, postalcode, city, county, state, country)

For the tab indicators, I wanted to keep it with the same clean style of the rest of the page. Also, I didn't think of it :)

For the order, I was following the API docs and just threw it into the order described: https://nominatim.org/release-docs/develop/api/Search/

Thinking about it now, I like the postal code at the end because if you're looking up a place, which are you more likely to know, the city, county, state, country or the exact postal code? Plus, I have found the postal code to often be wrong in OSM.

galewis2:
@mtmail I resolved everything except for the CSS comment you had above as I'm not 100% sure what you meant.

galewis2:
Is there anything further I need to do on this?

mtmail:
Sorry, this is on me to merge soon. We have the current frontent twice (the new one is https://github.com/osm-search/nominatim-ui) and I meant to apply that to both.

mtmail:
Thanks @galewis2 for your patience and contribution! I applied the same to the nominaim-ui, which we still develop separate, https://github.com/osm-search/nominatim-ui/pull/6

Issue #967, opened 2018-03-07
support tab as space delimiter


mtmail:
As reported in https://lists.openstreetmap.org/pipermail/geocoding/2018-March/001928.html

Test case:

1. curl -s 'https://nominatim.openstreetmap.org?q=berlin%20texas&format=json&limit=1'
2. curl -s 'https://nominatim.openstreetmap.org?q=berlin%09texas&format=json&limit=1'

The first is space separated and returns `Berlin, Washington County, Texas, 77832, United States of America`, the second is tab separated and returns no result.

chatelao:
Thanks for the nice textcase

chatelao:
Basically you could even include:
- LF (%10)
- HF (%11)
- FF (%12) 
- CR (%13)

AdyanAhmed99:
Hey, I'd like to work on this issue if it is still required for Nominatim. I'm new to open source, so I might need some help getting started.
I have installed Nominatim locally but haven't imported the database yet as I think it is not required for this issue.

lonvia:
You need to import something into your database to actually test your fix. We do not provide ready-made test databases. Ideally you end up with a test for this issue in `test/bdd/api/search/` but you still need to create the test database from https://www.nominatim.org/data/test/nominatim-api-testdata.pbf (see `test/README.md`).

On a general note, please do not comment on issues to ask for guidance. Use the mailing list instead, see also the [GSOC hint page](https://wiki.openstreetmap.org/wiki/User:Lonvia/GSoC_2020_Nominatim_Projects).

krahulreddy:
I found out that this issue could be fixed by editing the ParameterParser class. Is this the right way of solving this issue? If yes, I'll go ahead and send a PR.

krahulreddy:
My first attempt was by editing url-search-params.js file by editing find and replace, but this did not work. This file was not even being used(Not sure why! I commented the entire file but the search still works as earlier).

This is the fix I used for now in ParameterParser.php constructor:
`$this->aParams['q'] = str_replace('	', ' ', $this->aParams['q']);`

Also, this could be generalized over other white space characters

mtmail:
`url-search-params.js` a [polyfill](https://github.com/inexorabletash/polyfill/) for older browsers, it only loads if the browser doesn't already support the feature. All recent browsers, with exception Internet Explorer 11, do https://caniuse.com/#search=URLSearchParams

[...]

PR #1733, opened 2020-04-01
Support whitespace characters(x09-x0d) as single space

krahulreddy:
For #967:

Updated the utfasciitable such that x09-x0d(TAB, LF, VT, FF and CR) characters are mapped to x20(single space).
Also added Test scenarios to test the same.


lonvia:
Works fine for me locally. Thanks!

PR #1734, opened 2020-04-03
Added whitespace support for parseLatLon

krahulreddy:
Replaced ' ' with '\s' to support whitespaces in LatLon parsing and added few unit test cases.


lonvia:
The good news is that our Travis checks are back (they where disabled in the last weeks). The bad news is that someone broke phpcs in the mean time. Not related to your PR. I've fixed in on master. Can you please rebase and the force-push to this PR? Then Travis should be okay.

PR #1737, opened 2020-04-07
details JSON: also print isaddress addressline field

mtmail:
`details-html.php` uses `$aAddressLine['isaddress']` to set colour in the HTML table. This PR adds the field to the JSON output so nominatim-ui frontend can do the same.

image


lonvia:
This just broke the output because `isaddress` does not appear in `$aLinkedLines`.

lonvia:
Fixed in e68c1132da248d0225642987593d191f8291f412.

PR #1738, opened 2020-04-08
Obtain licence copyright from extratags

### LINKED ISSUE #1699: Different licence and copyright infomation for different data providers
Hello, 
As we need to use data from different sources, we would like to link the corresponding licence and copyright information with the place in the database. Meanwhile, we will keep the original openstreetmap licence for the original data as default. We are implementing this feature now. I am wondering if we can ask for a 'Pull request'  regarding this customization? 

Thanks a lot for your help!
Joy

joy-yyd:
The change obtains the licence and copyright content from the customized data which are imported as "extratags" of "data_licence"and "data_copyright". It is related to resolve #1699.
Thanks a lot for your review!

### DIFF
--- a/lib/PlaceLookup.php
+++ b/lib/PlaceLookup.php
@@ -25,7 +25,8 @@ class PlaceLookup
     protected $sAddressRankListSql = null;
     protected $sAllowedTypesSQLList = null;
     protected $bDeDupe = true;
-
+    protected $aLicence = 'Data © OpenStreetMap contributors, ODbL 1.0.';
+    protected $aCopyright = 'https://osm.org/copyright';
 
     public function __construct(&$oDB)
     {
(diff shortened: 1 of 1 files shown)


lib/PlaceLookup.php
```diff
@@ -421,6 +424,22 @@ public function lookup($aResults, $iMinRank = 0, $iMaxRank = 30)
 
         foreach ($aPlaces as &$aPlace) {
             $aPlace['importance'] = (float) $aPlace['importance'];
+
+            $licence_copyright_content = $this->oDB->getRow(
+                "SELECT extratags -> 'data_licence' AS licence, extratags -> 'data_copyright' AS copyright FROM placex where place_id = :placeid",
+                array(':placeid' => $aPlace['place_id'])
+            );
```
lonvia, line 431: You should enable returning extratags per default ($this->bExtratags) and then extract the information from there instead of running another query on the database.

joy-yyd:
I modify this PR into the functional change.
Thanks a lot for your review!
Joy

lonvia:
I'll close this now. The way to go with this in the future is to write a custom output format. This is already possible but currently needs hacking the code. The plan is that you will be able to put a custom formatter in your project directory.

I'm sorry this whole thing fell off the table. It's still a useful feature. Sometimes there are just other priorities.

PR #1743, opened 2020-04-10
Add JSON output for /deletable.php and /polygons.php

mtmail:


lib/template/polygons-html.php
```diff
@@ -0,0 +1,64 @@
...
+
+    <p>
+        Total number of broken polygons: <?php echo $iTotalBroken ?>.
+        Also available in <a href="<?php echo CONST_Website_BaseURL; ?>polygons.php?format=json">JSON format</a>.
+    </p>
+
+    <table class="table table-striped table-hover">
+
+<?php
+if (!$aPolygons) exit;
```
lonvia, line 22: This results in a broken html page.

mtmail: fixed

mtmail:
> Did you forget the new json format files or does this come later?

The ...-json.php files were two lines so I opted to put the `echo javascript_renderData` into the handlers directly.

PR #1760, opened 2020-04-18
Fix documentation links

krahulreddy:
Found links in the documentation that did not work. Updated them with working links.


mtmail:
thanks a lot!

PR #1773, opened 2020-04-29
Licence copyright format change

### LINKED ISSUE #1699: Different licence and copyright infomation for different data providers

joy-yyd:
The  format change obtains the licence and copyright for viewers. It is related to resolve #1699.
Thanks!


lonvia:
Same as #1738.

Issue #1677, opened 2020-02-11
Import is slow on Postgresql 12 and/or Postgis 3.0


lonvia:
The import is significantly slower on Postgresql 12 than on older versions unless a really beefy machine is used. The issue is with use of the new jit module and parallel workers. Postgres tends to overuse them. Currently I get best results with both features disabled. In the Postgresql config set:
```
random_page_cost = 1.5
jit_above_cost = -1
max_parallel_workers_per_gather = 0
```
This is similar to https://github.com/openstreetmap/osm2pgsql/issues/1045, so likely there are also some query statistics that are off. Still need to investigate which query/index might be at fault.

lonvia:
There are more problems with the query planner using wrong indexes. 2a6e8ad68e80dcda3f747ca47fa149a23d62676c changes one of the queries that make indexing of rank 30 slower by a factor of 100 or so.

lonvia:
c1ef56c870575a881ff81f409b696af0f7838a6a adds a warning against using Postgresql 12 and Postgis 3. I'll backport this to the release docs.

PR #1780, opened 2020-05-05
Add vagrant machine for Ubuntu 20.04

Simon-Will:
The instructions in
[`VAGRANT.md`](https://github.com/osm-search/Nominatim/blob/42c80893cb1038e21c05c2d0d30e3a7a20f0fa26/VAGRANT.md)
still work as before. The names of the Vagrant machines are updated so
that Ubuntu 18.04 (previously called `ubuntu`) is now called `ubuntu18`
and Ubuntu 20.04 is now called `ubuntu20`.

The version changes from Ubuntu 18.04 to Ubuntu 20.04 are:

  - Python: 3.6 to 3.8
  - Postgres: 10 to 12
  - PHP: 7.2 to 7.4

In the `apt-get`, I changed `--force` to `--allow-downgrades --allow-remove-essential --allow-change-held-packages`, because the former is deprecated. Cf. the [manpage of apt-get](http://manpages.ubuntu.com/manpages/focal/man8/apt-get.8.html)

The php module `codesniffer` was previously installed via Composer, but it is available via the Ubuntu repository, so I installed it via `apt-get` now.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -15,6 +15,15 @@ Vagrant.configure("2") do |config|
   end
 
   config.vm.define "ubuntu", primary: true do |sub|
+      sub.vm.box = "bento/ubuntu-20.04"
+      sub.vm.provision :shell do |s|
+        s.path = "vagrant/Install-on-Ubuntu-20.sh"
+        s.privileged = false
+        s.args = [checkout]
+      end
+  end
+
+  config.vm.define "ubuntu18", primary: true do |sub|
       sub.vm.box = "bento/ubuntu-18.04"
       sub.vm.provision :shell do |s|
         s.path = "vagrant/Install-on-Ubuntu-18.sh"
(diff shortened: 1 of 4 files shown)


lonvia:
Thanks, looks good.

The main issue I see with Ubuntu 20.04 is the Postgis version 3. There are still some performance issues that need investigating, see #1677. Can you add a bit fat warning about that at the top of the instructions?

lonvia:
Just noticed that the integration into the documentation is still missing. You need to add a link in `docs/admin/Installation.md` and compile the script in `docs/CMakeLists.txt`. Then you should be able to run `make doc` and view the new instructions with `mkdocs serve`.

Simon-Will:
Thanks for giving me some instructions about the docs. I hadn’t read the relevant part of the documentation, so I didn’t even really understand the strange format of the vagrant install scripts. But now I get it. I hope, this is alright now. Let me know if there is still something missing.

lonvia:
Looks all good now. The new instructions should appear on https://nominatim.org some time tonight.

Issue #1781, opened 2020-05-05
PHPUnit tests fail with PHPUnit 8


Simon-Will:
With PHP version 7.4 and PHPUnit version 8.5, when I’m in the directory `tests/php/` and execute `phpunit ./`, the following error comes up:

```
PHP Fatal error:  Declaration of Nominatim\AddressDetailsTest::setUp() must be compatible with PHPUnit\Framework\TestCase::setUp(): void in /home/vagrant/Nominatim/test/php/Nominatim/AddressDetailsTest.php on line 12
```

This is because apparently the signature of the `setUp` method changed to include an explicitly declared result type `void`. This also concerns other methods, but they are not used in Nominatim’s tests.

Update the tests so that they work with PHPUnit 8.

lonvia:
Fixed in #1782.

PR #1782, opened 2020-05-05
Make tests work with phpunit 8

Simon-Will:
This pull request has two commits. The first one directly addresses the issue in #1781, so PHPUnit 8 is running now.

The second commit updates the tests to use `assertEqualsWithDelta` which is deprecated in PHPUnit 8 and removed in PHPUnit 7.

As a result, merging this pull request will mean that PHPUnit < 7.3 will no longer work. The Vagrant machine of Ubuntu 18.04 uses PHPUnit 6, so tests will fail there.

### DIFF
--- a/docs/admin/Installation.md
+++ b/docs/admin/Installation.md
@@ -56,7 +56,7 @@ For running tests:
 
   * [behave](http://pythonhosted.org/behave/)
   * [nose](https://nose.readthedocs.io)
-  * [phpunit](https://phpunit.de)
+  * [phpunit](https://phpunit.de) >= 7.3
 
 ### Hardware
 
--- a/test/php/Nominatim/AddressDetailsTest.php
+++ b/test/php/Nominatim/AddressDetailsTest.php
@@ -9,7 +9,7 @@ require_once(CONST_BasePath.'/lib/AddressDetails.php');
 class AddressDetailsTest extends \PHPUnit\Framework\TestCase
 {
 
-    protected function setUp()
+    protected function setUp(): void
     {
         // How the fixture got created
         //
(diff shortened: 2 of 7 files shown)


mtmail:
Oh, very nice. The tests even run on travis-ci.org where we still use Xenial (Ubuntu 16) https://docs.travis-ci.com/user/reference/linux/

lonvia:
Can you adapt the Ubuntu 18 vagrant script to use the composer version, so that this still work? Don't worry about the Ubuntu 16 script. I'm tempted to just remove it.

We should also make a note about the minimum version. Maybe in `docs/admin/Installation.md`?

Simon-Will:
I added the minimum PHPUnit version (7.3) to `docs/admin/Installation.md` and used composer to install PHPUnit 8 so that the tests run successfully.

However, the strange thing was that I had to include an installation for composer because it was not installed by default. (Why should it be, after all?) Have I overlooked something here?

I chose to fix the version of the composer setup script by giving the link to a specific version and hard-coding this version’s hash in the installation script. All other methods of verifying the setup script seems shady to me because it doesn’t make much sense to download a hash from the same website as the setup script.

lonvia:
> However, the strange thing was that I had to include an installation for composer because it was not installed by default. (Why should it be, after all?) Have I overlooked something here?

Looks like it is pre-installed on travis, so we never noticed it was missing. What about the Ubuntu package of composer? Is that too old to use? If possible I would prefer to install as much as possible from packages.

Simon-Will:
Somehow I overlooked that composer is available in the Ubuntu repositories because I had expected it to be prefixed by `php-`. I was a bit confused that it wasn’t there, but then went to install it from the project’s repository. Thanks for pointing this out. I’m going to try it with the official package.

Simon-Will:
Using Ubuntu’s own composer package works of course.

lonvia:
That looks a lot less scary. ;) Thanks for testing it out.

PR #1805, opened 2020-05-25
Fix script names in README

woodpeck:


mtmail:
@woodpeck Thanks a lot!

PR #1814, opened 2020-05-30
Disable JIT and parallel workers when indexing 

### LINKED ISSUE #1677: Import is slow on Postgresql 12 and/or Postgis 3.0

lonvia:
Locally disable jit and parallel workers in the connection that
do indexing. The query planner tends to be overenthusiastic about
using JIT. But with the rather less complex queries we have, the
overhead tends to be larger than the performance gain.

Fixes #1677.

### DIFF
--- a/nominatim/nominatim.py
+++ b/nominatim/nominatim.py
@@ -124,6 +124,15 @@ class DBConnection(object):
         self.wait()
 
         self.cursor = self.conn.cursor()
+        # Disable JIT and parallel workers as they are known to cause problems.
+        # Update pg_settings instead of using SET because it does not yield
+        # errors on older versions of Postgres where the settings are not
+        # implemented.
+        self.perform(
+            """ UPDATE pg_settings SET setting = -1 WHERE name = 'jit_above_cost';
+                UPDATE pg_settings SET setting = 0 
+                   WHERE name = 'max_parallel_workers_per_gather';""")
+        self.wait()
 
     def wait(self):
         """ Block until any pending operation is done.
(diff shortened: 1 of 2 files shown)


nominatim/nominatim.py
```diff
@@ -124,6 +124,15 @@ def connect(self):
...
         self.cursor = self.conn.cursor()
+        # Disable JIT and parallel workers as they are known to cause problems.
+        # Update pg_settings instead of using SET because it does not yield
+        # errors on older versions of Postgres where the settings are not
+        # implemented.
+        self.perform(
+            """ UPDATE pg_settings SET setting = -1 WHERE name = 'jit_above_cost';
+                UPDATE pg_settings SET setting = 0 
+                   WHERE name = 'max_parallel_workers_per_gather';""")
+        self.wait()
```
anayrat, line 135: @lonvia instead of performing UPDATE on pg_settings you can do similar stuff thanks to `set_config`:
```sql
SELECT set_config('max_parallel_workers_per_gather','0',false) from pg_settings where name = 'max_parallel_workers_per_gather';
```

lonvia: This was intentionally done this way. UPDATE conveniently does nothing on older Postgresql versions where the settings do not exist. The alternative would have been to check for the Postgresql version first and that's just painful.

Issue #1097, opened 2018-07-17
Postgres automatic analyze fails to find function


otbutz:
```
2018-07-17 06:59:35.908 UTC [30641] ERROR:  function transliteration(text) does not exist at character 23
2018-07-17 06:59:35.908 UTC [30641] HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
2018-07-17 06:59:35.908 UTC [30641] QUERY:  SELECT gettokenstring(transliteration(name))
2018-07-17 06:59:35.908 UTC [30641] CONTEXT:  PL/pgSQL function public.make_standard_name(text) line 5 at assignment
        automatic analyze of table "nominatim.public.placex"
```

The following works though:
```
# psql -d nominatim -U www-data -c "SELECT transliteration('some w€êîrd text')"
 transliteration
-----------------
 some weird text
(1 row)
```
```
# psql -d nominatim -U postgres -c "SELECT transliteration('some w€êîrd text')"
 transliteration
-----------------
 some weird text
(1 row)
```

A search with Nominatim doesn't trigger any errors aswell. Seems like the analyzer process doesn't pick up the function correctly?

OS: Ubuntu 18.04
Postgres version: 10.4
Postgis version: 2.4

mtmail:
I once chased this error message but couldn't figure it out. You probably see it in your error exactly every minute. When you restart postgresql it's still every minute just slightly shifted. When you remove any traffic from the database (e.g. shut down webserver) it still happens. When you remove autovacuuming it disappears. Since searches themselves are not affected I ignore it.

lonvia:
@mtmail Did you also see this with Postgres 10.4 or also with earlier versions?

mtmail:
@lonvia 10.4. I can't remember seeing it in older versions and don't have old log files to check.

otbutz:
Done: https://www.postgresql.org/message-id/2118571382.49.1531905290326.JavaMail.tbutz%40BART

otbutz:
To quote Tome Lane:
> In recent PG releases, autovacuum executes stuff under a restrictive
search_path setting (just pg_catalog, in fact).  So you need to do
something explicit to reference stuff in other schemas; either schema
qualify the name, or add a "SET search_path" clause to the function
definition.  This is good practice anyway: what you're seeing is that this
function fails with any search_path other than the one you use by default.

User search path:
```
# psql -U www-data -d nominatim -c "SHOW search_path;"
   search_path
-----------------
[...]

PR #1818, opened 2020-06-02
Remove references to public schema

rlueckl:
Just a small PR which removes hardcoded references to the "public" schema in Postgres. The "public" schema is the default when using Postgres, so it's unnecessary to set it manually. Also, if you try to use a schema other than public the current SQLs prevent that and the import will fail.

With this PR I'd like to make the setup process more generic, so people can use a different schema.

**IMPORTANT NOTE:** `SET search_path = public, pg_catalog;` is also included in https://nominatim.org/data/country_grid.sql.gz where it should also be removed! I didn't find the source of the file in this repository, so I don't know how to do a PR for it. Should you accept this PR please also consider changing `country_grid.sql.gz` the same way as the other SQL files in this PR are changed!

### DIFF
--- a/data/country_name.sql
+++ b/data/country_name.sql
@@ -9,8 +9,6 @@ SET standard_conforming_strings = on;
 SET check_function_bodies = false;
 SET client_min_messages = warning;
 
-SET search_path = public, pg_catalog;
-
 SET default_tablespace = '';
 
 SET default_with_oids = false;
--- a/data/gb_postcode_table.sql
+++ b/data/gb_postcode_table.sql
@@ -10,8 +10,6 @@ SET check_function_bodies = false;
 SET client_min_messages = warning;
 SET escape_string_warning = off;
 
-SET search_path = public, pg_catalog;
-
 SET default_tablespace = '';
 
 SET default_with_oids = false;
--- a/data/us_postcode_table.sql
+++ b/data/us_postcode_table.sql
@@ -3,8 +3,6 @@ SET client_encoding = 'UTF8';
 SET check_function_bodies = false;
 SET client_min_messages = warning;
 
-SET search_path = public, pg_catalog;
-
 SET default_tablespace = '';
 
 SET default_with_oids = false;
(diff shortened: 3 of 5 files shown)


lonvia:
Unfortunately, it's not that simple. The reason we explicitly reference `public` in those functions is that `analyse` will fail otherwise. #1097 has the details. Cleaning out the search_path in the sql files would be okay.

rlueckl:
Sorry, I didn't see that issue. I've talked with our database admins and they see the same problem in our Postgres logs.

So basically we'd need a configurable schema, just as @otbutz wrote back in 2018.

I see that there are already some variables in the different `.sql` files which are replaced by the PHP code, so it shouldn't be too much of an issue to make the schema configurable. I think somebody with knowledge of the code should be able to do this pretty easily.

Could this be done by somebody at Nominatim?

lonvia:
It's unlikely that we add schema support unless somebody comes up with a very convincing use case.

rlueckl:
I think this use case is relevant for anybody who wants to use Nominatim in a live/production environment.

We have managed databases and we have to follow processes and take a look at security concerns. We also try to avoid to use the public schema simply because it's not good practice (it might be okay for a fire-and-forget type test). Also if you want to have multiple schemas on the same server, each with their own users, you can't really use "public".

Postgres documentation states: `there is no concept of a public schema in the SQL standard. For maximum conformance to the standard, you should not use the public schema.` (source: https://www.postgresql.org/docs/11/ddl-schemas.html#DDL-SCHEMAS-PORTABILITY).

lonvia:
Nominatim was never made to be used in a shared database. In fact, the first installation step is to create it's own. That's a deliberate design decision to minimise error propagation. If we ever get around to supporting schemas, it will be explicit by using fully qualified names everywhere.

If you remove the change in `normalization.sql`, I'm happy to merge this PR. Otherwise we can also just close it. As is it just breaks the database.

rlueckl:
We're looking for other solutions. If we have the time maybe we can provide a fully configurable schema. In the meantime I'm closing this PR, because removing `SET search_path` doesn't fix the issue in all places, it's only a half-solution and you'd still need to use the public schema.

PR #1802, opened 2020-05-18
Added setup website option

krahulreddy:
For issue #946:

**Following changes are implemented in this PR:**

- [x] Created `.env.example` with all variables exclusively used by `/website` module.
- [x] CMake builds `.env` to the build folder.
- [x] Add --setup-website option in `setup.php` to build `settings_test.php` from the `.env` file. The function setupWebsite() in `SetupClass.php` covers all the cases I came across but is still not perfect.
- [x] Updated `website.tmpl` to use `settings/settings_test.php` instead of `settings/settings.php`.
- [x] Commented out these website-only variables from `defaults.php`. Once the Further steps are implemented, we can safely dispose of `defaults.php`.
- [x] Added `lib/Config.php` class to use dotenv(). This will be used to get env variables for scripts.

**Setup instructions:**

1. Copy `.env.example` to `.env`. Make any required env changes.
2. Build nominatim.
3. Run `./utils/setup.php --setup-website` to set up `settings_test.php`.
4. Run tests if necessary.

**Further steps:**

- [ ] Replace CONST_* variables in scripts with getenv() function.
- [ ] Using overlapping(used by both scripts and website) variables.
- [ ] Renaming variables according to [this comment](https://github.com/osm-search/Nominatim/issues/946#issuecomment-476330531)
- [ ] Adapt tests according to these changes.
- [ ] Add setup instructions.

Issue #946, opened 2018-02-24
use dotenv files for PHP environment variables


mtmail:
Currently we use a PHP file `build/settings/settings.php` (for defaults) plus `local.php` (which the user can change), e.g.

```
<?php
@define('CONST_BasePath', '/home/vagrant/Nominatim');
// General settings
@define('CONST_Debug', false);
@define('CONST_Database_Web_User', 'www-data');
@define('CONST_Max_Word_Frequency', '50000');
```

During installation we create `settings.php` a template and add file paths (in the above example `/home/vagrant/Nominatim`.

https://github.com/vlucas/phpdotenv is the modern alternative.

A year ago melvyn-sopacua started working on that in https://github.com/openstreetmap/Nominatim/issues/634#issuecomment-281333425. Back then we were in the middle of changing the installation (the part that sets the file paths) in preparation for Nominatim 3.0 and we didn't want to introduce more changes.

chrisarmitage:
Hi. I'm an OSM mapping contributor and PHP Developer, and would like to help out with PHP milestone. Whats the best way to find out what need doing and the approach you want to take?

mtmail:
Thanks a lot for the offer! I'll write a more detailed plan.

If you haven't yet it helps to setup a local test environment. The `VAGRANT.md` describes how to set one up in 30 minutes. The `test/README.md` also has a lot of detail how to run tests. https://travis-ci.org/openstreetmap/Nominatim/ has the output of the most recent test runs (see `.travis.yml` what they run). We don't run the `bdd/api` tests on Travis-CI. They require a larger data import, about 11GB. It's a good second or third step but not immediately required. You will notice that more than half of the test suite is actually python (environment.py, queries.py) which in turn execute php-cgi, query.php or the import.

mtmail:
Do you have a base local Nominatim installation?

We still support PHP 5.4 because that's the version the Debian stable release installs. PHPUnit for example no longer supports 5.x releases https://phpunit.de/supported-versions.html 

Our configuration management is `define()` calls in `settings/*.php` files. Luckily we're not affected by PHP 7.3's deprecation of case-insensitive constants https://www.php.net/manual/en/migration73.deprecated.php

We'd like to move `settings/defaults.php` to a `.env.example` file. Users will need to copy it to `.env` during the installation. Optionally they make any local changes to the `.env` file, the usual is setting the database connection string or the base URL.

[...]

PR #1829, opened 2020-06-12
Added setup-website option

krahulreddy:
For issue #946:

This is a replacement for PR [1802](https://github.com/osm-search/Nominatim/pull/1802) and acts as the first step in moving to dotenv files.

**Changes implemented:**
* `utils/setup.php --setup-website` now sets up build/settings/settings-frontend.php.
* settings-frontend.php is used by website/*.php files. The website will be usable only after executing the above step.

PS: `rOutputFile` is a resource type variable. Hence, the prefix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,8 @@ script:
   - if [[ $TEST_SUITE == "tests" ]]; then phpcs --report-width=120 . ; fi
   - cd $TRAVIS_BUILD_DIR/test/php
   - if [[ $TEST_SUITE == "tests" ]]; then /usr/bin/phpunit ./ ; fi
+  - cd $TRAVIS_BUILD_DIR/build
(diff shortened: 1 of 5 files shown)


lonvia:
So this clashes with the way that the tests are executed. They injected their own settings before reading the default settings by setting the `NOMINATIM_SETTINGS` environment variable. That is what that line:

    if (file_exists(getenv('NOMINATIM_SETTINGS'))) require_once(getenv('NOMINATIM_SETTINGS'));

was for in build/settings/settings.php. For the moment it might be enough to just add this line again in the `website/*.php` scripts. In the long term we need to find a nicer mechanism. The environment variable setting is hacky.

lonvia:
I'm very sorry but merging #1840 broke this PR (the documentation, which now needs to go into docs/admin/Import.md). Do you mind rebasing on latest master? Other than that, this is all good to go.

Issue #1833, opened 2020-06-15
Error while running search indices


jonsonasMafjiozas:
2020-06-15 02:44:42 == Create Search indices
ERROR:  invalid page in block 166877 of relation base/19232/12950811
ERROR: pgsql returned with error code (3)
string(34) "pgsql returned with error code (3)"

 
 nominatim@vmi394996:~/build$./utils/check_import_finished.php                                      Checking database got created ... OK
Checking nominatim.so module installed ... OK
Checking place table ... OK
Checking indexing status ... OK
Search index creation
Checking index idx_word_word_id ... OK
Checking index idx_place_addressline_address_place_id ... OK
Checking index idx_placex_rank_search ... OK
Checking index idx_placex_rank_address ... OK
Checking index idx_placex_pendingsector ... Failed
        Hints:
        * Rerun the setup.php --create-search-indices step
nominatim@vmi394996:~/build$ setup.php --create-search-indices step
setup.php: command not found
nominatim@vmi394996:~/build$ ./utils/setup.php --create-search-indices
2020-06-15 04:16:13 == module path: /srv/nominatim/build/module
2020-06-15 04:16:13 == Create Search indices
ERROR:  relation "idx_word_word_id" already exists
ERROR: pgsql returned with error code (3)
string(34) "pgsql returned with error code (3)"
<img width="650" alt="Screen Shot 2020-06-14 at 7 20 45 PM" src="https://user-images.githubusercontent.com/15992747/84612001-28abab80-ae74-11ea-9370-4103c188e5e0.png">

mtmail:
The `invalid page in block 166877 of relation base/19232/12950811` points to a serious database issue, possibly not all data was written to disk or other hardware issue. You'd need to check the Postgresql documentation/issue tracker what exactly to do. Also check the Postgresql logfile for hints. If you want to be on the safe side delete the database and start the setup from the beginning.

`./utils/setup.php --create-search-indices` runs the SQL commands in https://github.com/osm-search/Nominatim/blob/master/sql/indices.src.sql and https://github.com/osm-search/Nominatim/blob/master/sql/indices_search.src.sql You can delete any existing index. The index names are highlighted in purple color in the github page. So:

```sql
DROP INDEX IF EXISTS idx_word_word_id;
DROP INDEX IF EXISTS idx_place_addressline_address_place_id;
```
etc. From your screenshot it's not clear at which point the database crashed. You'll need to do some testing how many indices need to be deleted first.

[...]

PR #1834, opened 2020-06-15
FAQ entry for PostgreSQL -invalid page in block-

mtmail:
As mentioned in https://github.com/osm-search/Nominatim/issues/1833#issuecomment-644141701

### DIFF
--- a/utils/check_import_finished.php
+++ b/utils/check_import_finished.php
@@ -133,7 +133,8 @@ foreach ($aExpectedIndices as $sExpectedIndex) {
         $print_fail();
         echo <<< END
         Hints:
-        * Rerun the setup.php --create-search-indices step
+        * Run './utils/setup.php --create-search-indices --ignore-errors' to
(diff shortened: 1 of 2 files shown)


utils/check_import_finished.php
```diff
@@ -133,7 +133,8 @@ function isReverseOnlyInstallation()
         $print_fail();
         echo <<< END
         Hints:
-        * Rerun the setup.php --create-search-indices step
+        * Rerun the './utils/setup.php --create-search-indices' step
+          without or possibly with '--ignore-errors' parameter.
```
lonvia, line 137: I'd simplify things and just say: rerun './utils/setup.php --create-search-indices --ignore-errors'

The `--ignore-errors` is needed to skip over indexes that already exist and at this point the chance are high that at least some of them have already been created.

mtmail: great, even easier then

Issue #575, opened 2016-11-09
isolated_dwelling or suburb wrongly included in address


boothym:
Came across a few situations where a distant isolated_dwelling or suburb is included in an address when it really shouldn't be. 

For example this street has a suburb in the name, which is 2.8km away: http://nominatim.openstreetmap.org/details.php?place_id=81662773. There is also a note here http://www.openstreetmap.org/note/777205 which was "near Halbeath, Kingseat, Fife..." in my RSS feed, when it should simply read "near Hill of Beath, Fife" as it is right beside the village node!

Another example is an address in this town, e.g.: http://nominatim.openstreetmap.org/details.php?place_id=91980307 - which includes an isolated_dwelling about 2.5km away from the street. Or this pub in the nearby village which is even further away from the isolated_dwelling: http://nominatim.openstreetmap.org/details.php?place_id=49079181

**Is there a problem with Nominatim or can these issue be solved by improving the OSM data in some way?**

lonvia:
Nominatim is notoriously bad at those. For suburbs and similar it helps if you add the approximate area (tagging it with place=suburb/hamlet etc. as well or with landuse=residential, both works as long as there is a name tag with the same name as the node). For isolated_dwelling Nominatim really should reduce the search area, they are small by definition.

boothym:
Happy to add an approx area, but is there a way to combine the two with a relation to avoid duplication?

lonvia:
If they have the same name then Nominatim will automatically combine them.

boothym:
Ok, I've added some areas and they've really helped with those searches 👍 The reason I mentioned relations is to help renderers (e.g. OpenCycleMap) from displaying two names, by connecting them together somehow.

I agree about the search area of isolated_dwelling - but you also need to look at suburbs. Where there is no area around a town/village, suburbs are included in search results almost 5km away - for example http://nominatim.openstreetmap.org/details.php?place_id=69652973

Surely reducing this to 1 or 2km would be much better?

lonvia:
Areas for isolated dwellings and suburbs have been reduced in #1836.

Note that we haven't reimported the database. That means that a) hamlets and isolated dwellings will only disappear gradually as data in OSM is updated and b) there might be some odd artefacts as we change from old to new computation for small places. There will be a reimport eventually to get to a clean state but at the moment I'd like to see if the change breaks other stuff.

PR #1836, opened 2020-06-17
Change processing of place nodes in addresses

lonvia:
Place nodes are always a bit tricky to handle when considering them for addresses because we don't know the exact extent of the place.

So far we've used a buffer around a place node to define its potential address reach. This has two problems: the buffer is so large that addresses often contain false positives (see e.g. #575)  and the buffer is really distorted when getting closer to the poles.

This PR changes the buffer here to draw a (rectangular) bounding box at a certain distance in meters. This means that we always use the same box everywhere on the planet and can make the extent much smaller. Using a box has the advantage that it is much faster to figure out if a point is within the box.

The new distances are significantly smaller than the previous ones. This will reduce the number of false positives for hamlets but there is the real danger that especially postal cities in the US come out even worse than they do now. There is a test instance with this change at http://kilda.lonvia.de/places/. Please try it out and report any results that are worse off than https://nominatim.openstreetmap.org (It does only go down to street level, so don't look for house numbers and such.)

The distance to use is determined from the search rank of the object. This means that the search ranks for places needed some cleanup. This might have an impact on the result ranking. Again, please try the demo server and report any issues you find there.

**Warning:** this change will have a major impact on the database. It is not a breaking change in the sense that the database needs to be reimported for the code to work but the search will be different.


mtmail:
For http://kilda.lonvia.de/places/reverse.php?format=html&lat=49.695122730797486&lon=10.472384691238403&zoom=18 I would expect Zeisenbronn in the address, not Einsiedelei

image

http://kilda.lonvia.de/places/reverse.php?format=html&lat=45.6397481036539&lon=10.140380859375&zoom=18 maps the street to the correct place, when I set the zoom below 16 (street-level) the nearby locality is matched. I think that's expected because Nominatim can only go by distance to center node here.

lonvia:
The first example is one of the tricky ones. It is difficult to know that 'Einsiedelei' is not a part of Zeisenbronn but both are a part of Scheinfeld. That's the information that we'd need to exclude 'Einsiedelei' and not to have 'Zeisenbronn' overwritten by Scheinfeld.

I've opened #1839 for the second example. It's unrelated but fairly easy fixable.

Issue #1652, opened 2020-01-29
Get rid of separate place_classtype_* tables


lonvia:
The `place_classtype_*` tables are there so that places can be searched by their type, for example `restaurants in Hickeroy`. They are extra tables because it is not possible to have a combined index of columns where one column is a geometry. So we copy out the place_id and geometry of places of a certain class/type into a separate table and create an index over the geometry there.

We should replace this mechanism with simple partial indexes. Todo:

*  change script `./utils/special_phrases.php`. For each class/type create and index of the form `CREATE INDEX class_type_??? ON placex USING gist(centroid) WHERE class = ?? and type = ??`.
* Find queries that use `place_classtype_*` and replace those with a query on `placex`. (Check for the existence of a index class_type_?? to find out if the combination of class and type is supported.)
* Remove updates of `place_classtype_*`
* Run performance tests to find out if the performance for queries is equal to the solution with separate tables.

mark-lester:
its 47,000+ lines, I suggest as part of this job
1. split into language specific files <country code>.csv
2. de normalise into amenity, type, [phrases...], i.e one line per amenity/type per file
3. use csv format for accessibility

lonvia:
> its 47,000+ lines

Do you mean the file emitted by `./utils/specialphrases.php --wiki-import`? That is not meant to be human readable, it's meant to be readable by Postgresql. So CSV would be rather counterproductive. The largest part is about creating search terms (the `select getorcreate_amenity` queries). This should no be touched by this issue. The issue is only about the last part where the tables are created. That needs to be replaced by creating the indexes on placex.

mark-lester:
yes sorry. i am just feeling my way.  are those phrases generated form somewhere else, I guess they must be, its a database worth. I am in the right file, but i wont do anything till i have touched base multiple times

lonvia:
`./utils/specialphrases.php` downloads them from the [OSM wiki](https://wiki.openstreetmap.org/wiki/Nominatim/Special_Phrases).

mark-lester:
yes sorry, i realized that as soon as i said it, its the input to the program i am supposed to be fixing, this wont be the last stupid thing i say, please bear with me

lonvia:
[...]

PR #1862, opened 2020-07-01
Mark lester/1652

mark-lester:
for https://github.com/osm-search/Nominatim/issues/1652

### DIFF
--- a/lib/SearchDescription.php
+++ b/lib/SearchDescription.php
@@ -5,6 +5,7 @@ namespace Nominatim;
 require_once(CONST_BasePath.'/lib/SpecialSearchOperator.php');
 require_once(CONST_BasePath.'/lib/SearchContext.php');
 require_once(CONST_BasePath.'/lib/Result.php');
+require_once(CONST_BasePath.'/lib/names.php');
 
 /**
  * Description of a single interpretation of a search query.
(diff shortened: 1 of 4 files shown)


lib/SearchDescription.php
```diff
@@ -5,6 +5,7 @@
 require_once(CONST_BasePath.'/lib/SpecialSearchOperator.php');
 require_once(CONST_BasePath.'/lib/SearchContext.php');
 require_once(CONST_BasePath.'/lib/Result.php');
+require_once(CONST_BasePath.'/lib/names.php');
```
mtmail, line 8: The one function in the module does so little, I think we don't need a module.

lib/SearchDescription.php
```diff
@@ -522,26 +523,30 @@ private function queryNearbyPoi(&$oDB, $iLimit)
...
+        $sIndexName=$this->poiIndex();
+        $bIndex = $oDB->indexExists($sIndexName);
 
-        if ($oDB->tableExists($sPoiTable)) {
-            $sSQL = 'SELECT place_id FROM '.$sPoiTable.' ct';
-            if ($this->oContext->sqlCountryList) {
-                $sSQL .= ' JOIN placex USING (place_id)';
-            }
+        if ($bIndex) {
+            $qstart=false;
```
mtmail, line 530: boolean variables should start with b, eg. `$bSomething = false`

lib/SearchDescription.php
```diff
@@ -860,14 +865,16 @@ private function queryPoiByOperator(&$oDB, $aParentIDs, $iLimit)
         if ($this->iOperator == Operator::TYPE || $this->iOperator == Operator::NEAR) {
             $sClassTable = $this->poiTable();
             $bCacheTable = $oDB->tableExists($sClassTable);
+            $sIndexName=$this->poiIndex();
```
mtmail, line 868: I'm surprised PHPCS didn't enforce spaces around the = for consistency

lib/SearchDescription.php
```diff
@@ -923,7 +930,8 @@ private function queryPoiByOperator(&$oDB, $aParentIDs, $iLimit)
                         $sSQL .= " WHERE ST_Contains('$sPlaceGeom', l.centroid)";
                     }
 
-                    $sSQL .= $this->oContext->excludeSQL(' AND l.place_id');
+//                    $sSQL .= $this->oContext->excludeSQL(' AND l.place_id');
```
mtmail, line 933: remove commented out code

lib/template/includes/introduction.php
```diff
@@ -1,4 +1,4 @@
-<h2>Welcome to Nominatim</h2>
+<h2>Welcome to Nominatim DEVELOPMENT </h2>
```
mtmail, line 1: unrelated change

lib/SearchDescription.php
```diff
@@ -522,26 +523,30 @@ private function queryNearbyPoi(&$oDB, $iLimit)
 
         $aDBResults = array();
         $sPoiTable = $this->poiTable();
+        $sIndexName=$this->poiIndex();
+        $bIndex = $oDB->indexExists($sIndexName);
```
mtmail, line 527: Checking if an index exists during every search is slow. Better to check once (`$oDB->getListOfIndices()`), e.g. in a __construct, and cache the result.

lonvia: Unfortunately, we can't do this kind of caching because PHP is stateless. So, you'd recreate the list of indexes on every request. 

Looking forward to switching to Python exactly for those kind of reasons.

nspaeth: Is Switching to Python on a visible roadmap somewhere?

lonvia:
Closing, as this turned out to be unfeasible for reasons mentioned in https://github.com/osm-search/Nominatim/issues/1652#issuecomment-700735875.

PR #1863, opened 2020-07-01
move data-sources/ directory in new git repos

mtmail:
In the documentation navigation this still leaves a top-level 'External Data Sources' with now only one sub-page.


lonvia:
I've taken the liberty to move 'External Data Sources' into the 'Development' chapter.

PR #1940, opened 2020-09-07
country_name Macedonia => North Macedonia

mtmail:
The country name changed last year https://en.wikipedia.org/wiki/North_Macedonia

https://www.openstreetmap.org/relation/53293

I also updated https://wiki.openstreetmap.org/wiki/Nominatim/Country_Codes

---

```sh
cat data/country_name.sql | sed 's/country_name/country_name_new/' > data/country_name.new.sql

cat data/country_name.new.sql | psql nominatim

psql nominatim
```

```sql
SELECT name->'name:en' FROM country_name WHERE country_code='mk';
Macedonia

UPDATE country_name SET name=(
  SELECT name FROM country_name_new WHERE country_code='mk'
) WHERE country_code='mk';

DROP TABLE country_name_new ;

SELECT name->'name:en' FROM country_name WHERE country_code='mk';
North Macedonia
```


lonvia:
We probably should update this file as a whole from time to time.

PR #1945, opened 2020-09-14
Upgrade Travis-CI from Ubuntu 18 to 20

mtmail:
Updates Travis-CI to Ubuntu 20 (focal), Postgresql 12, PostGIS 3, Python is 3.8, PHP 7.4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,19 @@
 ---
 os: linux
-dist: bionic
-language: python
-python:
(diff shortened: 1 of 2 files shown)


lonvia:
I wonder if we should have these two in parallel: one representing the oldest supported versions, one the newest. What do you think?

lonvia:
NB: I don't think we need to install/setup apache for the travis tests...

PR #1952, opened 2020-09-15
utils/country_languages.php - wiki table structure changed

mtmail:
The data structure in https://wiki.openstreetmap.org/wiki/Special:Export/Nominatim/Country_Codes changed, probably June 2018 https://wiki.openstreetmap.org/w/index.php?title=Nominatim/Country_Codes&action=history

### DIFF
--- a/utils/country_languages.php
+++ b/utils/country_languages.php
@@ -16,16 +16,62 @@ getCmdOpt($_SERVER['argv'], $aCMDOptions, $aCMDResult, true, true);
 
 include(CONST_Phrase_Config);
 
-if (true) {
-    $sURL = 'https://wiki.openstreetmap.org/wiki/Special:Export/Nominatim/Country_Codes';
-    $sWikiPageXML = file_get_contents($sURL);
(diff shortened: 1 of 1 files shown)


lonvia:
I'm tempted to revert the wiki page instead. It explicitly stated that the page is being parsed automatically. Those wiki-fiddlers had no business breaking the page.

mtmail:
We haven't noticed for two years so we don't parse it often enough. Each row being multiple lines in the export output is likely mediawiki related, not by any chance of a user.

lonvia:
On second thought it might be better to have the languages in a configuration file in this repository. It is not something that gets changed a lot and people can make PRs on these rare occasions. The main disadvantage is that we end up with separate per-project lists.

If we have a config file then we should remove the language codes from country_names.sql entirely and set them from the config file on import. That would make the entire country_languages.php script unnecessary.

lonvia:
I see that https://github.com/OpenCageData/address-formatting/blob/master/conf/country2lang.yaml has a language list as well. Given that we probably use this repo anyway, we might as well just use that one.

lonvia:
I've removed the file in the meantime and will make a PR with an internal config file soon. Closing here.

PR #1957, opened 2020-09-17
Restructure vagrant scripts and installation documentation

### LINKED ISSUE #1831: Remove webserver from the basic installation
We've always required Apache in our installation instructions. It makes the installation more complex and is not really necessary if you want to develop for Nominatim, use it for photon exports or just query locally. 

I would suggest the following changes:

* add a new rule `make serve` which starts the website via `cd build/website; php -S 127.0.0.1:8000`. This can be used for testing and development
* move all instructions related to Apache and Nginx into a dedicated 'Deployment' section

Question is what happens with the vagrant scripts. Apache should disappear from the base instructions there as well. We could split the scripts in two, one for basic setup and one for webserver setup. Incidentally, that would also get rid of the duplications between ubuntu18-apache and ubuntu18-nginx.

lonvia:
This was meant to implement #1831 but ended up being a major rewrite of the documentation. Here are the important points that have changed:

* Webserver setup has been moved to a new chapter 'Deployment',
* Added a new make rule `serve-global` that listens to 0.0.0.0. We need that when running without webserver in a vagrant VM.
* Ubuntu vagrant scripts can now be run in three variants: no webserver, apache, nginx. The default does not install a webserver. The CentOS scripts are untouched because on CentOS installing php always pulls in apache.
* Changed Vagrantfile to support the libvirt provider. I find it much more stable to use these days. @mtmail could you check if I have broken your virtualbox setup with that? In particular the shared folders may not work with the generic boxes.
* Changed default website URL to `/`, so that it is compatible with `make serve`.
* Make CentOS 8 the 'centos' vagrant script and move centos 7 to `centos7`.
* Add a doc section on setting up Nominatim for development. This includes the former 'Building the Documentation section' and the instructions how to create the test database.
* Moved the README from `/test` into the documentation as a new Testing chapter.

Fixes #1831.


mtmail:
> mtmail could you check if I have broken your virtualbox setup with that

@lonvia I recreated the VM. Seems to still work fine

PR #1971, opened 2020-09-22
Drop support for is_in tag

lonvia:
The `is_in` tag had its use in OSM in times before we had administrative boundaries and the more specific `addr:*` and `is_in:*` tags. Nowadays it does more good than harm to process it. So it is time to drop support for it.

### DIFF
--- a/settings/import-extratags.style
+++ b/settings/import-extratags.style
@@ -211,7 +211,7 @@
     }
 },
 {
-    "keys" : ["addr:*", "is_in:*", "tiger:county", "is_in"],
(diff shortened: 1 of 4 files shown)


mtmail:
Also remove from https://git.openstreetmap.org/nominatim.git/blob/HEAD:/website/taginfo.json (https://taginfo.openstreetmap.org/projects/nominatim) and https://wiki.openstreetmap.org/wiki/Nominatim/Development_overview#Is_in

lonvia:
Thanks for the hints, I've updated the documentation.

PR #1985, opened 2020-09-30
migration guide: idx_placex_geometry_reverse_lookupPoint can be dropped

mtmail:
follow-up to https://github.com/osm-search/Nominatim/pull/1984


mtmail:
I used the wrong name in the branch name. Will replace

PR #2013, opened 2020-10-14
When result without housenumber is found, continue search first in same street, and then any street around to find an address with housenumber

wimvr:
Make additional searches in wider area in case a result is found without a housenumber. Can be both limited to same street or any street around.

Fixes ##1469

### DIFF
--- a/docs/api/Reverse.md
+++ b/docs/api/Reverse.md
@@ -101,6 +101,24 @@ Simplify the output geometry before returning. The parameter is the
 tolerance in degrees with which the geometry may differ from the original
 geometry. Topology is preserved in the result. (Default: 0.0)
 
+### Extended search for housenumber
+
+* `housenumbersamestreetsearchdistance=<value>`
+
+When an address without housenumber is found initially, continue search in the same
+street with a maximum distance of <value> meter to find an address with housenumber.
+This search is only done when initial search returns a street.
+
+* `housenumbersearchdistance=<value>`
+
+When an address without housenumber is found initially, continue search with a
+maximum distance of <value> meter to find an address with housenumber (may be in
+another street).
+
+Above options can be combined, in that case first a search is done in the same street,
+if that does not result in a result with housenumber, search is done also in other
+streets.
+
 ### Other
 
 * `email=<valid email address>`
--- a/lib/DB.php
+++ b/lib/DB.php
@@ -77,6 +77,7 @@ class DB
             $stmt = $this->getQueryStatement($sSQL, $aInputVars, $sErrMessage);
             $row = $stmt->fetch();
         } catch (\PDOException $e) {
+            file_put_contents(CONST_Log_File, $e . ": " . $sSQL . "\n", FILE_APPEND | LOCK_EX);
             throw new \Nominatim\DatabaseError($sErrMessage, 500, null, $e, $sSQL);
         }
         return $row;
@@ -184,6 +185,7 @@ class DB
                 $stmt = $this->connection->query($sSQL);
             }
         } catch (\PDOException $e) {
+            file_put_contents(CONST_Log_File, $e . ": " . $sSQL . "\n", FILE_APPEND | LOCK_EX);
             throw new \Nominatim\DatabaseError($sErrMessage, 500, null, $e, $sSQL);
         }
         return $stmt;
--- a/lib/ReverseGeocode.php
+++ b/lib/ReverseGeocode.php
@@ -345,4 +345,58 @@ class ReverseGeocode
         }
         return $oResult;
     }
+
+    public function lookupWithHousenumber($fLat, $fLon, $street, $maxHousenumberSameStreetDistance, $maxHousenumberDistance) {
+        $result = NULL;
+        if( isset($street) && $maxHousenumberSameStreetDistance > 0 ) {
+            $result = $this->lookupWithHousenumberInternal($fLat, $fLon, $street, $maxHousenumberSameStreetDistance, true);
+        }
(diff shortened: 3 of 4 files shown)


lonvia:
When talking about restricting to addresses, I was really just talking about filtering out results without housenumber [here](https://github.com/osm-search/Nominatim/blob/bf4d75458c6edf1fb957a8600aa6718811bb52e3/lib/ReverseGeocode.php#L265) and [here](https://github.com/osm-search/Nominatim/blob/bf4d75458c6edf1fb957a8600aa6718811bb52e3/lib/ReverseGeocode.php#L321) (aka removing the `name is not null` part).

Note that the reverse algorithm already does what you propose: find the closest object and if it is a street, find the closest house. It just does not 'jump streets' because we decided that it is confusing when the location is directly on a street but a different street name is returned because the house next to the street is attached to the next parallel street.

wimvr:
Take for example this query:

https://nominatim.openstreetmap.org/reverse?format=xml&lat=48.1078&lon=11.5998&zoom=18&addressdetails=1

It says Pfälzer-Wald-Straße, but without a housenumber.

With my adjustment (and adding "&housenumbersamestreetsearchdistance=150" to URL), result is this:

<reversegeocode timestamp="Fri, 06 Nov 20 09:52:46 +0100" attribution="Data © OpenStreetMap contributors, ODbL 1.0. http://www.openstreetmap.org/copyright" querystring="format=xml&lat=48.1078&lon=11.5998&zoom=18&addressdetails=1&housenumbersamestreetsearchdistance=150">
<result place_id="26052454" osm_type="node" osm_id="2908166925" lat="48.107878" lon="11.5996452" boundingbox="48.107828,48.107928,11.5995952,11.5996952">73, Pfälzer-Wald-Straße, Balanstraße-West, Bezirksteil Balanstraße-West, Ramersdorf-Perlach, München, Beieren, 81549, Duitsland</result>
<addressparts>
<house_number>73</house_number>
<road>Pfälzer-Wald-Straße</road>
<neighbourhood>Balanstraße-West</neighbourhood>
<suburb>Bezirksteil Balanstraße-West</suburb>
<city_district>Ramersdorf-Perlach</city_district>
<city>München</city>
<state>Beieren</state>
<postcode>81549</postcode>
<country>Duitsland</country>
<country_code>de</country_code>
</addressparts>
</reversegeocode>

In this example, I already think the wrong street has been chosen, but don't think there is much to do about this. The thing is, that with 2 additional arguments, you can choose yourself in which range to search for an object with housenumber, either in same street as the initial result, or also other streets.

I think both arguments have their own application. Take again above coordinate. Looking at the map, I think the person was actually in Chiemgaustraße.
Use case 1: person is driving. If I'm showing the location of the person, the better answer would be Chiemgaustraße, even if that would be without a housenumber. Otherwise you'll constantly see other streets appearing, probably of smaller streets. This gets even worse when driving a highway. Of course, you won't find any housenumbers there, so returning the nearest object with housenumber would be a house next to the highway, and you would think your truck driver has gone nuts, probably having his truck soon stuck.
Use case 2: person parked his car. The Chiemgaustraße is 2,5km long, so if I'm told I can find my car in Chiemgaustraße (without number), I'll have hard time finding it back. On the other hand, Pfälzer-Wald-Straße 73 would bring me just around the corner.

As from the initial request you don't know the use case, I think right now the answer returned is the best possible. But, with additional arguments, basically providing the context of the question, you can return a better answer.

Finding the right values for the arguments might be bit complicated maybe for people. So another option would be introducing a third argument:

context=[parking|driving]

Parking would result in housenumbersearchdistance=150, while driving in housenumbersamestreetsearchdistance=300 (numbers are just an example and probably need some research first). Using the 2 long arguments gives you more freedom, while the 'context' one is for quick win.

lonvia:
Your solution works well when looking for addresses near street corners. It will give you strange results in cases, where you have unconnected parallel streets. Take this example:

https://nominatim.openstreetmap.org/ui/reverse.html?format=html&lat=48.59436841744961&lon=13.462851438945563&zoom=18

Assume you have parked you car at the bus stop 'Hochstein'. Using your PR with `housenumbersamestreetsearchdistance=100&housenumbersearchdistance=50`, the result gives you 'Graf-Leuchtenberg-Weg 7'. Not even close to where you want to be. I don't know if the parameters were well chosen but it doesn't matter really because there will always be a real world case that gives you a similar bad result. What you are actually looking for is the closest address in _walking distance_ and that is not necessarily the same as the distance between the points. That is why Nominatim is relatively conservative here.

You are spot on with the observation that Nominatim is missing context information here about what exactly the caller wants. The `reverse` call used to just answer the question 'what is the closest OSM object'. Nowadays the answer is a bit closer to 'where am I'. When a navigation wants to know, which street it is on, we could give a much better answer if we knew where the car has been one or two seconds before. Recently there was [this question](https://help.openstreetmap.org/questions/77325/nominatim-reverse-geocode-city-nearby) to get the closest town to the position. Also a valid request but again different from 'where am I here?'. So your thinking about context goes into the right direction but is still too much geared towards the problems you currently have in mind. I'd like to see if there isn't a solution that solves all the use cases and also the future ones that we haven't thought of yet.

People occasionally ask for `reverse` to return more than one result. You are actually making a good case for implementing something like that. Then you could just ask for the 5 closest house numbers and if you feel fancy then you could even do a quick router call to find out which one has the shortest walking distance. In any case, you could do what ever suits you best with the five results, we wouldn't need to bake the logic into the geocoder.

lonvia:
Closing this here. Implementing a reverse call with multiple results is the better route to go.

PR #2032, opened 2020-10-29
Remove HTML output

### LINKED ISSUE #1799: Move HTML frontend to nominatim-ui
@mtmail has been working on a pure javascript frontend for Nominatim, the new [nominatim-ui](https://github.com/osm-search/nominatim-ui). Our plan is to remove HTML output completely from this repository and point people who need a UI to the new nominatim-ui repo instead.

This ticket is for tracking all the tasks we need to do to get to the html-free Nominatim:

- [x] Update documentation with installation instructions for nominatim-ui.
- [x] Add a migration guide for Apache and nginx to map the old URL schema to a split installation.
- [x] Move nominatim.openstreetmap.org to nominatim-ui as a test balloon.
- [x] Remove html format and GUI code

With the move we will likely remove the special request format `search/foo/bar`, which was never documented anyway and causes a bit of a headache for URL handling in the webserver.

lonvia:
This PR removes the debug UI provided so far by Nominatim. As a replacement there is now a separate UI under https://github.com/osm-search/nominatim-ui.

Some points to note:
* `/search` used html as its default output. I have changed this to `jsonv2`. The other option would have been to leave it blank and require a format parameter. This would give us more flexibility in the future but require more effort from the client. Opinions?
* `/hierarchy` is now gone. It has only HTML output, is half finished and most of the functionality provided can be queried via `/details`. It was never officially part of the UI, so I don't have an issue to remove it without notice.
* `CONST_Website_BaseURL` would only be needed in two places: for the more URL and the icon URL. For the more URL, I simply use the URL information from the webserver (in $_SERVER). The icons are no longer part of the repository, so we need an external URL for them. I'm not sure how many people have really used the `icon` field, especially when using their own installation. I've introduced a `CONST_MapIcon_URL` which can point to an installation of the UI. If it is unset (the default) then the `icon` field is simply no longer added. The latter should be fine for clients as the `icon` field was optional anyway.

Fixes #1799.


mtmail:
Looks good.

I agree /search should default is JSON. `&format=html` will return HTTP 400 'bad request'

I think /status should also default to JSON, and remove the text output.

Is it weird that /reverse and /lookup defaults to XML while all others default to JSON?

lonvia:
> Is it weird that /reverse and /lookup defaults to XML while all others default to JSON?

Totally. Still, I don't want to touch the other defaults as it potentially breaks other people's scripts. So I guess we file this under 'historical reasons'.

Issue #2036, opened 2020-11-01
Searching "Carrer de xxx" returns the wrong result


Xevib:
# Observed problem
- If I search "Carrer València, Girona" https://nominatim.openstreetmap.org/ui/search.html?q=Carrer+Val%C3%A8ncia%2C+Girona&dedupe=0 I get this street https://www.openstreetmap.org/way/84829665 when the expected is this https://www.openstreetmap.org/way/61947463
- If I search "Carrer de València,Girona" I get the correct street link: https://nominatim.openstreetmap.org/ui/search.html?q=carrer+de+Val%C3%A8ncia%2CGirona&dedupe=0

# Question
- If I add the "carrer de" string here it will be fixed? https://github.com/osm-search/Nominatim/blob/26bc83c9844ceba531837bfed79cd51b12a6b45a/module/tokenstringreplacements.inc#L447 

# Why 
- The "Carrer de" is very used in catalan Overpass query https://overpass-turbo.eu/s/ZAQ

jmontane:
Hi,

This is an annoying issue for years! Users expect get same results searching for "carrer "+foobar and "carrer de "+foobar.

I made PR #2037 fixing this issue for Catalan ("carrer" lines) and Spanish ("calle" lines) languages. That's for "street" class only-

lonvia:
Duplicate of #753.

PR #2037, opened 2020-11-01
Fixing " carrer de " != " carrer ", fixes Issue #2036

jmontane:
These changes fix missing expected results when user search by "carrer" or "carrer de" (In English 'street' or 'street of')

Lines with "carrer" are for Catalan language
Lines with "calle" are for Spanish language

### DIFF
--- a/module/tokenstringreplacements.inc
+++ b/module/tokenstringreplacements.inc
@@ -34,6 +34,7 @@
 	str_replace(buffer, &len, &changes, " qucyng truong ", 15, " qt ", 4, 0);
 	str_replace(buffer, &len, &changes, "strada statale ", 15, " ss ", 4, 0);
 	str_replace(buffer, &len, &changes, " state highway ", 15, " sh ", 4, 0);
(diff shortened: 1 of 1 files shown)


lonvia:
As already said in #925, we don't take any changes to the special phrases there at the moment as they tend to cause a lot of false matches. Sorry.

jmontane:
I understand it, but how can we improve output results then?

Xevib:
@lonvia There is any other way to solve this problem?
Perhaps we can have a list of stopwords to remove on the start of the name ? Somthing like a list of words to ignore a the start like "of"?
We are really interested in fixing this

lonvia:
Accepting this change would have two effects:

1. Everybody would have to do a complete reimport of their database or otherwise they would find that some streets cannot be found anymore at all, no matter what variant.
2. Searches would suddenly create a lot more false positives, causing a lot more new issues and complaints.

Both points on their own would be a show stopper to accepting a change. We need to address the underlying issues first and this is no simple matter that is fixed with an evening of coding. There is also no simple short cut around all this.

I'm closing this as the new infrastructure for abbreviations will be vastly different and the PR not applicable by then.

PR #2054, opened 2020-11-16
Merge places into address lists referred to by addr:* tags but not computed by Nominatim

### LINKED ISSUE #148: Take addr:* tags into account for computing the final address
When searching for an address in a US Census Designated Place (CDP) with an administrative boundary, the boundary overrides the address tag. The example is search for 7732 234th Place Northeast, Redmond, WA. Nominatim will return 0 results. However, the address is tagged with the following:
addr:city=Redmond
addr:housenumber=7732
addr:street=234th Place Northeast
addr:postcode=98053
name=7732

Omitting the city, Nominatim returns the correct location. Short of removing the admin boundary, I don't know how to correct the problem. I've changed the admin level from city (8) to neighborhood (9) with no help. 

I expected nominatim to return based on either addr:city or admin boundary. This is especially prevalent when areas are annexed by cities. Houses in annexed areas may have been using a different city name until annexed.  

The boundary, in this case could be removed, but I wonder how many other area might have similar problems. 

Thanks,
Clifford

lonvia:
This is the last building block towards full handling of addr:* tags.

With this PR the addr:* tags are no longer just added to the search terms but they get the full address treatment. That means that we try to look up a matching place in OSM (with a much larger search radius than the normal address building algorithm) and add the results to the place_addressline table. As a result, the addr:* terms are now displayed as part of the address. As we use the information from OSM entries, they also come with translations. The drawback is that the addr:* tags (with the expressed exception of addr:place) need to have a corresponding place in OSM.

To make the algorithm work for rank 30, POIs need to be allowed their own entries into the place_addressline table instead of only using the ones of their parents. To ensure that the table size doesn't explode, entries are only added, when the parent misses an entry for an addr:* tag. When retrieving the address, the entries for the parent and the POI are mixed and merged on the fly.

This PR also contains a cleanup of the `search_name_*` tables. The rank_search column is removed and the last uses converted to using the more correct rank_address. This change should be backward compatible with existing installations.

Fixes #148.

### DIFF
--- a/sql/functions/address_lookup.sql
+++ b/sql/functions/address_lookup.sql
@@ -90,7 +90,7 @@ DECLARE
   for_place_id BIGINT;
   result TEXT[];
   search TEXT[];
-  found INTEGER;
+  current_rank_address INTEGER;
   location RECORD;
   countrylocation RECORD;
   searchcountrycode varchar(2);
@@ -103,7 +103,7 @@ DECLARE
   search_unlisted_place TEXT;
   countryname HSTORE;
 BEGIN
-  -- The place ein question might not have a direct entry in place_addressline.
+  -- The place in question might not have a direct entry in place_addressline.
   -- Look for the parent of such places then and save if in for_place_id.
 
   postcode_isexact := false;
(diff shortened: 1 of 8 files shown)


lonvia:
This change needs a migration after all. The search_name_* tables need different indexes. Create them with the following shell command:

```
for table in `psql -d nominatim -c "SELECT tablename FROM pg_tables WHERE tablename LIKE 'search_name_%'" -tA | grep -v search_name_blank` ;
do
    psql -d nominatim -c "DROP INDEX idx_${table}_centroid_place; CREATE INDEX idx_${table}_centroid_place ON ${table} USING gist (centroid) WHERE ((address_rank >= 2) AND (address_rank <= 25)); DROP INDEX idx_${table}_centroid_street; CREATE INDEX idx_${table}_centroid_street ON ${table} USING gist (centroid) WHERE ((address_rank >= 26) AND (address_rank <= 27))"; 
done
```

PR #2060, opened 2020-11-18
Admin levels 8 and 10 of the Netherlands are municipal / city

hendrikmoree:
Admin levels in the Netherlands are little different. See https://wiki.openstreetmap.org/wiki/Template:Admin_level_11

For example, this is a small village: https://www.openstreetmap.org/relation/334641. Definitly not a suburb/neighbourhood/hamlet. This one should be rank 16
It's part of this municipality: https://www.openstreetmap.org/relation/418760. This one should be rank 14

I tried this change for a Photon import and the result is correct.

### DIFF
--- a/settings/address-levels.json
+++ b/settings/address-levels.json
@@ -164,6 +164,14 @@
           "administrative8" : 14
       }
   }
+},
+{ "countries" : [ "nl" ],
+  "tags" : {
+      "boundary" : {
+          "administrative8" : 14,
+          "administrative10" : 16
+      }
+  }
 }
 ]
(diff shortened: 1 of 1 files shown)


lonvia:
Looks good, thanks! I've added custom levels for admin level 7 and 9 in f89e71a861e1c447de63658a334383640cb831ea, just to make sure that the ranks and admin levels keep the same order when somebody suddenly starts using the admin levels in the future. You'll note that the address rank is 0, meaning that even if they show up, they will not be used in addresses.

PR #2115, opened 2020-12-20
Switch configuration to dotenv

lonvia:
This implements #946 and switches the configuration to dotenv format. I've used `symfony/dotenv` instead of the library by vlucas for the simple reason that it is packaged in Debian/Ubuntu and can therefore be included with minimal hassle. The drawback is that it completely breaks CentOS builds. Afaics CentOS hasn't even packaged composer which would allow to install dotenv as an external dependency and installing composer from scratch is way beyond what I'd like to see in our installation instructions. I hope that the use of the library will be very short-lived anyway. My plan after this is to write a python-based fronting that completely reorganises the Nominatim tools. The fronting would initially call the PHP scripts and then we can migrate step-by-step to native Python code. This fronting can take over parsing of the dotenv files and simply hand in the configuration as environment variables to PHP thus making the dotenv library redundant.

The new default configuration can be found in `settings/env.defaults`. There is much more extensive documentation now which is hopefully useful. Most of the settings are a 1:1 translation but there were a couple of additional changes needed:

* CONST_BasePath is no longer configurable. Instead the scripts are expected to know per installation time where their library and data files are.
* CONST_InstallPath is replaced by the notion of a 'project directory'. That is the directory that holds all information about one specific Nominatim import. Currently this still must be the build directory. The project directory can be set via the `--project-dir` parameter. The default is the current working directory. Thus nothing changes in the handling for the user yet.
* The HTTP proxy settings need to be applied explicitly now by util scripts that potentially need them. We don' t have the necessary information anymore when ` cmd.php`  is called.
* Many config settings are now empty by default. The documentation mentions what is used when they are unset. The reasoning behind this is that these settings point to internal scripts which the config file shouldn't have to know about the exact location.
* CONST_Search_AreaPolygons and CONST_Search_ReversePlanForAll have been removed.
* The `website` directory and its contents in the build directory are now created when `--setup-website` is called instead of at build time. The configuration is written into the php files directly instead of a separate `settings-frontend.php`.

There are a couple of follow up changes necessary to this, especially to get to the point where the project directory can be truly separate from the source and build directory, which in turn is necessary to get to a properly installable software package. But for now the PR is big enough as it is.

If you want to migrate you need to create a `.env`  in your build directory and port your custom settings there. Then rerun ` cmake ..`  and `./utils/setup.php --setup-website`. Afterwards the `settings` directory in the build directory can be safely deleted.


nspaeth:
https://github.com/osm-search/Nominatim/blob/6cc06828dbc722deb1c06dc4176400f727eb24dc/settings/env.defaults#L13

In this line, `www-data` is quoted, but seems to be used as if it were unquoted in https://github.com/osm-search/Nominatim/blob/e14e7c6235a40dbec451146bcb3aaec013d659c9/nominatim/tools/database_import.py#L64 
Producing `Web user '"www-data""' does not exists. Create it with...`

lonvia:
Please open a new issue if you run into problems with the code.

PR #2145, opened 2021-01-19
Add interface to search via command line tool

lonvia:
Adds the commands `search`, `reverse`, `lookup`, `details` and `status` to the CLI tool. These call directly the corresponding API endpoints using php_cgi. This way it won't be necessary anymore to install a webserver when Nominatim should just be used locally.

The new commands deprecate the `query.php` utility which had its own parameter input and output format and thus duplicated quite a bit of the code of the search endpoint.


missinglink:
This is a really cool feature ;)

PR #2204, opened 2021-03-08
Ported tiger-data-import to Python and Added Tarball Support

### LINKED ISSUE #2187: Import Tiger data directly from tarball
The [function for importing TIGER data](https://github.com/osm-search/Nominatim/blob/204fe20b4b660bdf34953b87763f9f7d06525a28/lib-php/setup/SetupClass.php#L430) still needs porting to Python. While doing this, we should add support to import the data directly from the tarball, so that the user no longer needs to have the unpacked version on the disk.

* porting a PHP setup function to Python, see for example [load-data](https://github.com/osm-search/Nominatim/commit/57db5819efb1eac7497aca2dbdb2493a67bd33f9)
* [Python tarfile support](https://docs.python.org/3.8/library/tarfile.html)
* [running asynchronous SQL commands](https://github.com/osm-search/Nominatim/blob/e14e7c6235a40dbec451146bcb3aaec013d659c9/nominatim/tools/database_import.py#L222) for parallel execution

darkshredder:
Fixes #2187 
### This PR includes the following changes
- Added Tarball Support while importing Tiger data
- Ported import-tiger-data php function to python
- Importing supports asynchronous parallel execution


darkshredder:
@lonvia @mtmail  please review this and tell me if any changes are required

lib-php/admin/setup.php
```diff
@@ -161,7 +161,7 @@ function run($oCmd)
 if ($aCMDResult['import-tiger-data']) {
     $bDidSomething = true;
     $sTigerPath = getSetting('TIGER_DATA_PATH', CONST_InstallDir.'/tiger');
-    $oSetup->importTigerData($sTigerPath);
```
lonvia, line 164: It should be possible to remove the `importTigerData()` function now.

nominatim/tools/tiger_data.py
```diff
@@ -0,0 +1,98 @@
+"""
+Functions for setting up and importing a new Nominatim database.
```
lonvia, line 2: The comment could be more specific.

nominatim/tools/tiger_data.py
```diff
@@ -0,0 +1,98 @@
...
+            file = tar.extractfile(sql_file)
+        lines = 0
+        end_of_file = False
+        total_used_threads = place_threads
+        while True:
+            if end_of_file:
+                break
+            for imod in range(place_threads):
+                conn = DBConnection(dsn)
+                conn.connect()
```
lonvia, line 62: Creating a new connection for every single line to import is a little bit expensive. Try the following: create a pool of n connections before starting file processing (in line 47) and register them with `sel`. Then, in this line, you can simple get the next non-busy connection using `sel.select(1)` followed by conn.is_done() to check that the connection is really ready.

nominatim/tools/tiger_data.py
```diff
@@ -0,0 +1,98 @@
...
+import os
+import tarfile
+import selectors
+
+from ..db.connection import connect
+from ..db.async_connection import DBConnection
+from ..db.sql_preprocessor import SQLPreprocessor
+
+# pylint: disable=R0912
+# pylint: disable=R0914,R0915,W0702
```
lonvia, line 14: Please really fix those warnings by splitting up your function in smaller ones.

nominatim/tools/tiger_data.py
```diff
@@ -0,0 +1,98 @@
...
+                        todo -= 1
+
+    if is_tarfile:
+        tar.close()
+    print('\n')
+    LOG.warning("Creating indexes on Tiger data")
+    with connect(dsn) as conn:
+        sql = SQLPreprocessor(conn, config, sqllib_dir)
+        sql.run_sql_file(conn, 'tiger_import_finish.sql')
+
```
lonvia, line 98: That's odd. The trailing newline should have triggered in pylint, too.

lonvia (changes requested):
This looks pretty good already. I've added some comments inline.

It would be nice if you could add some tests for the new function `add_tiger_add()` in `test/python`. The tests don't need to import actual Tiger data. They should just prove that the function correctly reads directory and tar files and executes SQL. You can write them in a way that they just execute files with arbitrary SQL and then you check that you see the effects in the database.

darkshredder:
> It would be nice if you could add some tests for the new function `add_tiger_add()` in `test/python`. The tests don't need to import actual Tiger data. They should just prove that the function correctly reads directory and tar files and executes SQL. You can write them in a way that they just execute files with arbitrary SQL and then you check that you see the effects in the database.

@lonvia  I think a test is already there for testing add_tiger_add() function which is doing the same thing as you mentioned.
https://github.com/osm-search/Nominatim/blob/35f4695b67b50a41caf69378ea13044ea91f1d2c/test/python/test_cli.py#L68

lonvia:
> I think a test is already there for testing add_tiger_add() function which is doing the same thing as you mentioned.

This test just tests the cli functionality, i.e. checking that CLI calls are routed to the correct function. It mocks away the function that does the actual work, so that it never gets executed. The actual functions are tested separately, for example, have a look at https://github.com/osm-search/Nominatim/blob/35f4695b67b50a41caf69378ea13044ea91f1d2c/test/python/test_tools_refresh_create_functions.py#L15.

Actually, if the test in test_cli.py still passes, that means that you have forgotten to replace the legacy call [here](https://github.com/osm-search/Nominatim/blob/35f4695b67b50a41caf69378ea13044ea91f1d2c/nominatim/cli.py#L170) with a call to your new function.

darkshredder:
@lonvia, I have made changes and added tests. Can you please review it again?

test/python/test_tools_tiger_data.py
```diff
@@ -0,0 +1,45 @@
...
+@pytest.mark.parametrize("threads", (1, 5))
+def test_add_tiger_data(dsn, src_dir, def_config, monkeypatch,tmp_path,
+                        temp_db_cursor, threads, temp_db):
+    monkeypatch.setenv('NOMINATIM_DATABASE_MODULE_PATH', '.')
+    temp_db_cursor.execute('CREATE EXTENSION hstore')
+    temp_db_cursor.execute('CREATE EXTENSION postgis')
+    temp_db_cursor.execute('CREATE TABLE place (id INT)')
+
+    database_import.import_base_data('dbname=' + temp_db, src_dir / 'data',
+                                     ignore_partitions=False)
```
lonvia, line 21: There is a [fixture sql_preprocessor](https://github.com/osm-search/Nominatim/blob/35f4695b67b50a41caf69378ea13044ea91f1d2c/test/python/test_db_sql_preprocessor.py#L26), which does the minimum setup required to be able to use the SQLPreprocessor class. If you move that into `conftest.py`, you can use the fixture here as well. That would replace the setenv and import_base_data. Less typing and when the requirements for the SQLPreprocessor change later, there is only one function that needs adapting.

test/python/test_tools_tiger_data.py
```diff
@@ -0,0 +1,45 @@
...
+from pathlib import Path
+
+import pytest
+import tarfile
+
+from nominatim.tools import tiger_data, database_import
+
+
+@pytest.mark.parametrize("threads", (1, 5))
+def test_add_tiger_data(dsn, src_dir, def_config, monkeypatch,tmp_path,
```
lonvia, line 13: Missing space before tmp_path.

nominatim/tools/tiger_data.py
```diff
@@ -0,0 +1,114 @@
...
+def handle_threaded_sql_statements(sel, file):
+    """ Handles sql statement with multiplexing
+    """
+
+    lines = 0
+    end_of_file = False
+    # Using pool of database connections to execute sql statements
+    while True:
+        if end_of_file:
+            break
```
lonvia, line 49: These three lines can be shortened to `while not end_of_file:`

nominatim/tools/tiger_data.py
```diff
@@ -0,0 +1,114 @@
...
+                    lines += 1
+                    if not sql_query:
+                        end_of_file = True
+                        break
+                    conn.perform(sql_query)
+                    if lines == 1000:
+                        print('. ', end='', flush=True)
+                        lines = 0
+            except Exception as exc: # pylint: disable=broad-except
+                LOG.error('Wrong SQL statement: %s', exc)
```
lonvia, line 64: Can you move this to level 'info'? We have silently ignored those errors already before.

lonvia (commented):
The tests are good now, you can leave them like this. I'd just like to point out that both having only one file and having only one SQL statement in the file are edge cases in terms of testing. As a result your tests give you good code coverage but don't test all the functionality of your function. For example, even with the 5-thread-test there are no SQL statements truly run in parallel.

nominatim/tools/tiger_data.py
darkshredder: Sure

darkshredder:
@lonvia, Can you please review it again. I have made changes as you mentioned above.

lonvia (approved):

PR #2212, opened 2021-03-12
Ported createCountryNames() to python and Added tests

### LINKED ISSUE #2202: Port country name setup to Python
The function for [adding country names to the word index](https://github.com/osm-search/Nominatim/blob/35f4695b67b50a41caf69378ea13044ea91f1d2c/lib-php/setup/SetupClass.php#L203) still needs porting from PHP to Python. The task is fairly straightforward as only a couple of SQL commands need to be executed. The ported function then also needs some tests.

* porting a PHP setup function to Python, see for example [load-data](https://github.com/osm-search/Nominatim/commit/57db5819efb1eac7497aca2dbdb2493a67bd33f9)
* [tests for ported setup functions](https://github.com/osm-search/Nominatim/blob/master/test/python/test_tools_database_import.py)

darkshredder:
Fixes #2202 

### This PR includes the following changes

- Ported createCountryNames() function from PHP to Python 
- Added test for the ported function

### DIFF
--- a/lib-php/setup/SetupClass.php
+++ b/lib-php/setup/SetupClass.php
@@ -130,33 +130,6 @@ class SetupFunctions
         $this->db()->exec($sSQL);
     }
 
-    public function createCountryNames()
-    {
-        info('Create search index for default country names');
-
-        $this->pgsqlRunScript("select getorcreate_country(make_standard_name('uk'), 'gb')");
-        $this->pgsqlRunScript("select getorcreate_country(make_standard_name('united states'), 'us')");
-        $this->pgsqlRunScript('select count(*) from (select getorcreate_country(make_standard_name(country_code), country_code) from country_name where country_code is not null) as x');
-        $this->pgsqlRunScript("select count(*) from (select getorcreate_country(make_standard_name(name->'name'), country_code) from country_name where name ? 'name') as x");
-        $sSQL = 'select count(*) from (select getorcreate_country(make_standard_name(v),'
-            .'country_code) from (select country_code, skeys(name) as k, svals(name) as v from country_name) x where k ';
-        $sLanguages = getSetting('LANGUAGES');
-        if ($sLanguages) {
-            $sSQL .= 'in ';
-            $sDelim = '(';
-            foreach (explode(',', $sLanguages) as $sLang) {
-                $sSQL .= $sDelim."'name:$sLang'";
-                $sDelim = ',';
-            }
-            $sSQL .= ')';
-        } else {
-            // all include all simple name tags
-            $sSQL .= "like 'name:%'";
-        }
-        $sSQL .= ') v';
-        $this->pgsqlRunScript($sSQL);
-    }
-
     /**
      * Return the connection to the database.
      *
--- a/nominatim/tools/database_import.py
+++ b/nominatim/tools/database_import.py
@@ -306,3 +306,38 @@ def create_search_indices(conn, config, sqllib_dir, drop=False):
     sql = SQLPreprocessor(conn, config, sqllib_dir)
 
     sql.run_sql_file(conn, 'indices.sql', drop=drop)
+
+def create_country_names(conn, config):
+    """ Create search index for default country names.
+    """
+
+    with conn.cursor() as cur:
+        cur.execute("""select
+                       getorcreate_country(make_standard_name('uk')
+                       , 'gb')""")
+        cur.execute("""select getorcreate_country(make_standard_name('united states'), 'us')""")
+        cur.execute("""select count(*) from
+                       (select getorcreate_country(make_standard_name(country_code),
+                       country_code) from country_name where country_code is not null) as x""")
+        cur.execute("""select count(*) from
+                        (select getorcreate_country(make_standard_name(name->'name'),
+                        country_code) from country_name where name ? 'name') as x""")
+        sql_statement = """select count(*) from (select getorcreate_country(make_standard_name(v)
+                           , country_code) from (select country_code, skeys(name)
+                           as k, svals(name) as v from country_name) x where k """
+
+        languages = config.LANGUAGES
+
+        if languages:
+            sql_statement += 'in '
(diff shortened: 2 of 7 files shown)


nominatim/tools/database_import.py
```diff
@@ -306,3 +306,38 @@ def create_search_indices(conn, config, sqllib_dir, drop=False):
...
+        cur.execute("""select
+                       getorcreate_country(make_standard_name('uk')
+                       , 'gb')""")
+        cur.execute("""select getorcreate_country(make_standard_name('united states'), 'us')""")
+        cur.execute("""select count(*) from
+                       (select getorcreate_country(make_standard_name(country_code),
+                       country_code) from country_name where country_code is not null) as x""")
+        cur.execute("""select count(*) from
+                        (select getorcreate_country(make_standard_name(name->'name'),
+                        country_code) from country_name where name ? 'name') as x""")
```
lonvia, line 324: Please adopt to the style of SQL statements in the rest of the code. In particular: keywords should be upper case. Also line breaks should go where they make sense semantically, certainly not before a comma.

nominatim/tools/database_import.py
```diff
@@ -306,3 +306,38 @@ def create_search_indices(conn, config, sqllib_dir, drop=False):
...
+            sql_statement += 'in '
+            delim = '('
+            for language in languages.split(','):
+                sql_statement += delim + "'name:" + language + "'"
+                delim = ','
+
+            sql_statement += ')'
+        else:
+            sql_statement += "like 'name:%'"
+        sql_statement += ') v'
```
lonvia, line 341: I'm aware that this is a 1:1 copy of the original code but fragmenting the SQL like that makes it hard to read and maintain. So it would be nice if you can improve the code here: create the string for the where clause and then use format() to create the overall statement with the where clause inserted at the appropriate place.

test/python/test_tools_database_import.py
```diff
@@ -200,3 +200,26 @@ def test_load_data(dsn, src_dir, place_row, placex_table, osmline_table, word_ta
...
+                                               lookup_country_code varchar(2))
+                            RETURNS INTEGER
+                            AS $$
+                            BEGIN
+                                INSERT INTO country_name VALUES (5, lookup_word);
+                                RETURN 5;
+                            END;
+                            $$
+                            LANGUAGE plpgsql;
+                               """)
```
lonvia, line 222: You can't fake the function like that. The purpose of create_country_names() is to take information from the country_name table and add it to the word table. If you insert values into country_name in your test getorcreate_country() you invert the purpose.

test/python/test_tools_database_import.py
```diff
@@ -200,3 +200,26 @@ def test_load_data(dsn, src_dir, place_row, placex_table, osmline_table, word_ta
...
+    temp_db_cursor.execute('CREATE TABLE country_name (id int, country_code varchar(2), name hstore)')
```
lonvia, line 223: Again, this hits all lines but doesn't really test if the language selector works. You should add a couple of lines in that table.

lonvia (changes requested):

nominatim/tools/database_import.py
darkshredder: Sure will do that

test/python/test_tools_database_import.py
darkshredder: So Should I recreate the whole function?

nominatim/tools/database_import.py
darkshredder: Sure will fix it

test/python/test_tools_database_import.py
lonvia: No, a simple function like this is fine. Just create an extra table to insert the lines to, so that you do not change the input table `country_name`.

darkshredder: Sure will do this then. Actually, initially, my idea was this only. But later, I thought that it would be better to use a country table instead of creating a new table so that I don't have to create a new table.

darkshredder:
@lonvia, Can you please review it.

PS: I have made the test function in such a way that when languages are not present in env, then only the getorcreate_country function will be triggered and have one more row added int word table, i.e. will execute five times in SQL script else it will be triggered in only first four SQL script

nominatim/tools/database_import.py
```diff
@@ -306,3 +306,35 @@ def create_search_indices(conn, config, sqllib_dir, drop=False):
...
+        if languages:
+            sql_statement = "{} IN (".format(sql_statement)
+            delim = ''
+            for language in languages.split(','):
+                sql_statement = "{}{}'name:{}'".format(sql_statement, delim, language)
+                delim = ', '
+            sql_statement = '{})'.format(sql_statement)
+        else:
+            sql_statement = "{} LIKE 'name:%'".format(sql_statement)
+        sql_statement = "{}) v".format(sql_statement)
```
lonvia, line 338: Not quite what I meant. The SQL code is still fragmented in odd ways. For example, line 323 has an opening bracket which is closed in line 338. Similar brackets are in line 330 and 335. Anyway, lets leave it at that. The code can be optimised later.

lonvia (commented):
Please change `admin/setup.php`, then it's good to go.

lib-php/setup/SetupClass.php
```diff
@@ -130,33 +130,6 @@ public function calculatePostcodes($bCMDResultAll)
         $this->db()->exec($sSQL);
     }
 
-    public function createCountryNames()
```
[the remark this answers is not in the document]
darkshredder: Yeah, sorry, I forgot to add that file to commits, My bad.

nominatim/tools/database_import.py
darkshredder: Sure.

PS: Now I understood what your issue was. You wanted to refactor the whole SQL statement.

lonvia (approved):

Issue #1892, opened 2020-07-27
Replace normalization module with frontend-side normalization via libICU


lonvia:
The normalization module is one of the parts of Nominatim that make installation and deployment very difficult. It is also outdated and missing a lot scripts. We should get rid of the normalization module and do it in the frontend code instead. There we can easily make use of libICU or similar and will be flexible to add customized normalization modules.

There are two parts in the code where normalization is applied:

- **Import**: currently the normalization is done as part of the placex update trigger. To move the normalization out, we need to do the normalization in the indexing code of the nominatim.py module. So for each place we need to get the possible names out of the database, normalize them and reinsert them for further processing in the trigger code.

- **Query**: the normalization is the first stage of forward query processing (see `make_standard_name()`) in lib/Geocode.php.

The tricky part here is that the two parts are written in two different languages. Luckily both have decent support for libICU. Still we either first switch to Python for everything or do the normalization twice for a while.

otbutz:
Dealing with normalization in plpython is tricky: https://github.com/giggls/mapnik-german-l10n/issues/53

Issue #1362, opened 2019-04-20
New Special Phrases - autoimport - is it possible?


mutipg:
Hi, I found another old and closed thread in this subject, but I want ask for import new special phrases, which one I had added in Polish language. Some of them I had added 6 months ago.

I think, that special phrases should be importing automatically at least monthly. Is there a possibility to write any script for this? Frequently updating special phrases will be stimulating for adding new phrases in local language for users.

Instead of autoimport once more please to import Polish Special Phrases to Nominatim: Nominatim/Special_Phrases/PL

Thank you!

mark-lester:
I'll take this next. 
I dont know if you assign this to me I can in turn assign  it to a pull request. Lets try

lonvia:
We'd need a routine similar to the special phrase import. Instead of creating all the special phrases and tables after having read the data from the wiki, the update routine would have to:

* get the list of current phrases from the `word` table, and remove outdated entries
* add all phrases entries that are new to the `word` table
* remove all `place_classtype_*` tables that are no longer needed
* add `place_classtype_*` tables that are not yet there

AntoJvlt:
To be sure, do we need a new command for this, like: special-phrases --update-from-wiki?

Or should it be handled by the existing special-phrases --import-from-wiki by looking for an update if needed?

lonvia:
If `--import-from-wiki` magically does the right thing that would be perfect. But if it is easier to have an extra function then that is fine with me, too.

AntoJvlt:
Sure, I will start working on it.

AntoJvlt:
I am taking some time to solve this because I have a lot of other works to produce for this wenesday.

I have one question in order to not make any mistakes. I am pretty sure the word table is not only filled and used by the special phrases, how can I safely check for the current imported special phrases into it? Should I check that the operator is not null or that the class and type are not null, or checks both maybe?

lonvia:
Indeed, that would be useful to know. It's a special phrase if class is not null except for `class=place` and `type=house`, which is a housenumber term.

Best place to look up the different types of words is in the query code where the word entries are collected: https://github.com/osm-search/Nominatim/blob/c0f0b665091dfacd29727efed5efbe111c5900bf/lib-php/TokenList.php#L109

PR #2228, opened 2021-03-20
Import special phrases porting python

### LINKED ISSUE #2203: Import of special phrases needs porting to Python
The function to [import special phrases from the OSM wiki](https://github.com/osm-search/Nominatim/blob/master/lib-php/admin/specialphrases.php) still needs porting from PHP to Python. The ported function then needs some testing.

* porting a admin functions to Python, see for example [port of check_database functions](https://github.com/osm-search/Nominatim/commit/b169e4c88cd99075c7e932aeb6587c97f51e7ed1)
* [tests for ported setup functions](https://github.com/osm-search/Nominatim/blob/master/test/python/test_tools_database_import.py)

AntoJvlt:
This PR resolves #2203 :

## Changes for this PR

- Ported special_phrases.php to nominatim/tools/special_phrases.py
- Changed command from 'nominatim special-phrases' to 'nominatim import-special-phrases'
- import-special-phrases command is defined in nominatim/clicmd/special_phrases.py
- import-special-phrases --from-wiki now import special phrases from the wiki directly into the database, not to a temporary sql file
- Tests for nominatim/tools/special_phrases were added in test/python/test_tools_import_special_phrases.py
- settings/phrase_settings.php was removed and replaced by settings/phrase_settings.py
- Cmake now configure the path to the configure directory with a new /settings final directory (see later explanation): `${CMAKE_INSTALL_FULL_SYSCONFDIR}/${PROJECT_NAME}/settings`

## Details

Explanation of some changes

### Tansliteration in python

In php transliteration of inputs was done with the class Transliterator::createFromRules($rules) with the rules wrote in LDML. In python we can't reproduce this and use LDML rules easily (I think #1892 is quite related to this).
The only solution I found too was to use libICU but it doesn't sound the right moment to do it.

### Moved all normalization to SQL

In order to solve this Transliteration problem I found that the 2 SQL functions `getorcreate_amenity` and `getorcreate_amenityoperator` (from "normalization.sql") which are called to import the special phrases were using "lookup_word" as parameter which is normalized with the `make_standard_name` SQL function and a "normalized_word" parameter which is the same as "lookup_word" but normalized on php side with the Transliterator. 

As it was doing the same work 2 times and since we don't have Transliterator in python I choose to only keep the "lookup_word" parameter so that only `make_standard_name` is used.

As I have understood it could be good for the future to move all normalization away from SQL?

### Importation directly into the database

With php some of the variables exported from the wiki were escaped with `pg_escape_string()` from the postgres driver. With psycopg2 we can't do that and it is not recommended. It is preferred to use prepared statement and postgres auto formatting.
The only way to export the generated SQL to a string is to use the `SQL module` from psycopg2 and then do `query.as_string(conn)`. 

Since we need a database connection to do that I found that it would be much easier and that it would make much more sense to directly execute the query and import the data to the db in a secure way.

I also did this since I didn't found any use case other than generating the sql file and then executing it with `psql -f file.sql` in the old version, so I think it will be an upgrade and it will be more convenient. If it wasn't a good decision, please let me know.

That's why I changed the command name to `nominatim import-special-phrases --from-wiki`

### Settings folder

In order to use settings in the python code I removed settings/phrase_settings.php and created settings/phrase_settings.py.
Problem is that python couldn't access this file when the project was built because it couldn't find the module. 
So I created `__init__.py` in the /settings folder and changed CMakeLists.txt in order for it to create a new /settings folder where to put all configuration file so that python can understand the settings import and access configurations files.

### Tests

I did some tests for the new import-special-phrases command and for the functions in tools/special_phrases.py wich test data import to the database. I am not sure that this is enough, please let me know.


lonvia:
Thanks. Looks quite good already. I' ll have a closer look at the code later. Just some quick general comments for now.

Transliteration: you should use libICU for that. This is what the PHP code already does and in this particular case it is important that the Python code does exactly what the PHP code does. The normalization is not only used when importing but also [at query time](https://github.com/osm-search/Nominatim/blob/118befd7d7453b7abf89afaa614e3edfafc47a1b/lib-php/Geocode.php#L523). If the two normalizations don't match, then we are in trouble. To complicate things further: `make_standard_name` and the ICU normalization produce different kind of normalized terms and are used slightly differently. `make_standard_name` creates more lenient matches (for example, terms like 'street' or [abbreviated](https://github.com/osm-search/Nominatim/blob/master/module/tokenstringreplacements.inc)) while the ICU normalization only applies a minimal normalization, e.g. for case. We need both for the special terms. The normalization created by `make_standard_name` creates the initial match with the special phrase, while the ICU normalization is then used to ensure that the unabbreviated form of the word appears in the search query. That's all just to say: please keep both normalizations.

Settings: I would suggest to simply convert phrase_settings to a JSON file and then work with that. Then you don't have the difficulties with the module path. Having the file s a code file is some historical baggage we don't need to keep.  Bonus points if you can write a [migration](https://github.com/osm-search/Nominatim/pull/2200): if NOMINATIM_PHRASE_CONFIG is set to a custom configuration, convert the file to JSON if you find that it is still in PHP format.

lib-php/admin/specialphrases.php: can you write a tiny dummy wrapper [like this](https://github.com/osm-search/Nominatim/blob/118befd7d7453b7abf89afaa614e3edfafc47a1b/lib-php/admin/check_import_finished.php) instead of deleting it completely. We'd like to keep command-line compatibility for now, so that users can gradually switch to the new Python-based nominatim tool.

AntoJvlt:
Thanks for your review.

I am working on the changes you suggested. I am currently stuck with the use of ICU through the PyICU wrapper (the installation and creation of Transliterator object is working but I have some troubles with the transliteration rule's syntax creating errors).

I try to find a solution on ICU support side and will commit here when everything is ok.

AntoJvlt:
### Update

Changes made:

- Introduced PyICU to use Transliterator rules in python as we do in PHP, this python wrapper need LibICU to work.
- Modified documentation to add the new LibICU and PyICU dependency.
- Replaced phrase_settings.py by phrase-settings.json.
- Added little migration script in PHP (migration/phraseSettingsToJson.php) in order to convert a phrase settings file in PHP to a phrase settings file in JSON. This script is call only if NOMINATIM_PHRASE_CONFIG is set and that there is no corresponding JSON file already created. The new JSON file is created in the same directory as the configured PHP file.
- Added a wrapper in specialphrases.php to call the corresponding nominatim command.

To convert the PHP file to JSON file I did a PHP script because it is much easier. Otherwise we should use a third party library and I think it is quite too much for that. We could also parse the file ourselves but I think it is not 100% accurate and convenient.

settings/env.defaults
```diff
@@ -77,7 +77,7 @@ NOMINATIM_TIGER_DATA_PATH=
 NOMINATIM_WIKIPEDIA_DATA_PATH=
 
 # Configuration file for special phrase import.
-# When unset, the internal default settings from 'settings/phrase_settings.php'
+# When unset, the internal default settings from 'settings/phrase_settings.py'
```
lonvia, line 80: py -> json

test/python/test_tools_import_special_phrases.py
```diff
@@ -0,0 +1,174 @@
...
+    phrase_type = 'type'
+
+    assert execute_and_verify_add_word(temp_db_conn, phrase_label, normalized_label, 
+                                       phrase_class, phrase_type)
+    assert execute_and_verify_add_word_with_operator(temp_db_conn, phrase_label, normalized_label, 
+                                                     phrase_class, phrase_type, 'near')
+    assert execute_and_verify_add_word_with_operator(temp_db_conn, phrase_label, normalized_label, 
+                                                     phrase_class, phrase_type, 'in')
+
+def index_exists(db_connect, index):
```
lonvia, line 53: You can use temp_db_conn.index_exists() in the tests.

test/python/test_tools_import_special_phrases.py
```diff
@@ -0,0 +1,174 @@
+"""
+    Tests for import special phrases functions
+"""
+import pytest
+from nominatim.tools.special_phrases import _create_place_classtype_indexes, _create_place_classtype_table, _get_wiki_content, _grant_access_to_webuser, _process_amenity
+
+def test_get_wiki_content():
+    assert _get_wiki_content('fr')
```
lonvia, line 8: The unit tests here should avoid retrieving actual data from the Internet. It tends to fail at inconvenient times. It's fine if you keep _get_wiki_content() untested because of this. Where you need an actual page for the tests, download a current version, put it into `test/testdata` and mock _get_wiki_content() to return the saved file.

test/python/test_tools_import_special_phrases.py
```diff
@@ -0,0 +1,174 @@
...
+        INTO return_word_id;
+        IF return_word_id IS NULL THEN
+            return_word_id := nextval('seq_word');
+            INSERT INTO word VALUES (return_word_id, lookup_token, normalized_word,
+                                    lookup_class, lookup_type, null, 0, op);
+        END IF;
+        RETURN return_word_id;
+        END;
+        $$
+        LANGUAGE plpgsql;""")
```
lonvia, line 174: If you want, you can greatly simplify these function to just do a dummy action and/or return a dummy result. The SQL is not under test here. The test just need to check that the functions were called at all.

lonvia (commented):
I still want to run a few tests against the code but it looks pretty good already. Find some small nitpicking stuff inline.

nominatim/tools/special_phrases.py
```diff
@@ -0,0 +1,249 @@
...
+def _create_place_classtype_table(db_connection, sql_tablespace, phrase_class, phrase_type):
+    """
+        Create table place_classtype of the given phrase_class/phrase_type if doesn't exit.
+    """
+    with db_connection.cursor() as db_cursor:
+        db_cursor.execute(SQL(f"""
+                CREATE TABLE IF NOT EXISTS {{}} {sql_tablespace} 
+                AS SELECT place_id AS place_id,st_centroid(geometry) AS centroid FROM placex 
+                WHERE class = {{}} AND type = {{}}""")
+                          .format(Identifier(f'place_classtype_{phrase_class}_{phrase_type}'),
```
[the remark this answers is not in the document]
AntoJvlt: That's something I was worry off, it is not a big deal if we don't secure the input?  (I know we are not in a big sensitive application there but I still wonder)

.github/actions/build-nominatim/action.yml
```diff
@@ -6,7 +6,8 @@ runs:
     steps:
         - name: Install prerequisites
           run: |
-            sudo apt-get install -y -qq libboost-system-dev libboost-filesystem-dev libexpat1-dev zlib1g-dev libbz2-dev libpq-dev libproj-dev python3-psycopg2 python3-pyosmium python3-dotenv python3-psutil python3-jinja2
+            sudo apt-get install -y -qq libboost-system-dev libboost-filesystem-dev libexpat1-dev zlib1g-dev libbz2-dev libpq-dev libproj-dev libicu-dev python3-psycopg2 python3-pyosmium python3-dotenv python3-psutil python3-jinja2
+            sudo pip install PyICU
```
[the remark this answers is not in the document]
AntoJvlt: I only find the python-pyicu package on debian and on xenial and bionic distributions of ubuntu. It isn't available on my ubuntu distribution (focal) so I went for the PyICU python package. Is pip making any trouble?

nominatim/tools/special_phrases.py
```diff
@@ -0,0 +1,249 @@
...
+    Functions to import special phrases into the database.
+"""
+import logging
+import os
+import re
+import subprocess
+import sys
+import json
+from os.path import isfile
+from icu import Transliterator # pylint: disable-msg=no-name-in-module
```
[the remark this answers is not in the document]
AntoJvlt: icu is supposed to be the PyICU python module, to be honest I didn't get why I got this pylint error that's why I ignored it.

nominatim/tools/special_phrases.py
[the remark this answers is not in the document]
lonvia: It's just the [`f'...'` construct](https://docs.python.org/3/reference/lexical_analysis.html#f-strings), which you can't use. You need to go back to good old [format strings](https://docs.python.org/3/library/string.html#formatstrings) a la `"...".format()`.

THE `SQL()`  and `Identifier()` constructs from psycopg are fine. They were introduced in psycopg 2.7, which is the minimum version we require. In fact, it' s excellent that you have use them here as this code processes input from the internets.

.github/actions/build-nominatim/action.yml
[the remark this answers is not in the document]
lonvia: Ah, just for the fun of it the Python3 version is called `python3-icu`.

Using pip is fine. We just try to give installable packages the preference when they are available because they are less likely to cause clashes with other installed software.

nominatim/tools/special_phrases.py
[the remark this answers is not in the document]
AntoJvlt: Oh yes sorry, I have read way too fast. I will change this, thanks.

cmake/tool-installed.tmpl
```diff
@@ -2,7 +2,9 @@
 import sys
 import os
 
-sys.path.insert(1, '@NOMINATIM_LIBDIR@/lib-python')
+sys.path.insert(0, '@NOMINATIM_LIBDIR@/lib-python')
+#Add config directory to the python path for module importation
+sys.path.insert(1, '@NOMINATIM_CONFIGDIR@/..')
```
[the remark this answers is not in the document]
AntoJvlt: Yes, I will delete it.

nominatim/tools/special_phrases.py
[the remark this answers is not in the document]
AntoJvlt: Ok you were right since PyICU is generating the object dynamically to wrap the C++ one, it wasn't detected by pylint. Adding icu to extension-pkg-whitelist doesn't solve the problem, some people reported this on pylint's github, so I went for the solution with ignored-modules=icu instead.

AntoJvlt:
I made all changes. The python3-icu package is indeed available, it was my mistake.

.gitignore
```diff
@@ -9,3 +9,4 @@ data/wiki_specialphrases.sql
 data/osmosischange.osc
 
 .vagrant
+.vscode
```
lonvia, line 12: We don't take editor-specific ignores here. Those should usually go into the .gitignorefile in your home directory.

nominatim/tools/special_phrases.py
```diff
@@ -120,9 +120,10 @@ def _check_sanity(lang, phrase_class, phrase_type, pattern):
     """
     try:
         if len(pattern.findall(phrase_class)) < 1 or len(pattern.findall(phrase_type)) < 1:
-            sys.exit()
-    except SystemExit:
-        LOG.error("Bad class/type for language %s: %s=%s", lang, phrase_class, phrase_type)
+            raise UsageError("Bad class/type for language {}: {}={}".format(
+                lang, phrase_class, phrase_type))
+    except UsageError:
+        LOG.error(UsageError)
```
lonvia, line 126: You can also simply put the `LOG.error()` before the raise and get rid of the whole try/except/raise pattern.

lonvia (commented):
You have accidentally committed `data/country_osm_grid.sql.gz`. Can you please remove that with a 'git rebase', so that it does not show in the history. Sorry about that. I should have added that file to .gitgnore earlier to avoid that kind of accident.

AntoJvlt:
Ok I made the changes, I rebased to remove the file accident and I took the opportunity to squash my last little commits to make it simpler. I made quite a mess with the force-push, sorry for that.

nominatim/tools/special_phrases.py
```diff
@@ -0,0 +1,245 @@
...
+    LOG.warning('Create tables and indexes...')
+
+    sql_tablespace = config.TABLESPACE_AUX_DATA
+    if sql_tablespace:
+        sql_tablespace = ' TABLESPACE '+sql_tablespace
+
+    with db_connection.cursor() as db_cursor:
+        db_cursor.execute("CREATE INDEX idx_placex_classtype ON placex (class, type)")
+
+    for pair in class_type_pairs.items():
```
lonvia, line 159: I get a 

```
  File "Nominatim/nominatim/tools/special_phrases.py", line 159, in _create_place_classtype_table_and_indexes
    for pair in class_type_pairs.items():
AttributeError: 'set' object has no attribute 'items'
```

AntoJvlt: This error is my bad as I was only testing the first part of the process... Fixed.

AntoJvlt:
I cleaned up the code in tools/special_phrases by creating the SpecialPhrasesImporter class.

I also took the opportunity to add tests for `_process_xml_content` method and for `import_from_wiki`.

I had to introduce pytest-mock in order to mock the `_get_wiki_content` method to return the content of testdata/special_phrases_test_content.txt instead of fetching from the internet.

Now we at least test every methods' execution so we won't have the last error you reported again.

.gitignore
```diff
@@ -9,3 +9,4 @@ data/wiki_specialphrases.sql
 data/osmosischange.osc
 
 .vagrant
+data/country_osm_grid.sql.gz
```
AntoJvlt, line 12: As you suggested I have add it to the .gitignore, I hope it doesn't make any trouble if it is in this history.

nominatim/tools/special_phrases.py
```diff
@@ -0,0 +1,259 @@
...
+    def import_from_wiki(self, languages=None):
+        """
+            Iterate through all specified languages and
+            extract corresponding special phrases from the wiki.
+        """
+        #Get all languages to process.
+        languages = self._load_languages() if not languages else languages
+
+        #array for pairs of class/type
+        class_type_pairs = set()
```
lonvia, line 43: Something is very wrong now. This set is never written, so I suspect line 50 does nothing? You might need a test for the full function of import_from_wiki() as well.

NB: the comment in the line above is wrong. This is not an array.

nominatim/tools/special_phrases.py
```diff
@@ -0,0 +1,259 @@
...
+
+        self._create_place_classtype_table_and_indexes(class_type_pairs)
+        self.db_connection.commit()
+        LOG.warning('Import done.')
+
+    def _load_white_and_black_lists(self):
+        """
+            Load white and black lists from phrases-settings.json.
+        """
+        settings_path = str(self.config.config_dir)+'/phrase-settings.json'
```
lonvia, line 58: Have a look at the pathlib module and use it where possible.

test/python/test_tools_import_special_phrases.py
```diff
@@ -0,0 +1,120 @@
...
+"""
+from pathlib import Path
+import pytest
+from nominatim.tools.special_phrases import SpecialPhrasesImporter
+
+TEST_BASE_DIR = Path(__file__) / '..' / '..'
+
+def test_process_amenity_with_operator(special_phrases_importer, getorcreate_amenityoperator_funcs):
+    special_phrases_importer._process_amenity('', '', '', '', 'near')
+    special_phrases_importer._process_amenity('', '', '', '', 'in')
```
lonvia, line 12: You'd still want to check that the expected SQL functions got executed at all. You just don't need to care what the SQL functions themselves do. You can do that by still having a simple 'INSERT .. INTO' into some custom database table and then check that there is a new line in that database table.

test/python/test_tools_import_special_phrases.py
```diff
@@ -0,0 +1,120 @@
...
+    place_id_index_exists = temp_db_conn.index_exists(index_prefix + 'place_id')
+
+    assert centroid_index_exists and place_id_index_exists
+
+def test_create_place_classtype_table(temp_db_conn, placex_table, special_phrases_importer):
+    phrase_class = 'class'
+    phrase_type = 'type'
+    special_phrases_importer._create_place_classtype_table('', phrase_class, phrase_type)
+
+    with temp_db_conn.cursor() as temp_db_cursor:
```
lonvia, line 39: There is a 'temp_db_cursor' fixture for your convenience.

test/python/test_tools_import_special_phrases.py
```diff
@@ -0,0 +1,120 @@
...
+    phrase_type = 'type'
+    table_name = 'place_classtype_{}_{}'.format(phrase_class, phrase_type)
+
+    with temp_db_conn.cursor() as temp_db_cursor:
+        temp_db_cursor.execute('CREATE TABLE {}()'.format(table_name))
+
+    special_phrases_importer._grant_access_to_webuser(phrase_class, phrase_type)
+
+    with temp_db_conn.cursor() as temp_db_cursor:
+        temp_db_cursor.execute(f"""
```
lonvia, line 59: There are still format literals here.

test/python/test_tools_import_special_phrases.py
```diff
@@ -0,0 +1,120 @@
...
+                WHERE table_name='{table_name}' 
+                AND grantee='{def_config.DATABASE_WEBUSER}' 
+                AND privilege_type='SELECT'""")
+        result = temp_db_cursor.fetchone()
+    assert result
+
+def test_create_place_classtype_table_and_indexes(
+        placex_table, getorcreate_amenity_funcs, 
+        getorcreate_amenityoperator_funcs, special_phrases_importer):
+    pairs = {('class1', 'type1'), ('class2', 'type2')}
```
lonvia, line 70: Shouldn't that be a set? Here also, checking for the expected results of the function would have caught this.

test/python/test_tools_import_special_phrases.py
```diff
@@ -0,0 +1,120 @@
...
+def test_process_xml_content(special_phrases_importer, getorcreate_amenity_funcs,
+                             getorcreate_amenityoperator_funcs):
+    special_phrases_importer._process_xml_content(get_test_xml_wiki_content(), 'en')
+
+def mock_get_wiki_content(lang):
+    return get_test_xml_wiki_content()
+
+def test_import_from_wiki(mocker, special_phrases_importer, placex_table, 
+                          getorcreate_amenity_funcs, getorcreate_amenityoperator_funcs):
+    mocker.patch.object(special_phrases_importer, '_get_wiki_content', new=mock_get_wiki_content)
```
lonvia, line 83: Wouldn't a simple monkeypatch do here? What extra functionality does mocker offer?

AntoJvlt: Indeed it works, I though monkeypatch was only for environment and module mocking, I will amend my last commit to remove pytest-mock and make this mocking with monkeypatch.

nominatim/tools/special_phrases.py
AntoJvlt: Yes I am doing big mistakes there I should have tested my code more. By doing it too fast I am not transferring the set throughout functions...

test/python/test_tools_import_special_phrases.py
AntoJvlt: At first I didn't use it because in my SpecialPhrasesImporter class I am only committing changes to the database when the whole import process is done (at the end of "import_from_wiki"), so if I use 2 different connection I get an AssertionError as the result is not found (as my class SpecialPhrasesImporter take in a db_connection and temp_db_cursor get a new one too).

Should I commit more often during the process and use this temp_db_cursor fixture?

test/python/test_tools_import_special_phrases.py
AntoJvlt: I am pretty sure that this syntax create a set, but I agree this syntax is not quite explicit and it only work from python 2.7, I will create it with the syntax set([...]). I will also make the test more efficient.

AntoJvlt:
### Changes
- Amended commit " Encapsulation of tools/special_phrases.py into SpecialPhrasesImporter class and add new tests" to remove pytest-mock and use monkeypatch.
- Fixed errors and hopefully everything is fine now.
- Improved many tests to make them more efficient and added some others tests in order to cover more of the code.
- Cleaned some of the code and added some comments to the tests.

test/python/test_tools_import_special_phrases.py
lonvia: Indeed. There is still something new to learn about Python. Sorry for the noise.

test/python/test_tools_import_special_phrases.py
```diff
@@ -0,0 +1,346 @@
...
+def test_convert_settings_wrong_file(special_phrases_importer):
+    """
+        Test that _convert_php_settings_if_needed() raise an exception
+        if the given file is not a valid file.
+    """
+
+    with pytest.raises(UsageError) as exceptioninfos:
+        special_phrases_importer._convert_php_settings_if_needed('random_file')
+
+    assert str(exceptioninfos.value) == 'random_file is not a valid file.'
```
lonvia, line 62: Minor nitpicking on the use of `pytest.raises`: you don't need an explicit assert. Just the `with pytest.raises()` expression will do. It includes the assert. And if you want to check for a certain message in your exception, use the [match parameter](https://docs.pytest.org/en/stable/assert.html#assertions-about-expected-exceptions).

lonvia (approved):

lonvia:
I've merged this now. Thanks for patiently addressing all comments. For the remaining comment feel free to open a quick new PR.

If you want to continue along this topic, you could have a look at #1362. Now that you know the details about what the special phrases import does, it should be fairly simple to implement the update.

PR #2260, opened 2021-04-11
Fix default languages loading for special phrases import

AntoJvlt:
This PR fix a bug when languages are load during the import of special phrases.

By splitting languages list from the config, a result is always returned even if the config is empty (return ['']), so default language was never returned.


lonvia:
Good catch.

PR #2263, opened 2021-04-12
Implemented auto update of special phrases while importing them

### LINKED ISSUE #1362: New Special Phrases - autoimport - is it possible?

AntoJvlt:
Fix #1362

### What has been added with this PR:

Special phrases are now automatically updated when `nominatim special-phrases --import-from-wiki` is called.
The algorithm does the following when initializing:
- Retrieve and store into the set `words_phrases_to_delete` all special phrases entries from the word tables.
- Retrieve and store into the set `table_phrases_to_delete` all `place_classtype_` tables names from the database.

Then, while handling each entries of the wiki it does:
- Ignore the entry if it is contained in the `words_phrases_to_delete` set.
- Add the entry to the `words_phrases_still_exist`.
(This 2 set mechanism is necessary as there is duplicate entries among the whole wiki).

When looping on each class/type to create the tables it does:
- Ignore the creation of the table and creation of indexes and access-granting, if the table name is already contained in the `words_phrases_to_delete` set.
- Remove the table name entry from the `table_phrases_to_delete` set.

When every phrase has been processed:
- Entries of the `words_phrases_still_exist` are removed from `words_phrases_to_delete`.
- Remove all entries contained in `words_phrases_to_delete` from the word table.
- Remove all tables from the database which are contained in `table_phrases_to_delete`.

I discovered that, for example, Vietnamese phrases contains operator not written in English (so not "in" or "near"). As we only handle english operators, those phrases were never used. It conducted me to add this code: 
`phrase_operator = '-' if phrase_operator not in ('near', 'in') else phrase_operator`
 
I might need to open an issue for that. I didn't as I am not sure if we should handle this or if it should be written in english on the wiki.

- Tests were also added to cover new behaviors.

### DIFF
--- a/test/python/test_tools_import_special_phrases.py
+++ b/test/python/test_tools_import_special_phrases.py
@@ -2,6 +2,7 @@
     Tests for import special phrases methods
     of the class SpecialPhrasesImporter.
 """
+from mocks import MockParamCapture
 from nominatim.errors import UsageError
 from pathlib import Path
 import tempfile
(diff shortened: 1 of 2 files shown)


lonvia:
> I might need to open an issue for that. I didn't as I am not sure if we should handle this or if it should be written in english on the wiki.

The wiki is wrong in this case. The instructions clearly state: For the operator please use either '-', 'in', 'near' or 'named'.

test/python/test_tools_import_special_phrases.py
```diff
@@ -11,6 +12,72 @@
...
+    """
+    with temp_db_conn.cursor() as temp_db_cursor:
+        query ="""
+            INSERT INTO word VALUES(99999, 'lookup_token', 'normalized_word',
+            'place', 'postcode', null, 0, 'near');
+        """
+        temp_db_cursor.execute(query)
+
+    special_phrases_importer._fetch_existing_words_phrases()
+    assert not special_phrases_importer.words_phrases_to_delete
```
lonvia, line 66: This and the previous test only differ in 'house' vs. 'postcode'. Can you make that a single parametrized test (see `@pytest.mark.parametrize`).

test/python/test_tools_import_special_phrases.py
```diff
@@ -11,6 +12,72 @@
...
+        """
+        temp_db_cursor.execute(query)
+
+    assert not special_phrases_importer.words_phrases_to_delete
+    special_phrases_importer._fetch_existing_words_phrases()
+    contained_phrase = special_phrases_importer.words_phrases_to_delete.pop()
+    assert contained_phrase == ('normalized_word', 'class', 'type', 'near')
+
+def test_fetch_existing_words_phrases_housenumber(special_phrases_importer, word_table,
+                                                  temp_db_conn):
```
lonvia, line 35: You really should be using the temp_db_cursor fixture instead of creating the cursor yourself every time. The temp_db_conn fixture is only needed when you test a function that expects a connection as a parameter.

lonvia (commented):
Code looks good but for this part it would be really good to have a test that tests import_from_wiki  as a whole only mocking the download from the wiki. I.e. add a couple of phrase to word, create a couple of classtype tables, run import_from_wiki and check that afterwards the expected entries and tables are added, untouched or deleted.

lonvia:
I've merged this now to avoid conflicts with future changes. Would still be good if you could clean up the tests as suggested in an extra PR.

PR #2280, opened 2021-04-17
Fix regex and sanity check for the import of special phrases and tests cleaning.

AntoJvlt:
## Details for this PR

- Added parametrize test for special cases of postcode and housenumber terms fetching during import of special phrases
- Use of temp_db_cursor fixture when possible. I can't use it when a method of SpecialPhrasesImporter which make changes on the database is called as they use the db connection property of the class. Changes are only committed when the whole process is done (at the end of import_from_wiki) so I can't check for database changes inside tests by using another connection.
- Upgraded test_import_from_wiki to test if wrong phrases from the database are removed (table and word entries) and if correct ones are preserved.

I also discovered that the current regex "occurence_pattern" (and the old one of php) were missing a lot of phrases from the wiki. I don't know why but on the xml exported content, there is cases where no space are added between the content and the "||", like this one: 
image
As there is no space between "near" and "||" the regex wasn't catching the whole phrase. I checked and it doesn't seems to be a problem of how it was added to the wiki. So, I changed the rexeg to support one or multiple spaces before and after "||". Correct me if I am wrong and I will reverse the changes on the regex.

By changing that, importing special phrases started to detect garbage phrases, detected by the `_check_sanity` method. I decided to log a warning when it detects garbage phrase and to not import this phrase, instead of raising an exception which would stop the whole import process. I did this as wrong entries on the wiki should not be handled by us.

### DIFF
--- a/nominatim/tools/special_phrases.py
+++ b/nominatim/tools/special_phrases.py
@@ -153,8 +153,10 @@ class SpecialPhrasesImporter():
         class_matchs = self.sanity_check_pattern.findall(phrase_class)
 
         if len(class_matchs) < 1 or len(type_matchs) < 1:
-            raise UsageError("Bad class/type for language {}: {}={}".format(
-                lang, phrase_class, phrase_type))
+            LOG.warning("Bad class/type for language %s: %s=%s. It will not be imported",
+                        lang, phrase_class, phrase_type)
+            return False
+        return True
 
     def _process_xml_content(self, xml_content, lang):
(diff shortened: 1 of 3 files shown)


lonvia:
Thanks!

The missing spaces is likely one of these errors that has crept into the wiki page over the years. This regex wasn't checked a lot for validity. It might help to add a small statistics in the output of each language how many terms were imported and how many ignored.

PR #2281, opened 2021-04-19
fix index on location_property_tiger (parent_place_id)

changpingc:
Looks like 2af82975cd968ec09683ae5b16a9aa157a7f2176
accidentally renamed an index. Because of the added "if not
exists" clause, the index doesn't get created. This
significantly slows down reverse queries because they now
require full scans on location_property_tiger.

Without this fix, reverse queries can take 8s on a full
planet install on an r5.8xlarge instance in EC2.


changpingc:
cc @darkshredder

lonvia:
Thanks. I'll look into doing a patch release for 3.7 in the next days.

PR #2288, opened 2021-04-22
Replace "nominatim-update" with "nominatim".

RhinoDevel:
If I am not mistaken, the correct command to index imported data via commandline is "nominatim index".


lonvia:
Indeed. Thank you.

The build failure is unrelated.

PR #2291, opened 2021-04-22
Special phrases statistics

AntoJvlt:
## Content of this PR
- Reorganization of the files which handle the import of special phrases in the tools module. It was done to make thing clearer because I introduced a new file.
- Implemented the class `SpecialPhrasesImporterStatistics` which contain counters about operations occurring during the import of special phrases. This class also display basic statistics at the end of each lang processed and at the end of the whole process.
- The class `SpecialPhrasesImporter` contain an instance of `SpecialPhrasesImporterStatistics` and notifies it whenever an operation is done.

It didn't covered this with tests yet as I think that is not a big priority. However, I will add some tests for that later.

### DIFF
--- a/nominatim/tools/special_phrases/importer_statistics.py
+++ b/nominatim/tools/special_phrases/importer_statistics.py
@@ -1,5 +1,5 @@
 """
-    Contain the class which handle statistics for the
+    Contains the class which handles statistics for the
     import of special phrases.
 """
 import logging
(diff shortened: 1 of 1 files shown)


nominatim/tools/special_phrases/importer_statistics.py
```diff
@@ -0,0 +1,130 @@
...
+        LOG.warning('Statistics for the import of %s:', lang)
+        LOG.warning('- %s phrases were invalid.', self.lang_phrases_invalid)
+        if self.lang_phrases_invalid > 0:
+            LOG.warning('  Those invalid phrases have been skipped.')
+        LOG.warning('- %s phrases were ignored as they are already in the database',
+                    self.lang_phrases_ignored)
+        LOG.warning('- %s phrases were added to the database', self.lang_phrases_added)
+        LOG.warning('====================================================================')
+
+        self._set_lang_values_to_0()
```
lonvia, line 130: Both notifications are quite verbose now. I would suggest to move the message you have now in notify_import_done() and notify_current_lang_done() to LOG.info() and for LOG.warning only print a one-liner with the number of invalid phrases (if there are any).

lonvia (commented):
Apologies, I had to do quite a few Python 3.5-compatibility fixes on master including in special_phrases.py in preparation for the 3.7.1 release. That has caused a couple of conflicts. Could you please rebase this PR on master?

AntoJvlt:
I made the changes and I rebased this branch on master.

Issue #1028, opened 2018-04-30
Place with postalcode=... not found


foodev:
When I search for a place with `postalcode=16356&city=Werneuchen` I won't get any results:

https://nominatim.openstreetmap.org/search.php?format=json&countrycodes=de&postalcode=16356&city=Werneuchen

```json
[]
```

But when I remove the `postalcode` parameter, I get the correct result (among others):

https://nominatim.openstreetmap.org/search.php?format=html&countrycodes=de&city=Werneuchen

```json
[{"place_id":"179025849","licence":"Data © OpenStreetMap contributors, ODbL 1.0. https:\/\/osm.org\/copyright","osm_type":"relation","osm_id":"1318353","boundingbox":["52.5784223","52.7136828","13.6374307","13.8587715"],"lat":"52.6334769","lon":"13.7340939","display_name":"Werneuchen, Landkreis Barnim, Brandenburg, Deutschland","class":"boundary","type":"administrative","importance":0.21723658035362,"icon":"https:\/\/nominatim.openstreetmap.org\/images\/mapicons\/poi_boundary_administrative.p.20.png"}]
```

Though the corresponding node has the postcode:

https://nominatim.openstreetmap.org/details.php?place_id=157634

JLuc:
I confirm same issues for french area cf https://github.com/openstreetmap/Nominatim/issues/1239 : 

In some cases, nominatim **fails when being provided both city and postalcode** : 
https://nominatim.openstreetmap.org/search?postalcode=40430&city=Argelouse&format=json&limit=1&countrycode=fr

But it succeeds and finds the right place when provided **only with city name** :
https://nominatim.openstreetmap.org/search?city=Argelouse&format=json&limit=1&countrycode=fr

```
[{ "place_id": "199026058", 
"licence": "Data © OpenStreetMap contributors, ODbL 1.0. https://osm.org/copyright",
"osm_type": "relation", "osm_id": "2114363",
"boundingbox": [ "44.345529", "44.4001763", "-0.672487", "-0.599563" ],
"lat": "44.3671374", "lon": "-0.6449692",
"display_name": "Argelouse, Mont-de-Marsan, Landes, Nouvelle-Aquitaine, France métropolitaine, 40430, France",
"class": "boundary", "type": "administrative", 
"importance": 0.637409167502591,
"icon": "https://nominatim.openstreetmap.org/images/mapicons/poi_boundary_administrative.p.20.png" }]
```

[...]

PR #2300, opened 2021-04-27
Fixed `postalCode` parameter case

mscheibler:
The `postalCode` parameter seems to be case sensitive. Without the CamelCase writing the results from OSM are empty.

[working request](https://nominatim.openstreetmap.org/search?city=Dresden&format=json&accept-Language=de&postalCode=01328&country=DE)

[failing Request](https://nominatim.openstreetmap.org/search?city=Dresden&format=json&accept-Language=de&postalcode=01328&country=DE)

### DIFF
--- a/docs/api/Search.md
+++ b/docs/api/Search.md
@@ -37,7 +37,7 @@ The search term may be specified with two different sets of parameters:
 * `county=<county>`
 * `state=<state>`
 * `country=<country>`
-* `postalcode=<postalcode>`
+* `postalCode=<postalcode>`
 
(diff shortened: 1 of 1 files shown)


mtmail:
It's case sensitive, but the lowercase is the correct one. Setting `postalCode=` is the same as not setting the parameter. What you see is a problem with the search logic. Sometimes setting a postalcode leads to worse result than ommitting it, e.g. https://github.com/osm-search/Nominatim/issues/1028

mscheibler:
@mtmail You are right. This is a bit confusing

PR #2310, opened 2021-05-04
2nd try: Add hint about replication update & recheck intervals being in seconds.

RhinoDevel:
Add hint about replication update & recheck intervals being in seconds in documentation and environment file.


lonvia:
No problem. Thanks for rebasing.

PR #2317, opened 2021-05-06
Add array_key_last function for PHP <7.3

### LINKED ISSUE #2316: Some searches result in Call to undefined function Nominatim\array_key_last()
On a  plain Nominatim install with Ubuntu 18.04 (i.e. PHP 7.2.24), some searches fail with

    { "error": {"code": 0,"message": "Call to undefined function Nominatim\\array_key_last()" } }

The reason appears to be that this function is only supported from PHP 7.3 on. Yet Nominatim claims to only require "PHP >= 7.0".

Not all search terms trigger this bug. One search term that would currently trigger this is "2818 skelton st, richmond, va".

woodpeck:
This patch adds an array_key_last function if it doesn't yet exist, fixes #2316. It is tested on PHP 7.2.24 but not PHP 7.3.

### DIFF
--- a/lib-php/lib.php
+++ b/lib-php/lib.php
@@ -227,3 +227,11 @@ function closestHouseNumber($aRow)
 
     return max(min($aRow['endnumber'], $iHn), $aRow['startnumber']);
 }
+
+if (!function_exists('array_key_last'))
+{
+    function array_key_last(array $array) 
+    {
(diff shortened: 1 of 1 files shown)


lonvia:
Merge with style fixes as fe39185894cc7c7f1d4d50b15cc96476b35641d9. I was hoping to enable PHP 7.1 on the CI but that is a bigger project.

PR #2321, opened 2021-05-10
CSV import for special phrases and loader refactoring

### LINKED ISSUE #2311: Add CSV import for special phrases
We currently only have the ability to import special phrases from the wiki. It would be nice if the user could supply a simple CSV instead with their own custom phrases. For the moment these CSV should follow the wiki structure and expect the following columns: phrase, key, value, operator. The reader should simply ignore any other columns, so that we can require additional optional data in the future.

This CSV importer would also be needed [for the BDD tests](https://github.com/osm-search/Nominatim/blob/3c67bae86884bdd4b7f053b553fe697066b5c887/test/bdd/steps/nominatim_environment.py#L195). Now that the task of importing the special phrases into the word table has been outsourced to the tokenizer, the simple approach of having an SQL file does not work anymore.

@AntoJvlt Is that a task I could interest you in?

AntoJvlt:
This PR resolves #2311 

### What have been done

- I refactored the special phrases importer module a bit. I extracted the logic which is responsible of loading special phrases from the wiki. Therefore, I implemented a strategy pattern with the SPLoader class so that we can easily switch the way phrases are loaded from an external source. Thus, the SPImporter class is now only responsible of importing the phrases into the database.
- The SPLoader class is an iterator, so that we can fetch phrases from multiple sources, it is usefull for the wiki import and we might imagine a world where we could have multiple csv files.
- I implemented the SPCsvLoader which allows to load special phrases from a csv file.
- I added a SpecialPhrase class in order to have a common class model to transfer through each unit.

### Possible issues

- Should we handle multiple types of csv delimiters? Like ',' or ';' currently only ',' is supported.
- Currently, when we choose to import from a csv file, all phrases which are not in the file will be removed from the database (auto update), maybe for this functionality we should have an option to only add new phrases?

### To do before merge

I opened this PR for early reviews if you want, I am not done yet and it may be better if I finish everything concerning the csv import before the final merge.

- Update tests and add tests for the new Loader system and especially for the csv loader. Currently tests were modified only to make them pass, they should be updated accordingly to the refactoring that I have done.
- Documentation of the new command parameter `--import-from-csv`.


nominatim/tools/special_phrases/sp_csv_loader.py
```diff
@@ -0,0 +1,54 @@
...
+import csv
+import os
+from nominatim.tools.special_phrases.special_phrase import SpecialPhrase
+from nominatim.tools.special_phrases.sp_loader import SPLoader
+from nominatim.errors import UsageError
+
+class SPCsvLoader(SPLoader):
+    """
+        Base class for special phrases loaders.
+        Handle the loading of special phrases from external sources.
```
lonvia, line 15: Comment looks copy/pasted.

nominatim/tools/special_phrases/sp_loader.py
```diff
@@ -0,0 +1,16 @@
...
+    """
+        Base class for special phrases loaders.
+        Handle the loading of special phrases from external sources.
+    """
+    def __iter__(self):
+        return self
+
+    @abstractmethod
+    def __next__(self):
+        pass
```
lonvia, line 16: As Python is not a strictly typed language, you don't need a common base class for your loader class. You can rely on duck typing instead and simply make sure that they implement all functions needed for the loader. So in this case, the SP*Loader classes can simply inherit from `collections.abc.Iterator` to get the `__iter__` implementation. And this SPLoader class can be dropped.

nominatim/tools/special_phrases/sp_wiki_loader.py
```diff
@@ -0,0 +1,71 @@
...
+    def __init__(self, config, languages=None):
+        if languages is not None and not isinstance(languages, list):
+            raise TypeError('The \'languages\' parameter should be of type list.')
+        super().__init__()
+        self.config = config
+        #Compile the regex here to increase performances.
+        self.occurence_pattern = re.compile(
+            r'\| *([^\|]+) *\|\| *([^\|]+) *\|\| *([^\|]+) *\|\| *([^\|]+) *\|\| *([\-YN])'
+        )
+        self.languages = self._load_languages() if not languages else languages
```
lonvia, line 24: Make a copy of the `languages` parameter here. You later manipulate the list and that might confuse the caller. (As a side effect, you can drop the check in line 16.)

nominatim/tools/special_phrases/sp_wiki_loader.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def __next__(self):
+        if not self.languages:
+            raise StopIteration
+
+        lang = self.languages.pop(0)
+        loaded_xml = SPWikiLoader._get_wiki_content(lang)
```
lonvia, line 31: General comment: preferably use the self notation, even when calling static methods. I.e. `self._get_wiki_content(lang)`. Python will do the right thing and it becomes really easy to switch between using static and member methods.

nominatim/tools/special_phrases/sp_csv_loader.py
```diff
@@ -0,0 +1,54 @@
...
+    def __next__(self):
+        if self.has_been_read:
+            raise StopIteration()
+
+        self.has_been_read = True
+        SPCsvLoader.check_csv_validity(self.csv_path)
+        return SPCsvLoader.parse_csv(self.csv_path)
+
+    @staticmethod
+    def parse_csv(csv_path):
```
lonvia, line 31: It seems odd that you make this function and the one below a static method and then hand in a member variable. Why not make them normal member methods that have access to self.csv_path?

nominatim/tools/special_phrases/sp_importer.py
```diff
@@ -101,89 +99,48 @@ def _load_white_and_black_lists(self):
...
-            if (
-                    phrase_class in self.white_list.keys() and
-                    phrase_type not in self.white_list[phrase_class]
-            ):
-                continue
+        #blacklisting: disallow certain class/type combinations
+        if (
+            phrase.p_class in self.black_list.keys() and
+            phrase.p_type in self.black_list[phrase.p_class]
+        ): return None
```
lonvia, line 127: Now I do have to complain about style. I'm not a big fan about the PHP multiline-styling with brackets on separate lines, but I can live with that. But having the `return None` on its own line is mandatory.

nominatim/tools/special_phrases/sp_importer.py
```diff
@@ -101,89 +99,48 @@ def _load_white_and_black_lists(self):
...
+            self.statistics_handler.notify_one_phrase_invalid()
+            return None
 
-            self.word_phrases.add((phrase_label, phrase_class,
-                                   phrase_type, phrase_operator))
+        self.word_phrases.add((phrase.p_label, phrase.p_class,
+                               phrase.p_type, phrase.p_operator))
 
-        return class_type_pairs
+        return set({(phrase.p_class, phrase.p_type)})
```
lonvia, line 143: Doesn't need to be a set. A set can be updated from any iterable.

lonvia (commented):
Looks good overall. I added a couple of minor comments inline.

Regarding the CSV delimiter: I thought that the Python csv module would try to autodetect when you leave out the delimiter parameter in the DictReader() constructor. If not, then just supporting comma is fine for now.

Having a command-line parameter to switch between adding and replacing sounds like a good idea.

nominatim/tools/special_phrases/sp_csv_loader.py
AntoJvlt: Indeed, I wanted to make them as independent as possible but their logic is highly related to the SPCsvLoader class so yes, I will make them normal member methods.

nominatim/tools/special_phrases/sp_wiki_loader.py
AntoJvlt: I generally prefer this self notation too, but I had some trouble with python and this notation in the past. Currently it seems to not be causing any troubles so I switched to the self notation.

nominatim/tools/special_phrases/sp_importer.py
AntoJvlt: I don't know if I am missing something or another better style to use. I had no choice to do this as the line would be too long, is there a better way to handle this?

lonvia: I usually use slashes in such cases:

```
        if phrase.p_class in self.black_list.keys() \
           and phrase.p_type in self.black_list[phrase.p_class]:
            return None
```

I find that more readable. But it's a matter of taste.

What I insist on: line break after the colon. (I.e., the style until now was acceptable.)

nominatim/tools/special_phrases/sp_wiki_loader.py
AntoJvlt: I am sorry I don't understand what you mean here, could you elaborate please?

nominatim/tools/special_phrases/sp_importer.py
AntoJvlt: Should I return a list? What would be the difference with returning a set?
The problem I got here is that if I only return the tuple (phrase.p_class, phrase.p_type), when I update the main set with this tuple it gets exploded inside the set: {p_class, p_type} instead of {(p_class, p_type)}. That's why I return a set containing the well constructed tuple.

nominatim/tools/special_phrases/sp_importer.py
AntoJvlt: Thanks, I didn't know this notation and I find it better.
For the `return None` this is my bad, it should be on a new line.

AntoJvlt:
### Update
- I made the changes
- Added and updated tests
- Added --no-replace parameter to only add new phrases and not update the other ones
- Updated documentation to add the new command and parameter

docs/admin/Import.md
```diff
@@ -277,7 +277,60 @@ If you want to be able to search for places by their type through
...
+the argument `--no-replace` to the import command. For example:
+
+```sh
+nominatim special-phrases --import-from-csv <csv file> --no-replace
+```
+
+This will add the phrases present in the csv file into the database without
+removing the other ones.
+
+## Installing Tiger housenumber data for the US
```
lonvia, line 309: There is a merge error here. This section has moved to `docs/admin/Customization.md` in #2328.

I would also suggest that you move the extended documentation about the CSV format for special phrases to that file, so that the Import chapter does not become too long.

nominatim/tools/special_phrases/sp_importer.py
lonvia: I would have returned a single-item tuple of tuple, which is a bit cheaper: `((phrase.p_class, phrase.p_type), )`

However, as you always return a single item, the better solution here would be to simple return the single item: `return (phrase.p_class, phrase.p_type)` and above, where the function is called, use `add()` instead of `update()`. `add()` expects a single item, so it will do what you want.

nominatim/tools/special_phrases/sp_wiki_loader.py
lonvia: Lists are handed in by reference to a function. So, when I do something like:

```
langs = ['de' , 'en']
loader = SPWikiLoader(config, langs)
print(langs)
```
then, any modification you do to `languages` in your constructor, will be visible outside in the print, because they work on the same list. So when you consume `self.languages` with `pop()` below, the caller that called your constructor will see the list disappear as well.

This is not a problem in your particular code because the caller doesn't keep the list. It's still not good style because it is surprising and can lead to subtle problems in the future. The general rule is: if you keep a list you get handed into a function and/or modify it for your own private purposes, you should make a copy that does not interfere with what the caller later does with the list it has handed in. (Or make it very clear in the documentation that you modify/destroy the list, i.e. take ownership over it. But that would unnecessarily complicate things in this particular case.)

AntoJvlt: Oh ok thanks, that make a lot more sense now. I was quite confuse at first as I didn't understand that you wanted me to copy the list.

AntoJvlt:
I made the small changes, corrected my documentation mistakes and moved the doc about the CSV import into Customization.md.

docs/admin/Customization.md
```diff
@@ -69,3 +69,33 @@ entire US adds about 10GB to your database.
...
+
+But, it is also possible to import some phrases from a csv file. 
+To do so, you have access to the following command:
+
+```sh
+nominatim special-phrases --import-from-csv <csv file>
+```
+
+Note that the 2 previous import commands will update the phrases from your database.
+This mean that if you import some phrases from a csv file, only the phrases
```
lonvia, line 89: mean -> means

docs/admin/Import.md
```diff
@@ -277,7 +277,12 @@ If you want to be able to search for places by their type through
...
-    nominatim special-phrases --import-from-wiki
+```sh
+nominatim special-phrases --import-from-wiki
+```
 
 Note that this command downloads the phrases from the wiki link above. You
 need internet access for the step.
+
+You can also import special phrases from a csv file, for more 
+informations please read the [Customization chapter](Customization.md).
```
lonvia, line 288: informations -> information

lonvia (commented):
Two tiny typos, otherwise all good.

Issue #2267, opened 2021-04-13
Disable search API when database was imported with --reverse-only


lonvia:
When the database is a reverse-only database (i.e. the table search_name) is missing, then the refresh-website commend should create a dummy script for the endpoint `search.php` that just returns a 404 with the explanation that the installation does not support forward searching.

Also: the [documentation](https://nominatim.org/release-docs/develop/admin/Import/#testing-the-installation) recommends the `query.php` endpoint for testing. There should be a hint there to use `reverse.php` when importing with `--reverse-only`.

darkshredder:
I would like to take up this issue if it is available to work on.

lonvia:
The website scripts are always [(re)created as the last step of the import process](https://github.com/osm-search/Nominatim/blob/40cb17d2995baaba6bec299d4e6dc65b25dbdc28/nominatim/clicmd/setup.py#L142). So as long as you make the change in the `setup_website` function in `refresh.py`, the order does not matter.

I'd like to avoid runtime checks because the reverse script needs to be really fast.

PR #2323, opened 2021-05-13
Feat: Disabled search API for --reverse-only imports

darkshredder:
Fixes: #2267 

### This PR involves the following changes:

- Removed search query when --reverse-only data is imported.
- Updated documentation for adding a note to use a reverse query for testing --reverse-only imports.

### DIFF
--- a/lib-php/website/reverse-only-search.php
+++ b/lib-php/website/reverse-only-search.php
@@ -1,0 +1,12 @@
+<?php
+
+require_once(CONST_LibDir.'/init-website.php');
+require_once(CONST_LibDir.'/ParameterParser.php');
+
+$oParams = new Nominatim\ParameterParser();
+
+// Format for output
+$sOutputFormat = $oParams->getSet('format', array('xml', 'json', 'jsonv2', 'geojson', 'geocodejson'), 'jsonv2');
+set_exception_handler_by_format($sOutputFormat);
+
+throw new Exception('Reverse-only import does not support forward searching.', 404);
--- a/docs/admin/Import.md
+++ b/docs/admin/Import.md
@@ -252,6 +252,9 @@ to verify that your installation is working. Go to
 `http://localhost:8088/status.php` and you should see the message `OK`.
 You can also run a search query, e.g. `http://localhost:8088/search.php?q=Berlin`.
 
+Note that search query is not supported for reverse-only imports. You can run a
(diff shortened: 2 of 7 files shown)


lib-php/website/reverse-only-search.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+require_once(CONST_LibDir.'/ParameterParser.php');
+
+$oParams = new Nominatim\ParameterParser();
+$sOutputFormat = $oParams->getSet('format', array('text', 'json'), 'text');
```
lonvia, line 6: This should support the same output formats as normal search. Otherwise people get confusing ' wrong format'  responses even though that is not the problem.

lib-php/website/reverse-only-search.php
```diff
@@ -0,0 +1,19 @@
...
+    $aResponse = array(
+                  'status' => 404,
+                  'message' => 'Reverse-only import does not support forward searching.',
+                  'software_version' => CONST_NominatimVersion
+                 );
+    javascript_renderData($aResponse);
+} else {
+    header('HTTP/1.0 404 Not Found');
+    echo 'Error: Reverse-only import does not support forward searching.';
+}
```
lonvia, line 19: Instead of formatting this yourself, you could also use `set_exception_handler_by_format()` and then simply throw a Exception.

lib-php/website/reverse-only-search.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+require_once(CONST_LibDir.'/ParameterParser.php');
```
lonvia, line 3: This is missing at least: require_once(CONST_LibDir.'/init-website.php');
I get PHP errors when testing.

lib-php/website/reverse-only-search.php
darkshredder: Sure, Done

lib-php/website/reverse-only-search.php
darkshredder: Done.

lib-php/website/reverse-only-search.php
darkshredder: set_exception_handler_by_format() also requires it, so had to add it. Not sure why I was not getting PHP errors.

PR #2338, opened 2021-05-19
fix: a more compatible way to get request path without the query string

mogita:
This PR fixes the problem of using Nominatim behind the Apache server can cause variable `$_SERVER['DOCUMENT_URI']` being empty.

A little details on this: `DOCUMENT_URI` is not a standard PHP server environment variable making the `if` statement unreliable. To get a clean path of the request without query string, PHP offers a built-in function [`parse_url()`](https://www.php.net/manual/en/function.parse-url.php).

This is a related problem to https://github.com/osm-search/Nominatim/pull/2337

Thanks for the review!


lonvia:
I never got around doing the necessary background research on this, sorry. Given that we're now fading out the PHP code, it's a bit late now. Closing as won't fix.

PR #2355, opened 2021-06-01
Documentation update

AntoJvlt:
A few documentation updates.

About the .env file, I am not absolutely sure of me, I am maybe missing something like the possibility of creating our own .env file for the configuration instead of using the one under Nominatim/settings.

### DIFF
--- a/docs/admin/Import.md
+++ b/docs/admin/Import.md
@@ -41,14 +41,17 @@ all commands from the project directory.
 ### Configuration setup in `.env`
 
 The Nominatim server can be customized via an `.env` configuration file in the
-project directory. This is a file in [dotenv](https://github.com/theskumar/python-dotenv)
+Nominatim's source code directory. This is a file in [dotenv](https://github.com/theskumar/python-dotenv)
 format which looks the same as variable settings in a standard shell environment.
 You can also set the same configuration via environment variables. All
 settings have a `NOMINATIM_` prefix to avoid conflicts with other environment
 variables.
 
 There are lots of configuration settings you can tweak. Have a look
-at `settings/env.default` for a full list. Most should have a sensible default.
+at `Nominatim/settings/env.default` for a full list. Most should have a sensible default.
+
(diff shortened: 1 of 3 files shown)


docs/admin/Import.md
```diff
@@ -41,14 +41,17 @@ all commands from the project directory.
...
 You can also set the same configuration via environment variables. All
 settings have a `NOMINATIM_` prefix to avoid conflicts with other environment
 variables.
 
 There are lots of configuration settings you can tweak. Have a look
-at `settings/env.default` for a full list. Most should have a sensible default.
+at `Nominatim/settings/env.default` for a full list. Most should have a sensible default.
+
+After editing the file, make sure to rebuild nominatim with the command `sudo make install`
+from the build folder. See [Building Nominatim](Installation#building-nominatim).
```
lonvia, line 54: This was really meant as written here: you create a file .env in the project directory. If you want to set a variable to something different than the default, you add a line in there with the custom setting.

docs/admin/Import.md
```diff
@@ -193,12 +196,15 @@ can be found in the development section.
...
 
 Download the data to import. Then issue the following command
-from the **build directory** to start the import:
+from the **project directory** to start the import:
 
 ```sh
 nominatim import --osm-file <data file> 2>&1 | tee setup.log
 ```
 
+The **project directory** is the one that you have set up at the beggining.
```
lonvia, line 205: begging -> beginning

docs/admin/Import.md
AntoJvlt: Okay, yes, that makes more sense. I tried this and it didn't work for me, that's why I was confused.

AntoJvlt:
I accidentally deleted the remote branch instead of force push. The new branch has been committed on my fork, maybe you can re-open this PR @lonvia?

Thanks

lonvia:
Doesn't look like I can reopen. Just open a new one.

PR #2360, opened 2021-06-07
Use place instead of placex to compute postcodes

### LINKED ISSUE #2049: Calculate postcodes from place table
If we could compute the `location_postcode` table directly from `place` instead of `placex`, it wouldn't be necessary anymore to copy over the about 3M postcode helper points into placex. That would be a nice save of space and would allow to get rid of a couple of special cases in the placex triggers.

The main issue is that we don't have country_codes in the `place` table, so they would need to be computed on the fly. It largely depends on how much this degrades performance of postcode updates, if this idea is feasible.

Pointers to the relevant code:
* [calculatePostcodes()](https://github.com/osm-search/Nominatim/blob/0d9fe6e49c3cee131d2fd0c2c0ad2e59065621f9/lib-php/setup/SetupClass.php#L70) (should be ported to Python first, see #2220)
* [update postcode table](https://github.com/osm-search/Nominatim/blob/master/lib-sql/update-postcodes.sql)

AntoJvlt:
Resolves #2049

Elements of `class=place` and `type=postcode` are no longer added to the `placex` table.
So, the `place` table is now used to compute postcodes.

I am not sure on how to handle [this part](https://github.com/osm-search/Nominatim/blob/c6fe91bfa54a501261852464a1296b338fab947e/lib-sql/functions/place_triggers.sql#L204-L209) inside the place insert trigger, because of the special type `postal_code`. Should I keep it as it is and only remove the check for `type=place`?

Also, it might be good to add something for the "migration" when the release will be out.


nominatim/tools/postcodes.py
```diff
@@ -195,3 +197,15 @@ def update_postcodes(dsn, project_dir, tokenizer):
...
+
+def can_compute(dsn):
+    """
+        Check that the place table exists so that
+        postcodes can be computed.
+    """
+    with connect(dsn) as conn:
+        with conn.cursor() as cur:
+            cur.execute("""
+                select exists(select 1 from information_schema.tables where table_name='place')
```
mtmail, line 209: There's `table_exists` in `nominatim/db/connection.py`. Would that be a possibility to use here?

AntoJvlt: That's done. This could be checked without the can_compute() function, but I prefer to keep it for possible further checks and clarity.

lib-sql/functions/place_triggers.sql
```diff
@@ -77,6 +77,12 @@ BEGIN
 
   ELSE -- insert to placex
 
+    -- Pure postcodes are never queried from placex so we don't add them.
+    -- location_postcodes is filled from the place table directly.
+    IF NEW.class = 'place' AND NEW.type = 'postcode' THEN
+      RETURN NEW;
```
lonvia, line 83: We have to handle a type change here. For example if an OSM object was tagged `place=town,addr:postcode=123` and somebody deletes the `place` tag, then the final type changes from `town` to `placecode` and you need to make sure to remove the old town entry from placex. The same delete as in line 105 should do.

We really need to add tests to the BDD test suite for these kind of corner cases.

nominatim/tools/postcodes.py
```diff
@@ -195,3 +197,11 @@ def update_postcodes(dsn, project_dir, tokenizer):
             conn.commit()
 
         analyzer.update_postcodes_from_db()
+
+def can_compute(dsn):
+    """
+        Check that the place table exists so that
+        postcodes can be computed.
+    """
+    with _Connection(dsn) as conn:
```
lonvia, line 206: Please use `with connect(dsn) ...`.

nominatim/tools/postcodes.py
```diff
@@ -163,17 +163,19 @@ def update_postcodes(dsn, project_dir, tokenizer):
...
-                                 GROUP BY country_code, pc) xx
-                               WHERE pc is not null
-                               ORDER BY country_code, pc""")
+                cur.execute("""
+                SELECT cc as country_code, pc, ST_X(centroid), ST_Y(centroid)
+                FROM (
+                    SELECT 
+                        COALESCE(plx.country_code, get_country_code(ST_Centroid(pl.geometry))) as cc,
+                        token_normalized_postcode(pl.address->'postcode') as pc,
+                        ST_Centroid(ST_Collect(ST_Centroid(pl.geometry))) as centroid
```
lonvia, line 172: I wonder if it helps with performance to use COALSECE also with the centroid computation: `COALESCE(plx.centroid, ST_Centroid(pl.geometry)`

test/python/test_tools_postcodes.py
```diff
@@ -58,15 +63,17 @@ def postcode_table(temp_db_conn, placex_table, word_table):
...
     assert not postcode_table.row_set
 
 
-def test_postcodes_add_new(dsn, placex_table, postcode_table, tmp_path, tokenizer):
-    placex_table.add(country='xx', geom='POINT(10 12)',
-                     address=dict(postcode='9486'))
+def test_postcodes_add_new(dsn, postcode_table, placex_table, place_row,
+                           tmp_path, tokenizer):
+    placex_table.add(country='xx', geom='POINT(10 12)')
+    place_row(geom='SRID=4326;POINT(10 12)', address=dict(postcode='9486'))
```
lonvia, line 76: I might have missed it but do you also test the case somewhere where placex does not a corresponding line?

Also, consider adding a [pytest fixture with a factory](https://docs.pytest.org/en/stable/fixture.html#factories-as-fixtures)  for adding rows with postcodes, i.e the fixture would that wraps these two lines.

test/python/test_tools_postcodes.py
```diff
@@ -185,3 +197,8 @@ def test_postcodes_extern_bad_number(dsn, placex_table, postcode_table,
 
     assert postcode_table.row_set == {('xx', 'AB 4511', 10, 12),
                                       ('xx', 'CD 4511', -10, -5)}
+
+def test_can_compute(dsn, temp_db_cursor):
+    assert not postcodes.can_compute(dsn)
+    temp_db_cursor.execute('CREATE TABLE place()')
```
lonvia, line 203: Please use the `table_factory` fixture.

lonvia:
> I am not sure on how to handle this part inside the place insert trigger, because of the special type postal_code. Should I keep it as it is and only remove the check for type=place?

It is still valid for [postcode areas](https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dpostal_code) which are different from the postcodes derived from the extra postcode attributes that your change handles. If there are no more bugs in the code then those should only be `class=boundary` and `type=postal_code` now.

(BTW Ignore the random failures of CodeCov. That's their fault.)

test/python/test_tools_postcodes.py
AntoJvlt: I will add one test specially to test the call to get_country_code.

nominatim/tools/postcodes.py
AntoJvlt: I tried and it doesn't improve execution time

lib-sql/functions/place_triggers.sql
AntoJvlt: Concerning the BDD tests, I need to have a deeper look on how to make them right. As I try to focus on my priorities for the GSoC, is it ok to have another PR for those tests later? (Of course I will finish to correct this PR till it's ok to be merged)

nominatim/tools/postcodes.py
lonvia: Ok, thanks for testing.

lib-sql/functions/place_triggers.sql
lonvia: I agree. Let's leave that for a later PR.

lib-sql/functions/place_triggers.sql
```diff
@@ -92,6 +92,16 @@ BEGIN
     -- Get the existing place_id
     select * from placex where osm_type = NEW.osm_type and osm_id = NEW.osm_id and class = NEW.class and type = NEW.type INTO existingplacex;
 
+    -- Pure postcodes are never queried from placex so we don't add them.
+    -- location_postcodes is filled from the place table directly.
+    IF NEW.class = 'place' AND NEW.type = 'postcode' THEN
+      -- Remove old placex entry if the type changed to postcode.
+      IF existingplacex.type IS NOT NULL AND existingplacex.type != 'postcode' THEN
```
lonvia, line 99: Let's leave out the second condition. Without it, an existing database can slowly 'clean out' itself.

nominatim/tools/postcodes.py
```diff
@@ -163,17 +163,19 @@ def update_postcodes(dsn, project_dir, tokenizer):
...
-                                 GROUP BY country_code, pc) xx
-                               WHERE pc is not null
-                               ORDER BY country_code, pc""")
+                cur.execute("""
+                SELECT cc as country_code, pc, ST_X(centroid), ST_Y(centroid)
+                FROM (
+                    SELECT 
+                        COALESCE(plx.country_code, get_country_code(ST_Centroid(pl.geometry))) as cc,
+                        token_normalized_postcode(pl.address->'postcode') as pc,
+                        COALESCE(ST_Centroid(ST_Collect(plx.centroid)), ST_Centroid(ST_Collect(ST_Centroid(pl.geometry)))) as centroid
```
lonvia, line 172: The coalesce must go into the inner part to have an effect: `ST_Centroid(ST_Collect(COALESCE(plx.centroid, ST_Centroid(pl.geometry)))`

test/python/test_tools_postcodes.py
```diff
@@ -58,80 +63,75 @@ def postcode_table(temp_db_conn, placex_table, word_table):
...
-def test_postcodes_ignore_empty_country(dsn, placex_table, postcode_table, tmp_path, tokenizer):
-    placex_table.add(country=None, geom='POINT(10 12)',
-                     address=dict(postcode='AB 4511'))
-
+def test_postcodes_ignore_empty_country(dsn, postcode_table, tmp_path,
+                                        insert_implicit_postcode, tokenizer):
+    insert_implicit_postcode(1, None, 'POINT(10 12)', dict(postcode='AB 4511'))
     postcodes.update_postcodes(dsn, tmp_path, tokenizer)
-
+    print(postcode_table.row_set)
```
lonvia, line 117: Leftover debugging code?

nominatim/clicmd/refresh.py
```diff
@@ -52,13 +52,17 @@ def run(self, args):
...
+            if postcodes.can_compute(args.config.get_libpq_dsn()):
+                LOG.warning("Update postcodes centroid")
+                tokenizer = self._get_tokenizer(args.config)
+                postcodes.update_postcodes(args.config.get_libpq_dsn(),
+                                           args.project_dir, tokenizer)
+                indexer = Indexer(args.config.get_libpq_dsn(), tokenizer,
+                                  args.threads or 1)
+                indexer.index_postcodes()
+            else:
+                LOG.error("The place table doesn\'t exists. " \
```
lonvia, line 64: exists -> exist

AntoJvlt:
Usage of `COALESCE` to get the centroid of each postcode seems to increase performance by about 3 mins of execution time.

PR #2414, opened 2021-08-05
when replication --once set and no new updates, then exit

mtmail:
Scenario:
1. Database is several days behind, daily updates, no automatic updates (e.g. cronjob) set
2. Run `nominatim replication --once`. This will run all updates fine `Update completed. Import:
1:37:16. Indexing: 0:44:18 Total: 2:21:34. Remaining backlog: 14:37:50.`
3. Run `replication replication --check-for-updates`. Prints `Database is up to date` as expected.
4. Run `nominatim replication --once` again. It starts a long sleep `Sleeping for 33728 sec before
next update.`

I would expect `--once` doesn't sleep and exits immediately.

### DIFF
--- a/nominatim/clicmd/replication.py
+++ b/nominatim/clicmd/replication.py
@@ -90,6 +90,7 @@ class UpdateReplication:
                       update_interval=args.config.get_int('REPLICATION_UPDATE_INTERVAL'),
                       import_file=args.project_dir / 'osmosischange.osc',
                       max_diff_size=args.config.get_int('REPLICATION_MAX_DIFF'),
+                      sleep_between_intervals=not args.once,
                       indexed_only=not args.once)
 
         # Sanity check to not overwhelm the Geofabrik servers.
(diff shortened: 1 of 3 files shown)


lonvia:
This might be expected behaviour. What did you set the `NOMINATIM_REPLICATION_*` to?

lonvia:
I gave that some more thought and would like to keep the current behaviour. The reason is that it should be possible to use the once mode in a loop where you want to run some additional stuff after updating, i.e:

```
while 1
do
  nominatim replication --once
  ./my-fancy-postprocessing.sh
done
```

When `--once` respects the NOMINATIM_REPLICATION_UPDATE_INTERVAL this loop doesn't need to do the sleep externally. (For example, the [photon update script](https://github.com/komoot/photon/blob/60a975414131fed584100f73b205722732886d93/continuously_update_from_nominatim.sh) can be simplified.)

If you want `--once` to check and return immediately, simply set the update interval to 0, either permanently in your `.env` or temporarily, when running updates:

```
NOMINATIM_REPLICATION_UPDATE_INTERVAL=0 nominatim --once
```

Does that make sense?

mtmail:
Makes sense. I had the `NOMINATIM_REPLICATION_UPDATE_INTERVAL` set to 24h which was way too high.

PR #2427, opened 2021-08-17
Move US state hack into legacy tokenizer

lonvia:
The query code has a very ugly hack where it replaces 'IL', 'AL' and 'LA' with the respective US states to avoid that normalization detects the state codes as stop words and removes them. This hack is no longer necessary with the new ICU tokenizer because we don't do stop word elimination anymore. As a general rule, it should be the tokenizer's decision how to handle these difficult cases. So move the code into the legacy tokenizer.

Also adds tests to ensure that the move does not break the hacks.

### DIFF
--- a/lib-php/Geocode.php
+++ b/lib-php/Geocode.php
@@ -506,13 +506,6 @@ class Geocode
             userError('Query string is not UTF-8 encoded.');
         }
 
-        // Conflicts between US state abreviations and various words for 'the' in different languages
-        if (isset($this->aLangPrefOrder['name:en'])) {
-            $sQuery = preg_replace('/(^|,)\s*il\s*(,|$)/i', '\1illinois\2', $sQuery);
-            $sQuery = preg_replace('/(^|,)\s*al\s*(,|$)/i', '\1alabama\2', $sQuery);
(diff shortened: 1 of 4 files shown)


freyfogle:
Just FYI, here is a full list of US postal state codes that correspond to country ISO codes. 
this was [the OpenCage monthly geotrivia question in Jan 2021](https://blog.opencagedata.com/post/geotrivia)

```
    AE - United Arab Emirates and American Forces Europe
    AL - Albania and Alabama
    AR - Argentina and Arkansas
    AZ - Azerbaijan and Arizona
    CA - Canada and California
    CO - Colombia and Colorado
    DE - Germany and Delaware
    GA - Gabon and Georgia
    ID - Indonesia and Idaho
    IL - Israel and Illinois
    IN - India and Indiana
    KY - Cayman Islands and Kentucky
    MA - Morocco and Massachusetts
    MD - Moldova and Maryland
    ME - Montenegro and Maine
    MN - Mongolia and Minnesota
    MO - Macau and Missouri
    MS - Monserrat and Mississippi
    MT - Malta and Montana
    NC - New Caledonia and North Carolina
    NE - Niger and Nebraska
    PA - Panama and Pennsylvania
    SC - The Seychelles and South Carolina
    SD - Sudan and South Dakota
    TN - Tunesia and Tennessee
    VA - The Vatican and Virginia
```

PR #2470, opened 2021-10-17
Updata Taiwan language usage

Supaplextw:
Including other National language in Taiwan: Taiwanese Hokkein, Taiwanese Hakka

### DIFF
--- a/settings/country_settings.yaml
+++ b/settings/country_settings.yaml
@@ -1141,7 +1141,7 @@ tv:
 # Taiwan (臺灣)
 tw:
     partition: 25
-    languages: zh-hant
+    languages: zh-hant, zh, nan, hak
 
 # Tanzania (Tanzania)
 tz:


lonvia:
Thank you for the PR. Just to clarify, the languages field in this file should reflect which languages are used for the `name` tag in a country. That is not necessarily the same as the list of national languages. It depends on the convention for name tagging, the community. https://wiki.openstreetmap.org/wiki/Taiwan/Taiwan_tagging#.E5.BB.BA.E8.AD.B0.E7.94.A8.E6.B3.95 seems to suggest that the Chinese is always used in `name` but I might look at a bad translation.

So my question is this: are Taiwanese Hokkein and Taiwanese Hakka used in the `name` tags in Taiwan or are they used exclusively with `name:nan` and `name:hak`?

lonvia:
Closing for lack of response.

Issue #2365, opened 2021-06-15
Multi-region scripts don't work anymore since php scripts were removed


ionuttocila:
**Nomination/utils/setup.php not found when running utils/import_multiple_regions.sh**

**How to reproduce**
- follow the install instructions from official documentation, branch master for Nomination
- after cloning and build there is no setup.php file located in utils folder

i order to work i changed the import command to be 
`${SETUPFILE} import --osm-file ${UPDATEDIR}/tmp/combined.osm.pbf  2>&1`
where the SETUPTILE is the "nominatim" file from build directory

Also, when running this command  make sure to run it with the user which has access to Postgresql, like

` sudo -u nominatim bash utils/import_multiple_regions.sh`

The documentation should be updated or the code from utils directory to use the nomination file from build directory.

lonvia:
The multi region scripts currently don't work on master. I haven't made up my mind yet what is going to happen with them. We might add official support in the nominatim tool for multiple regions or not.

PRs are welcome to update the multi region scripts to use the nominatim cli tool. Otherwise, please use official releases if you need those scripts.

lonvia:
Reopening. Imports now work with `nominatim setup --osm-file <file1> --osm-file <file2>`. Updates are still on the todo list.

PR #2473, opened 2021-10-19
docs: add dotenv dependency to be installed

williancolognesitrimble:
It must be installed to be able to use utils scripts


williancolognesitrimble:
> You should

humm I didn't know that, does the nominatim cli has the `import_multiple_regions.sh` capability ?

lonvia:
The scripts for multiple regions are indeed in a bit of a limbo at the moment, see #2365. For imports use `nominatim import --osm-file <file1> --osm-file <file2>` on master. The last release indeed needs the util scripts for that. The hint would have to go to https://nominatim.org/release-docs/latest/admin/Advanced-Installations/ on the latest release docs. I'll have a look. Closing here.

Issue #307, opened 2015-08-12
debug script which shows all OS,Postgres,Nominatim parameters


mtmail:
[enhancement]
It's bothersome asking users for their Nominatim version, paths, PostGIS version. It would be cool to have a script, e.g. `utils/collect_os_info.sh` which prints all information we usually need in one go and make it a requirement for bug reports to include the report. Other projects do that (I can't remember one on top of my head.)
It might print sensitive information, e.g. the postgres connection string might include a password. We need to either remove that or make the user aware.

lonvia:
Can I bump this issue a bit? It would be really helpful to have more consistent reports from people.

lonvia:
This is half-way done now in #2552. Help would still be welcome to integrate the code into the nominatim CLI tool. See PR for details.

lonvia:
Implemented in #2739.

PR #2552, opened 2021-12-13
add utils/collect_os_info.sh script

micahcochran:
This was requested in issue #307 .  It is a bash script that generates a report of some the information needed for a bug report.

**What works:**
* OS version
* RAM
* Number of CPUs
* PostgreSQL version

**What doesn't work:**
* Nominatim Version - this will work if the script is ran from the "utils" folder.  Python does a local import of the version.
* PostGIS Version - I have an idea of what to do, but I don't have a test machine at the moment.
* Type and size of disks:  - I would need some help with what is the relevant information.  I put some links in the script for ways to get information.  This could vary wildly based on the the setup (bare metal/Cloud/VM/Container).
*  - bare metal/AWS/other cloud service: - Could just query the user and fill out common answers.

### DIFF
--- a/utils/collect_os_info.sh
+++ b/utils/collect_os_info.sh
@@ -1,0 +1,94 @@
+#!/usr/bin/env bash
+
+Description="The purpose of this script is to collect system information for bug reports.\n
+Submit issues to https://github.com/osm-search/Nominatim/issues"
+
+
+####### Gather the Information ##################################################
+# Separate the information gathering from the report generation.  Dividing these
+# makes it easier to make trivial changes by not have to learn the other portion
+# of this script.
+
+# Nominatium version
+# NOTE: Getting this version will NOT work if it is being ran from in another
+# folder than Nominatim/utils.  It call python3 to import version.py locally and
+# prints it in the version format. 
+NominatimVersion=`cd ../nominatim/ && python3 -c "import version; print('{0[0]}.{0[1]}.{0[2]}-{0[3]}'.format(version.NOMINATIM_VERSION))"`
+
+# PostgreSQL version
+PostgreSQLVersion=`postgres --version`
+if [ "$?" -ne "0" ]
+then
+  PostgreSQLVersion="Not installed"
+fi
+
+# - PostGIS version:
+# The command for this should look something like this:
+#       psql -U nominatim -d mydatabase -c 'SELECT PostGIS_full_version();'
+# ASSUME the username is nominatim
+# This needs to be ran under the account with the appropriate permissions.
+# This has been left blank.
+PostGISVersion=
+
+# There are different ways to getting the Linux OS information.
+# https://www.cyberciti.biz/faq/how-to-check-os-version-in-linux-command-line/
+# /etc/os-release has a number of representations of the OS
+# PRETTY_NAME is pity.
+OperatingSystem=`grep '^PRETTY_NAME' /etc/os-release | cut -d'=' -f2`
+
+RAM=`grep ^MemTotal /proc/meminfo | cut -d':' -f2`
+
+# In /proc/cupinfo: siblings seems to refer to total cores like hyperthreaded cores.
+# The hyperthreaded cores could be included if that is needed.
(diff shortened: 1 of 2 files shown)


lonvia (commented):
Thank you. This looks quite useful already. And very well documented. I added some thoughts to specific points inline.

I do wonder about how to best integrate such a script, though. When the issue was originally opened, Nominatim wasn't installable yet, so everybody had the source code with the utils directory lying around. That is not the case anymore. So we might be better off to offer this functionality as part of the nominatim command. I could imagine a verbose version output `nominatim --version -vv` or something as part of the admin command `nominatim admin --system-info`. It means that the script needs to be rewritten in Python. The added bonus is that you have access to stuff like the [postgresql and postgis version helper functions](https://github.com/osm-search/Nominatim/blob/master/nominatim/db/connection.py#L115). What do you think?

utils/collect_os_info.sh
```diff
@@ -0,0 +1,94 @@
...
+  PostgreSQLVersion="Not installed"
+fi
+
+# - PostGIS version:
+# The command for this should look something like this:
+#       psql -U nominatim -d mydatabase -c 'SELECT PostGIS_full_version();'
+# ASSUME the username is nominatim
+# This needs to be ran under the account with the appropriate permissions.
+# This has been left blank.
+PostGISVersion=
```
otbutz, line 31: Assuming that the default postgis3 version is installed, maybe something like this:

```sh
grep '^default_version' "$(pg_config --sharedir)/extension/postgis-3.control"
```

utils/collect_os_info.sh
```diff
@@ -0,0 +1,94 @@
...
+echo - PostGIS version: $PostGISVersion 
+echo - OS: $OperatingSystem
+echo
+echo
+
+
+echo "**Hardware Configuration (please correct the following information):**"
+echo - RAM: $RAM
+echo - number of CPUs: $NumCPUs
+echo - type and size of disks:
```
otbutz, line 88: systemd-detect-virt

micahcochran: That works really well to report if there is a container or VM and what type environment.  Thanks!

utils/collect_os_info.sh
```diff
@@ -0,0 +1,94 @@
...
+####### Gather the Information ##################################################
+# Separate the information gathering from the report generation.  Dividing these
+# makes it easier to make trivial changes by not have to learn the other portion
+# of this script.
+
+# Nominatium version
+# NOTE: Getting this version will NOT work if it is being ran from in another
+# folder than Nominatim/utils.  It call python3 to import version.py locally and
+# prints it in the version format. 
+NominatimVersion=`cd ../nominatim/ && python3 -c "import version; print('{0[0]}.{0[1]}.{0[2]}-{0[3]}'.format(version.NOMINATIM_VERSION))"`
```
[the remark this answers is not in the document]
micahcochran: Yes, I will give that a try.

micahcochran:
Thank you again for the feedback.

In commit 8bda59f, I have rewritten the script as a separate Python file.  I have not integrated with the cli.py, but at least this being written in Python it would be a step closer to being able to be integrated with the file.

There are a few fields that still need to be filled in.

I have included the bash file and the Python file so that they can both be available for the review.  Not looking for this to get merged at this point.  Just looking for feedback.

lonvia (commented):
Works like a charm.

To integrate it with the nominatim tool, just move the script to nominatim/tools and add a new subcommand in nominatim/clicmd/admin.py. Please run pylint against it and make sure you use spaces instead of tabs.

If you don't have the means to set up a test machine, you can use the CI for testing. Just add the execution of your command around [here](https://github.com/osm-search/Nominatim/blob/master/.github/workflows/ci-tests.yml#L284), push the change and check the action output in Github.

lonvia:
What you've done so far looks really good and I'd be happy to see that merged. So I'm tagging this with `help wanted`, so that maybe somebody else can pick up the integration into nominatim CLI.

PR #2566, opened 2022-01-07
add Caddy example to Deployment.md

WebFreak001:
based on my active config (on https://nominatim.webfreak.org) but reduced to just the API

It may be useful to know how to enable logging or add a custom homepage like on my website, but I don't think it's really that relevant to Nominatim and would bloat the docs a little. Though I could include it if wanted.

My extra config if interested:
```
        handle / {
                # custom homepage (static html page)
                root * /srv/http/org.webfreak.nominatim/public
                file_server
        }

        log {
                output file /data/logs/http/org.webfreak.nominatim.log {
                        roll_size 50MiB
                }
                format console
        }
```

### DIFF
--- a/docs/admin/Deployment.md
+++ b/docs/admin/Deployment.md
@@ -140,3 +140,51 @@ location ~ [^/]\.php(/|$) {
 Restart the nginx and php-fpm services and the website should now be available
 at `http://localhost/`.
 
+## Nominatim with Caddy
+
+### Installing the required packages
+
+You need to install a PHP FastCGI server such as php-fpm to use with caddy.
+
+Follow caddy installation instructions: https://caddyserver.com/docs/install
+
+On Ubuntu/Debian install php-fpm with:
+
+``` sh
+sudo apt install php-fpm
+```
+
+### Configure php-fpm with Caddy
+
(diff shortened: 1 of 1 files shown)


lonvia:
I have never heard of caddy before this. So there is a bit of a fundamental question how much niche installation instruction we allow here. While I don't see much issue with an additional independent section, there still is the problem of bit-rot. There are only so many configurations we can try and keep up-to-date. @mtmail, any opinion?

otbutz:
I would love to see caddy support as it makes LetsEncrypt integration *a lot* easier. Currently i set it up as a simple reverse proxy. Getting rid of apache/nginx would be nice.

Caddy is also widely used:

image

lonvia:
That's a very nice idea.

@WebFreak001 will you open a discussion under [Show and tell](https://github.com/osm-search/Nominatim/discussions/categories/show-and-tell) and move your description there?

WebFreak001:
https://github.com/osm-search/Nominatim/discussions/2580

lonvia:
https://github.com/osm-search/Nominatim/commit/d6b5f2f5da8b1a5bca9e27af55d755ed90249587 adds a link in the deployment section.

Closing here. Thank you @WebFreak001.

Issue #2575, opened 2022-01-12
potential memory leak in replication


woodpeck:
**Describe the bug**
A long-running `nominatim replication` process consumes more and more memory.

**To Reproduce**
Set up Nominatim as per instructions. Run `nominatim replication` for three months (with minutely updates). Monitor memory usage. On my system, the memory usage as per "systemctl status" climbs from around 7 GB after a fresh start to around 40 GB after three months. The memory usage as per "ps auxw" climbs from 1.3 GB VSZ/115kB RSS after a fresh start to 11 GB VSZ/7 GB RSS after three months. (All values taken while the process was in "sleeping before next update" state.)

**Software Environment (please complete the following information):**
- Nominatim version: 3.7.2
- Postgresql version: 12
- Postgis version: 3.0
- OS: Ubuntu 20.04

**Hardware Configuration (please complete the following information):**
- RAM: 64 GB
- number of CPUs: 1 (Ryzen 3700x, 8 cores)
- type and size of disks: 2 TB NVMe
- bare metal/AWS/other cloud service: bare metal 

**More context**

Here's a munin graph of memory usage on that system:

![image](https://user-images.githubusercontent.com/705471/149175078-41c24cf7-59f9-4392-9856-131addd16343.png)

lonvia:
Turns out that pyosmium replicaton is leaking, see https://github.com/osmcode/pyosmium/issues/195. Leaving this issue open as there might be changes in Nominatim necessary.

As a workaround, don't use the continuous updates but run `nominatim replication --once` in a loop instead.

PR #2598, opened 2022-02-01
Document how to set up systemd timers for --once updates

amandasaurus:
This approach (of using a systemd timer & service, instead of a long running service) can address https://github.com/osm-search/Nominatim/issues/2575 . Here I document how to do that.

### DIFF
--- a/docs/admin/Update.md
+++ b/docs/admin/Update.md
@@ -131,6 +131,63 @@ done
 
 A cron job then creates the file `/srv/nominatim/schedule-maintenance` once per night.
 
+##### One-time mode with systemd
+
+You can run the one-time mode with a systemd timer & service.
+
+Create a timer description like `/etc/systemd/system/nominatim-updates.timer`:
+
+```
+[Unit]
+Description=Timer to start updates of Nominatim
+
+[Timer]
+OnActiveSec=2
+OnUnitActiveSec=1min
+Unit=nominatim-updates.service
+
+[Install]
+WantedBy=multi-user.target
+```
+
+And then a similar service definition: `/etc/systemd/system/nominatim-updates.service`:
+
+```
+[Unit]
+Description=Single updates of Nominatim
+
+[Service]
+WorkingDirectory=/srv/nominatim
+ExecStart=nominatim replication --once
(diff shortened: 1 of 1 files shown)


lonvia:
Just some thought as I have been playing a lot with systemd lately:

My first attempt with once mode and systemd looked like this, too. But on second thought, the timer isn't really necessary. Just adding `Restart=on-success` (or `always` if you like to live dangerously) and `RestartSec=1min` should do pretty much the same and is easier to handle. The `--once` command will make sure that the process sleeps the appropriate time before asking for new updates.

Eventually I settled on a scheme that does use a timer but one with with a longer sleep time. The service unit then uses catch-up mode. That version is mainly useful when you don't need updates immediately but only want to update occasionally (once an hour or once a day).

amandasaurus:
> But on second thought, the timer isn't really necessary. Just adding Restart=on-success (or always if you like to live dangerously) and RestartSec=1min should do pretty much the same and is easier to handle.

Ah you maybe know more about systemd than me.

I like the separate timer & service. because I can stop the timer, and just wait for the service to finish. It's a clean way to stop updates. Is that possible with your `Restart..` approach?

lonvia:
> Ah you maybe know more about systemd than me.

Oh dear, don't be fooled, it's all solid "Halbwissen". ;)

> I like the separate timer & service. because I can stop the timer, and just wait for the service to finish. It's a clean way to stop updates. Is that possible with your Restart.. approach?

No, that wouldn't work. But then again, I don't bother with clean stops anyways. Waiting for indexing to finish on a larger batch of data can take a while. And updates will properly pick up again, no matter when they are stopped. There is a small issue that osm2pgsql won't get killed together with the python process, so you might have a zombie around that needs additional killing (or waiting to be finished). But I think that's rather behaviour we should fix eventually. Killing the osm2pgsql process half-way through is safe, too.

I wonder if the documentation should say something about stopping. It's not exactly obvious. Maybe you are in the mood to add a sentence. In any case, I'll merge this as soon as the CentOS CI has decided to cooperate. And then give it some thought if this shouldn't even become the recommended production setup and the forever mode be deprecated.

PR #2629, opened 2022-03-13
Move default country names into yaml configuration

### LINKED ISSUE #2616: Move default country names into yaml configuration
The initial country names are currently imported through a [SQL file](https://github.com/osm-search/Nominatim/blob/master/data/country_name.sql). This makes it hard for people to adopt it to their own needs. We should move the names to the [country yaml config file](https://github.com/osm-search/Nominatim/blob/master/settings/country_settings.yaml) and remove the SQL stuff.

This needs changing the import in https://github.com/osm-search/Nominatim/blob/master/nominatim/tools/country_info.py and the data needs to be moved, of course.

tareqpi:
- Moved the countries' names data that was in `country_name.sql` into `country_settings.yaml`
- Changed the import logic in `country_info.py` so it directly creates `country_name` table and loads the countries' names from `country_settings.yaml`
- Modified `CMakeLists.txt` so that `country_name.sql` is not required anymore
- Removed `country_name.sql`

This pull request resolves #2616.


lonvia:
A mixture of formats is not a good idea. The names should be in yaml format as well. The colons are going to be a bit tricky for that. Probably best to use a substructure for that, something along the lines of:

```
ad:
    partitions: 35
    languages: ca
    names:
        name:
            default: Andorra
            ab: Андора
            af: Andorra
            ....
```

tareqpi:
I wanted to know your opinion about the idea of modifying `country_info.py` so that we can append `name:` to each key just before copying the data into the `country_name` table so that the format of the names key-value pair is stored as `"name:en"=>"Monaco"` instead of `"en"=>"Monaco"`. This way, there won't be any major changes for the instances where `name:en` is used throughout the codebase. As a result, `country_settings.yaml` will adopt the format that you suggested and at the same time, the codebase won't have major changes.

lonvia:
The database must indeed store the keys in the format `name:en` etc. That must not change or the output will go wrong. What you can't see from the current SQL file is that it is possible to have other prefixes than `name`. You might have noticed the [two hard-coded short_name examples](https://github.com/osm-search/Nominatim/blob/2723553593f2c52ebf981d958942cc473f3db7e3/nominatim/tools/country_info.py#L109). This is something that should eventually move from being hard-coded to being configurable as well. So the format needs to keep an open mind about keys other than name.

nominatim/tools/country_info.py
```diff
@@ -74,8 +79,21 @@ def setup_country_tables(dsn, sql_dir, ignore_partitions=False):
             lang = props['languages'][0] if len(props['languages']) == 1 else None
             params.append((ccode, partition, lang))
 
+            name = add_prefix_to_keys(props.get('names').get('name'), 'name:')
```
lonvia, line 82: Could you make this generic for other prefixes than 'name'. Basically any keys under 'names' should be handled as a prefix.

nominatim/tools/country_info.py
```diff
@@ -74,8 +79,21 @@ def setup_country_tables(dsn, sql_dir, ignore_partitions=False):
...
+            cur.execute(
+                """ CREATE TABLE public.country_name (
+                        country_code character varying(2),
+                        name public.hstore,
+                        derived_name public.hstore,
+                        country_default_language_code text,
+                        partition integer
+                    ); """)
+            data = StringIO(country_names_data)
+            cur.copy_from(data, 'country_name', columns=('country_code', 'name'))
```
lonvia, line 96: Bonus points, if you merge the query below in here. The additional update was only necessary because the original content came from an external file. Now that you construct the table here from scratch, all the columns can be filled in one go.

Also: the copy is okay in principle. However, given that the table is relatively small, a mass-INSERT would just be fine as well. The advantage of the insert is that you don't have to correctly format and escape the fields yourself, i.e. the whole json.dumps is not necessary.

lonvia (changes requested):
The PR is missing some unit tests. They should go into `test/python/tools/test_country_info.py`. We need in particular tests for some corner cases: 'names' is missing in the config file, 'names' is there but empty, etc. Also: please add a test for a name that contains a backslash (\). I'm not convinced that json.dumps does the right thing for characters that need escaping in copy.

tareqpi:
Since you raised your concern in storing the escape characters correctly, I found out that the issue is not with `json.dumps()` but rather with how we are storing these characters in `country_settings.yaml`. These characters must be surrounded with double quotes so that they behave in the intended way.

`Reference:` https://yaml.org/spec/history/2001-08-01.html#:~:text=In%20YAML%2C%20text%20scalars%20can,%5C%5C%20to%20represent%20the%20backslash.

tareqpi:
As for writing the unit tests that are influenced by the structure of `country_settings.yaml`, I am not sure how to approach this since none of the functions in `country_info.yaml` use any of the parameters that are necessary to create the scenarios that you described above. I would appreciate your opinion on how to approach this.

lonvia:
Comfiguration files loaded through [`load_sub_configuration`](https://github.com/osm-search/Nominatim/blob/d33c82cb66a5d7edf7202e7e4ede8e2496011580/nominatim/config.py#L162) are first searched in the project directory and only then in the global config directory. This gives us a simple mechanism for custom configuration. Simply put a modified config file into your project directory. You can use this mechanism for the tests. [Here](https://github.com/osm-search/Nominatim/blob/d33c82cb66a5d7edf7202e7e4ede8e2496011580/test/python/tokenizer/test_icu.py#L84) is an example where ICU tokenizer tests overwrite their default config file. The `project_env` fixture helps you, setting up a project directory for the test. So something as simple as this should work:

```
def test_something(project_env):
    (project_env.project_dir / 'country_info.yaml').write_text('something')
```

test/python/tools/test_country_info.py
```diff
@@ -59,3 +59,39 @@ def test_create_country_names(temp_db_with_extensions, temp_db_conn, temp_db_cur
...
+de:
+    partition: 3
+    languages: de
+"""
+))
+def test_load(project_env, def_config, yaml_file_content):
+    (project_env.project_dir / 'country_settings.yaml').write_text(yaml_file_content)
+    
+    country_info._COUNTRY_INFO._info = def_config.load_sub_configuration(
+        (project_env.project_dir / 'country_settings.yaml'))
```
lonvia, line 87: You should call `setup_country_config(def_config)` here instead of loading it yourself.

test/python/tools/test_country_info.py
```diff
@@ -59,3 +59,39 @@ def test_create_country_names(temp_db_with_extensions, temp_db_conn, temp_db_cur
...
+    
+    for prop in country_info._COUNTRY_INFO._info.values():
+        if 'languages' not in prop:
```
lonvia, line 90: You make assumptions here that the import has worked right. That is not allowed. I.e. you cannot use the result of the function under test to decide which asserts to run. It is better to write 3 separate tests here.

test/python/tools/test_country_info.py
```diff
@@ -59,3 +59,39 @@ def test_create_country_names(temp_db_with_extensions, temp_db_conn, temp_db_cur
...
+
+def test_setup_country_config_languages_not_loaded(project_env, def_config):
+    (project_env.project_dir / 'country_settings.yaml').write_text("""
+de:
+    partition: 3
+    names: 
+        name: 
+            default: Deutschland
+""")
+    country_info.setup_country_config(def_config)
```
lonvia, line 72: Ah, sorry. `project_env` is already the configuration (see `test/python/conftest/`), so you would want to use `country_info.setup_country_config(project_env)`

nominatim/tools/country_info.py
```diff
@@ -72,15 +83,24 @@ def setup_country_tables(dsn, sql_dir, ignore_partitions=False):
             else:
                 partition = props.get('partition')
             lang = props['languages'][0] if len(props['languages']) == 1 else None
-            params.append((ccode, partition, lang))
-
+            name = add_prefix_to_keys(props.get('names')
+                 .get(_COUNTRY_INFO.key_prefix()), _COUNTRY_INFO.key_prefix())
+            name = json.dumps(name, ensure_ascii=False, separators=(', ', '=>'))
+            params.append((ccode, name[1:-1], lang, partition))
```
lonvia, line 89: You shouldn't need the conversion to json for name anymore here. psycopg should do the conversion for you. If it doesn't, then you need the [hstore adapter](https://www.psycopg.org/docs/extras.html?highlight=hstore#psycopg2.extras.register_hstore). Add a line `psycopg2.extras.register_hstore(cur)` below before the `execute()`.

lonvia (approved):

PR #2641, opened 2022-03-20
Transparantly reinitialize tokenizer directory when necessary

lonvia:
Recreate `tokenizer/tokenizer.php` and `tokenizer/nominatim.so` when they are not available in the project directory. This means that you can now again easily move to a different server or project directory. Simply rerun `nominatim refresh --website`. It's not possible to recreate any other configuration items or file, so there is still lots of opportunity to shoot yourself in the foot. So you should know what you are doing.

@leonardehrenfried I can't find the recent discussion anymore, but this should solved the problems with docker image updates. Spin up the image with the new version, point it to the old database, run `nominatim refresh --website` then `nominatim admin --migrate` and all should be well. Maybe you can confirm at some point.

### DIFF
--- a/docs/admin/Advanced-Installations.md
+++ b/docs/admin/Advanced-Installations.md
@@ -198,11 +198,10 @@ target machine.
     of a full database.
 
 Next install Nominatim on the target machine by following the standard installation
-instructions. Again make sure to use the same version as the source machine.
+instructions. Again, make sure to use the same version as the source machine.
 
-You can now copy the project directory from the source machine to the new machine.
-If necessary, edit the `.env` file to point it to the restored database.
-Finally run
+Create a project directory on your destination machine and set up the `.env`
+file to match the configuration on the source machine. Finally run
 
     nominatim refresh --website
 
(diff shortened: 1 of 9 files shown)


leonardehrenfried:
Thanks for this @lonvia!

The discussion happened in #osm-dev but I've made a screenshot and attached it to the PR back in the day: https://github.com/mediagis/nominatim-docker/pull/274

The hack referenced in the PR is the following: https://github.com/mediagis/nominatim-docker/pull/274/commits/64904faa4dc6bcd94182e56d661c3b42cec25550

Thanks to this feature we can now remove the hack - that's great! I'm knee-deep in some other project so realistically I won't be able to confirm that it works until there is a 4.1 release. I'm sorry about that but I don't want to promise something I cannot keep.

Question: What does `nominatim admin --migrate` do?

lonvia:
Okay, if it doesn't work against all expectation, then you still have a working hack.

`nominiatim admin --migrate` moves an existing database to a new Nominatim version, see https://nominatim.org/release-docs/latest/admin/Migration/. It is safe to simply call every time you spin up a docker container against an existing database.

leonardehrenfried:
Today we got around to upgrading our container image to nominatim 4.1. I'm pleased to inform you that we were able to remove the tokenizer copying hack.

Thanks for adding it!

PR #2654, opened 2022-03-30
Implement version subcommand

### LINKED ISSUE #2615: Add `--version` function to nominatim CLI tool
The nominatim CLI tool has no --version parameter, which is really something every tool should have. `nominatim/version.py` has the version of the software. A verbose version of the command (`--version -v`) should also print the version of the database if it can connect to a database.

TODO:

* [x] implement `--version` command
* [ ] output database version when run with `--version -v`)
* [x] output commit, when being built from a git clone

CicadaCinema:
Closes #2615

I initially tried to implement this with the single line `self.parser.add_argument('-V', '--version', action='version', version=f'{prog} {".".join([str(i) for i in NOMINATIM_VERSION])}')`, but this didn't allow me the flexibility to implement database checking functionality. I also avoided creating a new file for my class because `version.py` already seemed like a sensible choice.

I haven't actually tested the database checking functionality because I'm getting some weird behaviour with postgres on my machine - I'm going to detail that below since I think this is just a configuration error on my end, as opposed to an issue-worthy problem.


lonvia:
I have to disagree with both decisions, I'm afraid. The command has to be `nominatim --version` because that is the agreed-upon standard for command line tools. Anything that has to do with command line parsing has to go into `cli.py` or the `clicmd` module. The version code should be small enough, that you can simply add a function to the CommandlineParser class.

PR #2671, opened 2022-04-15
Warn about currently used locale

mmaridev:


mmaridev:
I was here once again since always start from a Proxmox CT image.

Maybe it's time to merge this into the wiki?

/cc @lonvia

lonvia:
I'm not a fan of having commands in these scripts that are purely there for printing something out for the user. The scripts main purpose is to be executed. 

Should we just add `dkpg-reconfigure locale` to the instructions, so it simply works? Could it do damage when locales are already set up? @mtmail what do you think?

mtmail:
I'd add a Troubleshooting FAQ entry for that error message https://nominatim.org/release-docs/latest/admin/Faq/ instead.

lonvia:
Good idea.

lonvia:
Replaced by #3491

mmaridev:
The goal of my PR was to have it published on the
https://nominatim.org/release-docs/develop/appendix/Install-on-Ubuntu-22/
page...

Il Lun 29 Lug 2024, 18:35 Sarah Hoffmann ***@***.***> ha
scritto:

PR #2672, opened 2022-04-15
#2615 Version command

xvalme:
Related to #2615. It is a version subcommand that checks the Nominatim version(pretty obvious, right?). Moreover, it also checks the repository´s version.py file to look for updates. This can be disabled with --offline. 
Note :It is in the 1st position of the available commands, since this is one of the most important ones when checking a new installation.


lonvia:
I really have to insist that it works like `nominatim --version` because that is the agreed-upon standard for command line tools, i.e please don't implement it as a subcommand..

xvalme:
> I really have to insist that it works like `nominatim --version` because that is the agreed-upon standard for command line tools, i.e please don't implement it as a subcommand..
Hum, ok. Can you point me out on what to do? Not finding where the other tools are defined to give it a look. From what I understand, I should add it by:
`group.add_argument('-vv', '--version', dest=clicmd.Version(), help='Get current and last versions.')`
Is that correct? And just have to then pass the arguments from the function call?

lonvia:
Yes, `add_argument` is the right way. Please don't add any short parameter version. Just `--version` will do.

I can't comment on the unrecognized argument error for obvious reasons (i.e. no code).

PR #2686, opened 2022-04-29
Install-on-Ubuntu-20.sh - correct php version

mtmail:
One line showed a different version from all others.


lonvia:
Looks like we should be testing the nginx version in the CI, too.

PR #2695, opened 2022-05-04
add git commit hash to 'nominatim --version' output

mtmail:
for https://github.com/osm-search/Nominatim/issues/2615

```bash
$ nominatim --version
Nominaim version 4.0.99.6 (8c073993)
```

Cmake run fails should git not be installed. A work-around could be to execute multiple commands in `execute_process`, the last being 'true' and setting `COMMAND_ERROR_IS_FATAL LAST`.

During development I got an error
```
fatal: unsafe repository ('/home/vagrant/Nominatim' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /home/vagrant/Nominatim
```
That's related to the recent https://github.blog/2022-04-12-git-security-vulnerability-announced/ Probably only affects my virtual machine setup and other users won't see this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,17 @@ set(NOMINATIM_VERSION "${NOMINATIM_VERSION_MAJOR}.${NOMINATIM_VERSION_MINOR}.${N
 
 add_definitions(-DNOMINATIM_VERSION="${NOMINATIM_VERSION}")
 
+# Setting GIT_HASH
+find_package(Git)
+if (GIT_FOUND)
+    execute_process(
+        COMMAND "${GIT_EXECUTABLE}" log -1 --format=%h
+        WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
+        OUTPUT_VARIABLE GIT_HASH
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+        ERROR_QUIET
+        )
+endif()
 
 #-----------------------------------------------------------------------------
 #  Configuration
--- a/cmake/tool-installed.tmpl
+++ b/cmake/tool-installed.tmpl
@@ -7,6 +7,9 @@ sys.path.insert(1, '@NOMINATIM_LIBDIR@/lib-python')
 os.environ['NOMINATIM_NOMINATIM_TOOL'] = os.path.abspath(__file__)
 
 from nominatim import cli
+from nominatim import version
+
+version.GIT_COMMIT_HASH = '@GIT_HASH@'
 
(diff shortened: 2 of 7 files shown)


CMakeLists.txt
```diff
@@ -26,6 +26,12 @@ set(NOMINATIM_VERSION "${NOMINATIM_VERSION_MAJOR}.${NOMINATIM_VERSION_MINOR}.${N
 
 add_definitions(-DNOMINATIM_VERSION="${NOMINATIM_VERSION}")
 
+execute_process(
+    COMMAND git log -1 --format=%h
+    WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
+    OUTPUT_VARIABLE GIT_HASH
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
```
lonvia, line 34: Check for the existence of git first, see [how osm2pgsql does it](https://github.com/openstreetmap/osm2pgsql/blob/43f027532afb21a0ac177ea6896931849d9f56ea/CMakeLists.txt#L78).

cmake/tool-installed.tmpl
```diff
@@ -7,6 +7,9 @@ sys.path.insert(1, '@NOMINATIM_LIBDIR@/lib-python')
 os.environ['NOMINATIM_NOMINATIM_TOOL'] = os.path.abspath(__file__)
 
 from nominatim import cli
+from nominatim import version
+
+version.GIT_COMMIT_HASH = '@GIT_HASH@'
```
lonvia, line 12: The same change is needed in `cmake/tool.tmpl`.

nominatim/cli.py
```diff
@@ -61,7 +61,10 @@ def __init__(self, prog, description):
     def nominatim_version_text():
         """ Program name and version number as string
         """
-        return "Nominatim version %s.%s.%s.%s\n" % version.NOMINATIM_VERSION
+        text = 'Nominaim version %s.%s.%s.%s' % version.NOMINATIM_VERSION
```
lonvia, line 64: Good thing we have tests. They caught that typo in Nominatim.

mtmail: Funny. I found some more even, will submit in a separate PR.

lonvia:
If I can do one more request: could you add a call to `nominatim --version` in the CI [around here](https://github.com/osm-search/Nominatim/blob/720c7b751906a41d2cdcfae7c461d74907771037/.github/workflows/ci-tests.yml#L259).

lonvia:
Yes, the usual behviour would be to return 0.

mtmail:
It's 0 now, tests pass.

Issue #2658, opened 2022-04-06
Better filtering for irrelevant address parts for POIs


lonvia:
#2082 has implemented filtering of address parts, where the POI is outside the area of the address part. This solves the issue for bad addresses when the parent street goes through multiple administrative area in most cases. For performance reasons we do this only, when multiple areas of the same address level are part of the address list of a street. This leaves some corner cases, where there is only one area of a certain address level that is partially touched by the street, see discussion in #2649.

Instead of guessing, which areas need rechecking, it would also be possible to explicitly mark areas that need rechecking in the `place_addressline` table. This would catch that case and make the rechecking a bit easier in general.

champagne-cmd:
@lonvia I would be interested in working on this issue. Do you think someone new to the Nominatim project could handle this issue?

lonvia:
It's challenging but doable if you are not afraid to dive head-deep into pl/PgSQL code.

Here are some pointers to get you started:

* The basic idea would be to add a boolean column to [place_addressline](https://github.com/osm-search/Nominatim/blob/a319b0a0b42c1acdc9b6a6e16fe54dd37128aec2/lib-sql/tables.sql#L130) which is set to true when the area of the parent place does not cover a street completely. A good approximation for being a street object is probably: it is of type LineString.
* place_addressline is filled during indexing of places. The function [insert_addressline](https://github.com/osm-search/Nominatim/blob/a319b0a0b42c1acdc9b6a6e16fe54dd37128aec2/lib-sql/functions/placex_triggers.sql#L447) is the relevant code. The tricky part here is that the parent objects are searched in the `location_area_large_*` tables. These tables contain chopped up geometries, so you can't use them. There is already a hack to load full geometries in the function [here](https://github.com/osm-search/Nominatim/blob/a319b0a0b42c1acdc9b6a6e16fe54dd37128aec2/lib-sql/functions/placex_triggers.sql#L559). That's the way to go.
* The recheck if a POI point is really inside the area is hidden in [this insane order statement](https://github.com/osm-search/Nominatim/blob/a319b0a0b42c1acdc9b6a6e16fe54dd37128aec2/lib-sql/functions/address_lookup.sql#L244).
* Tests for the dynamic address computation are in test/bdd starting [here](https://github.com/osm-search/Nominatim/blob/a319b0a0b42c1acdc9b6a6e16fe54dd37128aec2/test/bdd/db/import/addressing.feature#L405).

[...]

PR #2699, opened 2022-05-06
Add testcase dynamic address filtering 

champagne-cmd:
This is a testcase for the corner case described in #2658, where there is only one area of a certain address level that is partially touched by a street.

### DIFF
--- a/test/bdd/db/import/addressing.feature
+++ b/test/bdd/db/import/addressing.feature
@@ -402,6 +402,7 @@ Feature: Address computation
            | osm | display_name            |
            | N1  | Bolder, Wonderway, Left |
 
+    @testing
     Scenario: POIs can correct address parts on the fly
         Given the grid
             | 1 |   |   |   |  2 |   | 5 |
@@ -434,7 +435,7 @@ Feature: Address computation
            | osm | display_name               |
            | N2  | Leftside, Wonderway, Right |
 
-
+    @testing
     Scenario: POIs can correct address parts on the fly (with partial unmatching address)
         Given the grid
             | 1 |   |   |   |  2 |   | 5 |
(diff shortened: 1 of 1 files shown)


champagne-cmd:
This test case, as it currently stands, fails. But, it does not fail for the reason I want. I am trying to get it to fail by misplacing the neighborhood in the address returned (`display_name`). Instead, it will not include the neighborhood at all in the `display_name`. I am trying to replicate the output given in #2649. Is the discrepancy in the `osm_type`? I have tried changing that in the test, but whenever I do, the test throws an error complaining that there is no such thing as `osm_type`.

lonvia:
> This test case, as it currently stands, fails. But, it does not fail for the reason I want. I am trying to get it to fail by misplacing the neighborhood in the address returned (display_name). Instead, it will not include the neighborhood at all in the display_name.

Is that with the current state of the test? There is a 'Central' in the `display_name` that does not show up anywhere else (i.e. there is no boundary with that name), so that clearly would make the test fail.

You are on the right track with the test. Try leaving out R3 ('Super'). When you run the test, the display name still should have 'Super' in it even though the 'Bolder' node is clearly outside R3.

> Is the discrepancy in the osm_type? I have tried changing that in the test, but whenever I do, the test throws an error complaining that there is no such thing as osm_type.

`osm_type` and `osm_id` are together encoded in the `osm` column for the input data and the `object` column in the output data. You can't set osm_type separately because it is used as the unique identifier in the tests. You also have to be careful with changing the object type because there is a natural relation between the OSM object type and the kind of geometry you can use: OSM nodes only can be points, OSM ways can be ways and polygons, OSM relations are almost always polygons.

lonvia:
Closing this here as the whole approach from #2649 turned out to be infeasible.

Issue #2622, opened 2022-03-02
Update Date


maxsnts:
Hello

Updating a multiregion install using the script update_database.sh does not update the "lastimportdate" field of "import_status".

While a minor issue, that field us used in "Nominatim UI" as the "Last Update Date".

Thank you
Best regards

lonvia:
I'll leave that to a user contribution to enhance.

d-e-v-esh:
I would like to work on this issue 👍

d-e-v-esh:
@lonvia Will adding this line in the `update_database.sh`  solve the problem?
```sh
psql -U postgres -d nominatim -c "UPDATE import_status SET lastimportdate = NOW()"

```

maxsnts:
That is what i did on my instance, but i don't know if that will help on single region updates that are not done with this file.
(I also don't know if this affects single region updates as i only use multiple)

lonvia:
No. The date needed is the date of the data imported. That is not necessarily NOW(). You can get it from the diff files.

maxsnts:
I see. 
I did it only so that my users would not complaint about an out of date data. :)

Just out out of curiosity, do you know if the diff XML files guaranteed to be ordered with the most recent updates at the end?

lonvia:
They are ordered by OSM type/ID. You'd need to scan the whole file and find the maximum date.

sudeersharma:
To solve this issue, you will need to modify the update_database.sh script to update the lastimportdate field in the import_status table.

To do this, you can add a line of SQL code at the end of the script to update the lastimportdate field with the current date and time. Here's an example of how you could modify the script:
#!/bin/bash
set -e

# Update the database
cd /path/to/nominatim
./utils/update.php --import-osmosis-all

# Update the lastimportdate field
echo "UPDATE import_status SET lastimportdate=NOW();" | sudo -u postgres psql nominatim
This modified script will update the lastimportdate field in the import_status table to the current date and time each time the script is run. This should ensure that the "Last Update Date" displayed in the Nominatim UI is always up-to-date.

Note that you will need to replace /path/to/nominatim with the actual path to your Nominatim installation, and you may need to adjust the sudo -u postgres part of the SQL command depending on your PostgreSQL configuration.

mtmail:
@sudeersharma lonvia already explained that `NOW()` is not ideal and the date should be taken from the data files.

`./utils/update.php` is a Nominatim 3.x command, it no longer exists in Nominatim version 4.

[...]

PR #2700, opened 2022-05-06
Modify update_database.sh to update db date with python helper script

champagne-cmd:
This PR addresses #2622 and adds lines to the `update_database.sh` script to: (1) parse the timestamps from the `.gz` update files, (2) load the timestamps into a temporary file, (3) call a python helper script to determine the maximum timestamp, (4) update the timestamp in the database, and (5) remove the temporary timestamp file.

### DIFF
--- a/utils/parsedate.py
+++ b/utils/parsedate.py
@@ -1,0 +1,30 @@
+import sys
+import re
+from datetime import datetime
+
+if __name__ == "__main__":
+    if len(sys.argv) < 2:
+        raise TypeError("Must pass file to parse as an argument")
+
+    file_to_parse = sys.argv[1]
+    max_timestamp = None
+
+    with open(file_to_parse, "r") as f:
+        for line in f:
+            # split line with whitespace, '=', and '>' deliminators
+            split_line = re.split("[\s=>]", line)
+            # timestamp will be at index next to 'timestamp' string
+            if "timestamp" in split_line:
+                timestamp_idx = split_line.index("timestamp") + 1
+                timestamp_str = split_line[timestamp_idx]
+                # create timestamp object from string
+                timestamp = datetime.strptime(timestamp_str, "\"%Y-%m-%dT%H:%M:%SZ\"")
+                if max_timestamp is None:
+                    max_timestamp = timestamp
+                elif timestamp > max_timestamp:
+                    max_timestamp = timestamp
+
+        if max_timestamp is None:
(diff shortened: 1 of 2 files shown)


utils/update_database.sh
```diff
@@ -56,6 +58,12 @@ do
 
     echo "Attempting to import diffs"
     nominatim add-data --diff ${DIR}/${FILENAME}.osc.gz
+
+    echo "Attempting to update date of db"
+    zgrep timestamp ${DIR}/${FILENAME}.osc.gz | tee newtimestamps.txt
+    NEWTIME = $(python3 parsedate.py newtimestamps.txt)
```
lonvia, line 64: You could pipe the timestamps directly into your helper program. Then you don't need to worry about the temporary file.

Also, you don't catch the case here where there is no timestamp. I think that can happen, when there is no new data for a country (happens with small ones like Andorra).

utils/update_database.sh
```diff
@@ -22,7 +22,8 @@
 
 # REPLACE WITH LIST OF YOUR "COUNTRIES":
 #
-COUNTRIES="europe/monaco europe/andorra"
+#COUNTRIES="europe/monaco europe/andorra"
+COUNTRIES="europe/germany/sachsen"
```
lonvia, line 26: Development leftover?

utils/update_database.sh
```diff
@@ -39,6 +40,7 @@ FOLLOWUP="nominatim index"
 
 # ******************************************************************************
 UPDATEDIR="update"
+DATEUPDATE="state.txt"
```
lonvia, line 43: dto?

utils/parsedate.py
```diff
@@ -0,0 +1,30 @@
...
+    max_timestamp = None
+
+    with open(file_to_parse, "r") as f:
+        for line in f:
+            # split line with whitespace, '=', and '>' deliminators
+            split_line = re.split("[\s=>]", line)
+            # timestamp will be at index next to 'timestamp' string
+            if "timestamp" in split_line:
+                timestamp_idx = split_line.index("timestamp") + 1
+                timestamp_str = split_line[timestamp_idx]
```
lonvia, line 19: This will work with all current XML diff files, so probably fine.

More robust (and much shorter) is a simple regex-search: `line.search(r' timestamp="([^"]+)"').group(1)`

lonvia:
We'll generally need a different approach to multi-update. So closing here finally.

PR #2702, opened 2022-05-09
Clean up country name settings

lonvia:
Moves the country names into per-country include files to keep the files at a manageable size.

The parsing code is generalised, so that arbitrary name prefixes can be used in the name list. This allows to get rid of the last two hard-coded special cases for 'UK' and 'US'.


otbutz:
Could we split on a per continent basis?

lonvia:
I don't see the advantage in that.

PR #2706, opened 2022-05-10
PHP 8 behaves slightly different with in_array and usort

mtmail:
PHP 8.1 (default on Ubuntu-22) shows test suite failures we don't see with PHP 7.4

1. `php -r 'var_dump(in_array(0, array("option1","option2")));'`. PHP 7 returns `true`, PHP 8 returns `false`. PHP 8 is probably better here. Using strict mode https://www.php.net/manual/en/function.in-array.php both return `false`.

2. In `SimpleWordList.php` we use `usort` which can often have no winner. Thus nondeterministic sort outcome neither PHP 7 or PHP 8 can be said to be more correct than the other. I added a second clause to favor longer phrases. It's hopefully a relative cheap computation compared to hashing or full string comparison because the `usort` is called in a loop with potentially 100s of 1000s elements.

### DIFF
--- a/lib-php/ParameterParser.php
+++ b/lib-php/ParameterParser.php
@@ -70,7 +70,7 @@ class ParameterParser
             return $sDefault;
         }
 
-        if (!in_array($this->aParams[$sName], $aValues)) {
+        if (!in_array($this->aParams[$sName], $aValues, true)) {
             userError("Parameter '$sName' must be one of: ".join(', ', $aValues));
         }
 
--- a/lib-php/SimpleWordList.php
+++ b/lib-php/SimpleWordList.php
@@ -120,13 +120,18 @@ class SimpleWordList
         return array_slice($aWordSets, 0, SimpleWordList::MAX_WORDSETS);
     }
 
+    /**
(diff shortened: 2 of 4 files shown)


lonvia:
> It's hopefully a relative cheap computation compared to hashing or full string comparison because the usort is called in a loop with potentially 100s of 1000s elements.

Have you tried if sorting by string comparison is really an issue? Sorting 1000 elements shouldn't be much of a problem if they have implemented a half-way decent sorting algorithm. And it might make the code quite a bit easier.

mtmail:
String comparison is only done on the first element of the array. Shouldn't be an issue at.

mtmail:
This fixes the remaining test failure of the ubuntu-22 PR https://github.com/osm-search/Nominatim/pull/2691

lonvia:
I toke your comment that you wanted to switch to sorting by simple string comparison.

mtmail:
String comparison would require converting the array to a string first which has extra cost.

I read through https://www.php.net/manual/en/language.types.array.php and https://www.php.net/manual/en/ref.array.php again and the implementation in this PR seems the most performant already.

PR #2707, opened 2022-05-10
Make ICU tokenizer the default

lonvia:
The ICU tokenizer has been used in production for half a year on osm.org without major issues. Time to make it the default tokenizer. With the legacy tokenizer gone, there is also no need to automatically build the PostgreSQL module any longer. The code will remain and is still tested on the CI but the compilation needs to be explicitly activated. Existing installations should have the module around already, so there is very little need in the future to compile it.

@mtmail I'm aware that this conflicts rather strongly with the ongoing work in #2691. I'm happy to wait for it to be merged and resolve the conflicts here afterwards. But there are a couple of simplifications in the CI scripts that might make the other PR simpler as well. Among others, I have reduced the former 'icu-test', which now has become 'legacy-test' to be run on 20.04 only. Given that it is the past, it will be enough.

### DIFF
--- a/.github/actions/build-nominatim/action.yml
+++ b/.github/actions/build-nominatim/action.yml
@@ -5,6 +5,10 @@ inputs:
         description: 'Version of Ubuntu to install on'
         required: false
         default: '20'
+    cmake-args:
+        description: 'Additional options to hand to cmake'
+        required: false
+        default: ''
 
 runs:
     using: "composite"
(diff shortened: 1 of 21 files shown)


mtmail:
I'm happy to get Ubuntu-22 merge-ready again later, it's not pressing. Just something I know users eventually will ask for. We've been using ICU tokenizer fine for months, too, no issues.

Issue #125, opened 2014-04-05
undefined offset


SoulRaven:
i have found this error in my error log from nginx, 
"FastCGI sent in stderr: "PHP message: PHP Notice:  Undefined offset: 1 in /opt/Nominatim/lib/log.php on line 6""

The solution for me was:
"if (!isset($aStartTime[1])) $aStartTime[1] = '0';"

PR #2718, opened 2022-05-20
Undefined offset in error log

nslxndr:
Like #125, just in Apache error log and refers to end time:
`[Fri May 06 23:07:04.999970 2022] [php7:notice] [pid 1848445] [client X.X.X.X:54568] PHP Notice:  Undefined offset: 1 in /usr/local/lib/nominatim/lib-php/log.php on line 73`


lonvia:
Thank you!

Issue #2717, opened 2022-05-20
Icon URL contains duplicate path separator


otbutz:
e.g querying https://nominatim.osm.org/search.php?q=regensburg&format=jsonv2 yields:

`icon: https://nominatim.openstreetmap.org/ui/mapicons//poi_boundary_administrative.p.20.png`

Expected:

https://nominatim.openstreetmap.org/ui/mapicons/poi_boundary_administrative.p.20.png

lonvia:
Should be right now on all the OSMF servers.

PR #2724, opened 2022-05-23
Assume CONST_MapIcon_URL has trailing slash to avoid double slash

mtmail:
Fixes https://github.com/osm-search/Nominatim/issues/2717

I don't think a new unit test is needed is needed.

### DIFF
--- a/lib-php/ClassTypes.php
+++ b/lib-php/ClassTypes.php
@@ -270,7 +270,7 @@ function getIconFile($aPlace)
         return null;
     }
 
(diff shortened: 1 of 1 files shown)


lonvia:
We can't really do that, as it potentially breaks existing installations. The right thing to do is to remove the slash in the settings for the OSM servers, i.e. it's a system administration issue. I'll get around to do that eventually but it's really not a pressing matter.

mtmail:
Ok, so likely change to https://github.com/openstreetmap/chef/blob/master/cookbooks/nominatim/templates/default/nominatim.env.erb#L4 instead. Closing here.

PR #2739, opened 2022-06-07
integration of host system information script into Nominatim CLI tool

tareqpi:
- Refactored and included additional host system information into `collect_os_info.py`
- Integrated `collect_os_info.py` into Nominatim's CLI tool
- Added a test for the new feature in the CI flow

This pull request resolves #2552 and #307


nominatim/clicmd/admin.py
```diff
@@ -58,16 +61,23 @@ def run(args):
             LOG.warning('Analysing performance of indexing function')
             from ..tools import admin
             with connect(args.config.get_libpq_dsn()) as conn:
-                admin.analyse_indexing(conn, osm_id=args.osm_id, place_id=args.place_id)
+                admin.analyse_indexing(
+                    conn, osm_id=args.osm_id, place_id=args.place_id)
```
lonvia, line 65: Please do not reformat unrelated code.

nominatim/tools/collect_os_info.py
```diff
@@ -0,0 +1,161 @@
...
+
+        Optionally pass out or err parameters to redirect the output of stdout
+         and stderr to other file objects."""
+
+        # NOTE: This should be a report format.  Any conversions or lookup has be
+        #  done, do that action in the __init__() or another function.
+        message = """
+Use this information in your issue report at https://github.com/osm-search/Nominatim/issues
+Copy and paste or redirect the output of the file:
+    $ ./collect_os_info.py > report.md
```
lonvia, line 125: Needs adaption to the new command.

nominatim/tools/collect_os_info.py
```diff
@@ -0,0 +1,161 @@
...
+    def __init__(self, config):
+        self._memory: int = psutil.virtual_memory().total
+        self.friendly_memory: str = self._friendly_memory_string(self._memory)
+        # psutil.cpu_count(logical=False) returns the number of CPU cores.
+        # For number of logical cores (Hypthreaded), call psutil.cpu_count() or os.cpu_count()
+        self.num_cpus: int = psutil.cpu_count(logical=False)
+        self.os_info: str = self._os_name_info()
+
+        self.nominatim_ver: str = '{0[0]}.{0[1]}.{0[2]}-{0[3]}'.format(
+            NOMINATIM_VERSION)
```
lonvia, line 31: nominatim.version now has a convenience function `version_str()`.

lonvia (commented):
I've just added some minor comments. Looks okay in general. pylint has some more complaints, see [the CI](https://github.com/osm-search/Nominatim/runs/6770799994?check_suite_focus=true). You may need to install the latest version of pylint from pip to get the same reports locally.

The file `collect_so_info.sh` can be deleted, I think?

Could you please add a call to your new command in the CI run, around [here](https://github.com/osm-search/Nominatim/blob/800240550bf30a306ca7b23568bab24e3a894920/.github/workflows/ci-tests.yml#L244). Simply duplicate what is done for the version call. That way, we can see in the CI directly what the output looks like.

nominatim/tools/collect_os_info.py
tareqpi: I globally searched this function and couldn't find it.

lonvia: I mean this one here: https://github.com/osm-search/Nominatim/blob/800240550bf30a306ca7b23568bab24e3a894920/nominatim/version.py#L40

tareqpi: Yes, I have synced up the branch to the master branch because the branch was not updated for a long time and then found that function.

nominatim/tools/collect_os_info.py
```diff
@@ -0,0 +1,145 @@
...
+            if line.startswith(start):
+                result = line.split(sep)[fieldnum].strip()
+        return result
+
+
+def report_system_information(config):
+    """Generate a report about the host system including software versions, memory,
+    storage, and database configuration."""
+
+    with connect(config.get_libpq_dsn()) as conn:
```
lonvia, line 87: The CI test highlights a problem here: this will only work when a Nominatim database was already imported. To avoid that issue, try to connect against database 'postgres' instead. That will always exist.

In general, I would prefer that we catch any exception raised during connect and report something like 'Unable to connect to database' where the version string would be reported.

lonvia: To clarify, to connect to the postgres database you can simply use `connect("dbname=postgres")`

nominatim/tools/collect_os_info.py
```diff
@@ -95,14 +95,18 @@ def report_system_information(config):
...
-        with connect("dbname=postgres") as conn:
-            postgresql_ver: str = convert_version(conn.server_version_tuple())
-            postgis_ver: str = "Unable to connect to database"
+    with connect("dbname=postgres") as conn:
+        postgresql_ver: str = convert_version(conn.server_version_tuple())
+
+        with conn.cursor() as cur:
+            nominatim_db_exists = cur.execute("""
+            SELECT datname FROM pg_catalog.pg_database 
+            WHERE datname='nominatim'""")
```
lonvia, line 104: You cannot really make the assumption that the database is always called 'nominatim'. Users can use arbitrary names in their configuration.

That said, we cannot even make the assumption that they are using a local database. So I was wrong when recommending `connect("dbname=postgres")`. We need to use any other connection parameters from the user-supplied dsn. You can use [make_dsn()](https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.make_dsn) to replace the database name in the DSN that you get from `config.get_libpq_dsn()`:

```
connect(psycopg2.make_dsn(config.get_libpq_dsn(), dbname='postgres'))
```

You can also use [parse_dsn](https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.parse_dsn) to extract the database name from the DSN and then use the SELECT to find out if the database exists.

tareqpi:
Some of the third-party libraries' functions that are introduced in this pull request lack the correct type hints which is why `mypy` produces errors. I have raised an [issue](https://github.com/python/typeshed/issues/8502) to them and made the necessary pull requests to resolve it as well.

lonvia:
I've usually marked the calls with `# type: ignore[no-untyped-call]`, see for example https://github.com/osm-search/Nominatim/blob/eecc73ea1ab2005ca87b486e6bc2de41025a1ab4/nominatim/db/connection.py#L34

But if you contribute the necessary typing info to typeshed, even better.

lonvia:
With your typeshed PR merged (jay!), there was a new mypy error for a now unneded `ignore[no-untyped-call]`. I've already fixed that one master. Please rebase.

nominatim/tools/collect_os_info.py
```diff
@@ -0,0 +1,174 @@
...
+    ```
+    {postgresql_config}
+    ```
+    **Notes**
+    Please add any notes about anything above anything above that is incorrect.
+"""
+
+    try:
+        with open("report.md", "w", encoding='utf8') as output:
+            output.write(report)
```
lonvia, line 171: I would prefer if it just writes to stdout, i.e. uses print. Users can always redirect to a file if they want it in a file. If left like this we'd have to add a parameter to state the output filename and that makes things more complicated.

lonvia (commented):
Can you please rebase this once more on master. #2792 already had the adaptions for the type annotations. Also consider squashing your commits to get rid of the intermediate steps. If you do that, only squash your commits, not the ones from micahcochran.

PR #2746, opened 2022-06-16
Added additional languages for Pakistan in country settings

bgo-eiu:
Pakistan is a multi-lingual country and its official languages are lingua franca used in education + formal settings but which are not common native languages. Urdu for examples is spoken natively by about 8% of people, whereas Punjabi is spoken by about 40%.

When I am adding features in Pakistan on OSM, I usually use the `name:pnb` tag for place names in Punjab, since especially for rural areas that tends to be the original language of the place name. (See https://taginfo.openstreetmap.org/keys/name%3Apnb). I only just realized Nominatim does not have this in the list of languages for Pakistan, so this pull request adds the most commonly spoken languages besides English and Urdu: Punjabi (pnb, different from pa which is used in India because of a difference in script), Sindhi, Pashto, and Balochi.

I am in the process of compiling an abbreviation list for Pakistan to submit later, and hopefully eventually, a Normalization rule set usable with libICU.

### DIFF
--- a/settings/country_settings.yaml
+++ b/settings/country_settings.yaml
@@ -1247,7 +1247,7 @@ ph:
 # Pakistan (پاکستان)
 pk:
     partition: 14
-    languages: en, ur
+    languages: en, ur, pnb, sd, ps, bal
     names: !include country-names/pk.yaml
 
 


bgo-eiu:
Yes, the additional languages do meet that definition. There are many more languages in Pakistan but i stuck with the major languages for each province in this case. `name` is `pnb` Punjabi for many features in Punjab, `sd` Sindhi for many in Sindh, `ps` Pashto for many in Khyber Pakhtunkhwa, and `bal` Balochi for many in Balochistan. There is a mix of these languages, Urdu, and English used in `name` throughout the country. There isn't a hard and fast rule about which language ends up in `name` when there are multiple translations, so I just go with what's most common in signposts and writing, and I think others have been doing this similarly.

Many features on the map just have `name` at the moment with no additional language suffix tags to clarify them, but I have been starting to add those whenever I edit there.

lonvia:
Thank you for the background. Sounds all very sensible to me.

Issue #587, opened 2016-11-30
Add addr:unit support to nominatim


tpikonen:
(This is essentially a repost of bug 4886 in trac at https://trac.openstreetmap.org/ticket/4886 )

addr:unit is the recommended way of tagging staircases in European style apartment buildings. Support for addr:unit would be necessary in order to get a unique address to many buildings, i.e. there can be more than one building with the address 'Foo Street 5' and they are differentiated by the addr:unit on their entrance nodes. The address 'Foo Street 5 B' should then resolve to the (entrance) node tagged with addr:unit=B on a building tagged with addr:housenumber=5 and addr:street="Foo Street".

oiva:
At least in Finland the unit is most often a letter after the house number. One test case is [Kalastajanmäki 2](http://www.openstreetmap.org/way/48192753) which has four buildings with `addr:housenumber=2` and one of the buildings has two entrances, `addr:unit=A` and `addr:unit=B`.

Currently a search for "**Kalastajanmäki 2**" finds one of the buildings but a search for "**Kalastajanmäki 2 A**" doesn't find any of the buildings and instead only highlights the street Kalastajanmäki.

melvyn-sopacua:
In some regions a separator is used , like "2-A" or even "2 appt B". How is this stored in addr:unit if at all and is it consistent?

tpikonen:
i don't think a separator should be stored in addr:unit, since it is, as you said, dependent on region and the OSM data should be as universal as possible. The search engine (or renderer, or whatever is processing the human readable address) should infer the region from the rest of the address and use separator(s) appropriate for the region.

tvainika:
Also support for addr:flats would be nice.

E.g. searching for "Pietolankatu 59 A 1, Järvenpää, Finland", I would expect to find https://www.openstreetmap.org/node/2222227520 which is a node with addr:flats=1 in a way that has addr:unit=A (actually addr:unit is repeated in the node and the way)
and for "Pietolankatu 59 E, Järvenpää, Finland", I would expect to find just a way with addr:unit=E https://www.openstreetmap.org/way/197508914

Currently adding letters after number ruins the search at least in Finland, but how about other locations, languages, and address conventions?

james2432:
This is becoming more important now that the osm carto supports addr:units:
https://github.com/gravitystorm/openstreetmap-carto/compare/v4.3.0...v4.4.0

omgitsgela:
[...]

PR #2747, opened 2022-06-16
Added list of abbreviations for addresses in English in Pakistan

bgo-eiu:
You can take a look at some business directories like these to see how they are used in addresses:
https://www.dibpak.com/wp-content/uploads/2020/01/DubaiIslamicBank-Branch-Contact-Directory-Jan-2020_.pdf
https://psopk.com/files/station_finder/list_of_stations.pdf

I tried to make sure the terms in the YAML are all forms it makes sense for a geocoder to expand, some of the terms you see in the address listing above may look abbreviation-like but are not abbreviations or refer to places more specific than would be common in addresses. Let me know if you have any questions

### DIFF
--- a/settings/icu-rules/variants-en.yaml
+++ b/settings/icu-rules/variants-en.yaml
@@ -483,3 +483,50 @@
     -  Province -> Prov
     -  Senior High School -> SHS
     -  Subdivision -> Subd
+- lang: en
+  country: pk
+  words:
+    - Block -> B
+    - Block -> Blk
+    - Boulevard -> Blvd
+    - Cantonment -> Cantt
+    - Category -> Catogry
+    - Chak -> Chk
+    - Dera Ghazi -> DG
+    - District -> Distt
+    - District Council -> DC
+    - Federal Government Public School -> FGPS
+    - General Bus Stand -> GBS
+    - General Post Office -> GPO
+    - Government -> Govt
+    - Government Elementary School -> GES
+    - Government Primary School -> GPS
+    - Government Girls Primary School -> GGPS
+    - Government Girls High School -> GGHS
+    - Government High School -> GHS
+    - Gwadar Port Authority -> GPA
+    - House -> H
+    - House -> Ho
+    - House number -> HNO
+    - International -> Int
+    - Kilometer -> KM
+    - Kilometre -> KM
+    - Markaz -> Mkz
+    - Metropolitan Corporation - MC
+    - Mohammad -> Moh
+    - Muhammad -> Moh
+    - Mohammad -> Mohd
+    - Muhammad -> Mohd
+    - Municipal Corporation -> MC
+    - Number -> No
+    - Opposite -> Opp
+    - Plaza -> Plza
+    - Post Office -> PO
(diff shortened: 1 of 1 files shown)


settings/icu-rules/variants-en.yaml
```diff
@@ -483,3 +483,132 @@
...
+    - Shop -> Sh
+    - Showroom -> SR
+    - Station -> Stn
+    - Street -> St
+    - Suite -> Suit
+    - Survey Number -> SN
+    - Taluka -> Tehsil
+    - Taluka Municipal Administration -> TMA
+    - Tehsil -> Taluka
+    - Tehsil Municipal Administration -> TMA
```
lonvia, line 608: Taluka and Tehsil appear to be synonyms rather than abbreviation forms. Nominatim doesn't have a proper handling for that yet, but the abbreviation list is not the best place.

lonvia (commented):
Thank you for compiling the list. Just some observations from my side:

There are a couple of very generic terms in the list. They have limit use in the name search as currently implemented by Nominatim but they will increase the database quite a bit because they are so frequent. So I'd prefer to leave them out. These are: particles (between, opposite) and units (number, feet, mile, kilometer, etc.).

Also it is not necessary to repeat abbreviations that already appear in the generic English part (boulevard, street, road, etc.). Better to have a shorter list of abbreviations that are truly specific to Pakistan.

bgo-eiu:
Thank you for your feedback - good to know about the common words and overlapping abbreviations, I will revise to be more specific. Are some of the common ones that are abbreviated in a very short form often in Pakistan still acceptable? For example, `Block -> B` I don't think is as common in other countries which use English in addresses.

Taluka and Tehsil are synonyms yes, I was on the fence about including it. I do try to add both to the Wikidata "other forms" fields where applicable, or to `alt_name` at least, though the administrative subdivisions are not complete in OSM for Pakistan yet.

lonvia:
`block`, `floor`, `house` and `sector` are indeed a little bit problematic because of their different uses world-wide. Thing is: the abbreviations here are currently applied only to names in the OSM database. So it only makes sense to add an abbreviation if there is really a name tag that uses it.

Looking at the current Pakistan data, there is a clear use for 'block' and 'sector'. It make sense to leave that in.

The few usages of 'floor' are rather questionable and will be filtered out by Nominatim (Example: the housenumber in [this node](https://www.openstreetmap.org/node/4107567190)). In all examples I saw, the tag `addr:floor` should rather be used in stead of the actual word. So let's remove 'floor' abbreviations. (There is an open issue to handle unit and floor numbers, see #587.)

'house' is a bit of a mixed case. There are quite a few uses but I suspect that in many cases the abbreviation shouldn't be used, e.g. for [White House Grammar School](https://www.openstreetmap.org/way/895076110). I'd rather be on the conservative side for the moment and leave it out, too. We can revisit the decision later.

Regarding Taluka and Tehsil, adding `alt_name` is indeed the best option, although I see that there are quite a few of those. We will need to sort out synonyms eventually in Nominatim.

lonvia:
Superseeded by #4064.

PR #2779, opened 2022-07-26
Importance by OSM Views

tareqpi:
- Enabled PostGIS to work with raster files
- Nominatim CLI tool can now import OSM views data when [osmviews.tiff](https://qrank.wmcloud.org/download/osmviews.tiff) is found in the `nominatim-project` directory
- Nominatim CLI tool can now refresh OSM views data and recompute the importance score with `nominatim refresh --osm-views --importance`
- Integrated OSM views data into the algorithm that computes the places' importance scores which are used in ranking Nominatim's search results
- Added some unit tests
- Updated the documentation and added a [report](https://github.com/tareqpi/Nominatim/blob/importance-by-osm-views/docs/gsoc22.md) of the experiments conducted about this PR
- Further investigation and testing are needed before merging this branch


.pylintrc
```diff
@@ -13,6 +13,6 @@ ignored-classes=NominatimArgs,closing
 # 'too-many-ancestors' is triggered already by deriving from UserDict
 # 'not-context-manager' disabled because it causes false positives once
 #   typed Python is enabled. See also https://github.com/PyCQA/pylint/issues/5273
-disable=too-few-public-methods,duplicate-code,too-many-ancestors,bad-option-value,no-self-use,not-context-manager
+disable=too-few-public-methods,duplicate-code,too-many-ancestors,bad-option-value,no-self-use,not-context-manager,too-many-statements
```
lonvia, line 16: No global cheating. ;) If you have a function with too many statements that you cannot break up, please add a pylint exception to the function directly.

nominatim/clicmd/setup.py
```diff
@@ -58,6 +58,8 @@ def add_args(self, parser: argparse.ArgumentParser) -> None:
         group2.add_argument('--no-updates', action='store_true',
                            help="Do not keep tables that are only needed for "
                                 "updating the database later")
+        group2.add_argument('--osm-views', action='store_true',
+                           help='Import OSM views GeoTIFF')
```
lonvia, line 62: I think this is not needed anymore.

nominatim/tools/refresh.py
```diff
@@ -146,6 +148,33 @@ def import_wikipedia_articles(dsn: str, data_path: Path, ignore_errors: bool = F
...
+    if not datafile.exists():
+        return 1
+
+    with conn.cursor() as cur:
+        _require_version('PostgreSQL server',
+                         conn.server_version_tuple(),
+                         POSTGRESQL_REQUIRED_VERSION)
+        _require_version('PostGIS',
+                         conn.postgis_version_tuple(),
+                         POSTGIS_REQUIRED_VERSION)
```
lonvia, line 168: You don't need to do the version checks against the required versions again, it has been done on database import. However, as the postgis_raster extension only exists for postgis >= 3.0, you should check for that and return an appropriate error if the version is smaller. The result should be that on import the loading of the tiff is simply skipped and on refresh there is an error.

nominatim/tools/refresh.py
```diff
@@ -146,6 +148,33 @@ def import_wikipedia_articles(dsn: str, data_path: Path, ignore_errors: bool = F
...
+
+        cur.execute('DROP TABLE IF EXISTS "osm_views";')
+        conn.commit()
+
+        cmd = "raster2pgsql -s 4326 -I -C -t 100x100 osmviews.tiff \
```
lonvia, line 173: This loads 'osmvies.tiff' in the current directory not in the 'data_path".

nominatim/tools/refresh.py
```diff
@@ -146,6 +148,33 @@ def import_wikipedia_articles(dsn: str, data_path: Path, ignore_errors: bool = F
...
+
+    with conn.cursor() as cur:
+        _require_version('PostgreSQL server',
+                         conn.server_version_tuple(),
+                         POSTGRESQL_REQUIRED_VERSION)
+        _require_version('PostGIS',
+                         conn.postgis_version_tuple(),
+                         POSTGIS_REQUIRED_VERSION)
+
+        cur.execute('DROP TABLE IF EXISTS "osm_views";')
```
lonvia, line 170: No semicolon needed.

nominatim/tools/refresh.py
```diff
@@ -146,6 +147,30 @@ def import_wikipedia_articles(dsn: str, data_path: Path, ignore_errors: bool = F
...
+    postgis_version = conn.postgis_version_tuple()
+    if postgis_version[0] < 3:
+        return 2
+
+    with conn.cursor() as cur:
+        cur.execute('DROP TABLE IF EXISTS "osm_views"')
+        conn.commit()
+
+        cmd = f"raster2pgsql -s 4326 -I -C -t 100x100 {datafile} \
+            public.osm_views | psql nominatim > /dev/null"
```
lonvia, line 170: You need to use the configured database name here.

lonvia (changes requested):
Please also note the failing CI.

lib-sql/functions/importance.sql
```diff
@@ -98,19 +98,67 @@ $$
...
+  -- add importance by OSM views
+  views := get_osm_views(centroid);
+  result.importance := normalize_osm_views(views) * 0.35;
+
+  -- add importance by wiki data if the place has one
   FOR match IN SELECT * FROM get_wikipedia_match(extratags, country_code)
                WHERE language is not NULL
   LOOP
-    result.importance := match.importance;
+    result.importance := COALESCE(result.importance, 0) + match.importance * 0.65;
```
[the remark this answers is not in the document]
tareqpi: `result.importance` will be null when the user chooses not to import the GeoTIFF file.

test/python/cli/test_cmd_import.py
```diff
@@ -68,7 +69,7 @@ def test_import_full(self, mock_func_factory, with_updates, place_table, propert
         assert cf_mock.called > 1
 
         for mock in mocks:
-            assert mock.called == 1, "Mock '{}' not called".format(mock.func_name)
+            assert mock.called > 0, "Mock '{}' not called".format(mock.func_name)
```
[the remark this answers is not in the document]
tareqpi: The assertion was returning an error where `mock.called` was returning 2 instead of 1 when testing `import_osm_views_geotiff`. Initially, I thought it was testing the error scenarios for `import_osm_views_geotiff` since the function has two error scenarios: when `osmviews.tiff` is not found (function returns 1)  and when the postgis version is less than 3 (function returns 2). However, it was because `import_osm_views_geotiff` was called twice by mistake which I had previously fixed but forgot to change back the assertion statement.

lonvia (approved):

lonvia (changes requested):
For the record, here is a list of things that will need to be done before page views can be integrated:

- [ ] Move views table preparation to external project and smooth view numbers
- [ ] Use views as tie breaker only
- [ ] add tests

#### External table preparation

raster2pgsql simply creates an SQL file, so there really is no need to call it as part of the import process. Instead the whole views table preparation should go into an external project which creates a `tile_views.sql` out of the view logs and which then is simply read with `execute_file`. This would be exactly analogous to the wikimedia importance data which is prepared in the [wikipedia-wikidata](https://github.com/osm-search/wikipedia-wikidata) project and then simply imported.

The view logs processing script will also need some additional data filter and sanitation steps to filter out odd outliers.

#### Views as tie breaker

The view log numbers are not consistent with each other over the planet, simply because OSM is more popular in some continent and countries than other. This means that view numbers are not useful as absolute importance values. They are still useful as tie breakers. This means that the final importance formula should be something like:

    coalesce(wikipedia_importance, 0.75001 * rank_search/40) + 0.0001 * relative_tile_views

#### Tests

Add a small geographic extract of the final tileview SQL file covering `test/testdb/apidb-test-data.pbf` into the repo and make sure it is used in the BDD API tests.

Add unit tests for importing the tileviews based on the test extract above.

lonvia:
Finished in #2835.

PR #2788, opened 2022-08-05
Removing on-the-fly importance computation

tareqpi:
- Removed the on-the-fly default importance computations from the PHP side and moved it to `importance.sql`


lib-php/PlaceLookup.php
```diff
@@ -227,7 +227,7 @@ public function lookup($aResults, $iMinRank = 0, $iMaxRank = 30)
             }
             $sSQL .= '    avg(ST_X(centroid)) AS lon, ';
             $sSQL .= '    avg(ST_Y(centroid)) AS lat, ';
-            $sSQL .= '    COALESCE(importance,0.75-(rank_search::float/40)) AS importance, ';
+            $sSQL .= '    importance, ';
```
tareqpi, line 230: When removing the on-the-fly default importance computation here, the mentioned issue arises

lib-php/PlaceLookup.php
```diff
@@ -300,10 +300,11 @@ public function lookup($aResults, $iMinRank = 0, $iMaxRank = 30)
                 $sSQL .= 'null::text AS names,';
             }
             $sSQL .= '  ST_x(geometry) AS lon, ST_y(geometry) AS lat,';
-            $sSQL .= '  (0.75-(rank_search::float/40)) AS importance, ';
+            $sSQL .= '  importance, ';
             $sSQL .= $this->addressImportanceSql('geometry', 'lp.parent_place_id');
             $sSQL .= '  null::text AS extra_place ';
-            $sSQL .= 'FROM location_postcode lp';
+            $sSQL .= 'FROM location_postcode lp ';
+            $sSQL .= 'JOIN placex USING (place_id)';
```
lonvia, line 307: Postcodes don't have an importance, as you correctly noticed. Let's keep the rank formula here.

lib-sql/functions/importance.sql
```diff
@@ -125,7 +125,8 @@ BEGIN
     END LOOP;
   END IF;
 
-  RETURN null;
+  result.importance := 0.75 - (rank_search / 40);
```
lonvia, line 128: There have been issues with artihmetics when the importance was exactly 0. Could you change the formula to using 0.750001, so we never hit the 0?

lonvia (commented):
I tried this and looks good. There was one  missing importance for place=country.

tareqpi:
This pull request has been closed due to the need of updating `placex` table by removing all the null values in the `importance` column which is very expensive since `placex` has a large number of indexes.

PR #2856, opened 2022-11-02
Update DB.php

t-tomek:
Use SET instead of UPDATE queries to support read only databases

### DIFF
--- a/lib-php/DB.php
+++ b/lib-php/DB.php
@@ -38,23 +38,28 @@ class DB
 
         // https://secure.php.net/manual/en/ref.pdo-pgsql.connection.php
         try {
-            $conn = new \PDO($this->sDSN, null, null, $aConnOptions);
+            $this->connection = new \PDO($this->sDSN, null, null, $aConnOptions);
         } catch (\PDOException $e) {
             $sMsg = 'Failed to establish database connection:' . $e->getMessage();
             throw new \Nominatim\DatabaseError($sMsg, 500, null, $e->getMessage());
         }
-
-        $conn->exec("SET DateStyle TO 'sql,european'");
-        $conn->exec("SET client_encoding TO 'utf-8'");
+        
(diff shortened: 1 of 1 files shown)


lonvia:
UPDATE is used because it won't fail when the Postgresql version doesn't have the setting in question. To switch to SET you have to make it work for Postgresql < 11.

mtmail:
@lonvia Are the settings needed for Postgresql < 11 ? If I recall JIT feature was added first in Postgresql 11 https://www.percona.com/blog/2018/10/18/postgresql-11-our-first-take-on-the-new-release/

lonvia:
No, they are not. UPDATE is a noop on these versions. That's the beauty of it.

otbutz:
Postgres 9.6 is already EOL and Postgres 10 will follow in about a week.

lonvia:
That's not relevant.

mtmail:
@t-tomek `DB.php` library has a method ` getPostgresVersion()` so if you move the `$this->connection =` a bit higher it should be possible to check for the version and make adjustments.
I hope postgresql < 11 get deprecated in future Nominatim versions but that'd be part of a different discussion.
The current CI tests include tests for Postgresql 9.6, 13, 14
```
                    - ubuntu: 18
                      postgresql: 9.6
                      postgis: 2.5
                      pytest: pytest
                      php: 7.2
```

t-tomek:
@lonvia  I've just updated the patch. Now Postgres >= 11 uses SET, < 11 UPDATE statement.

t-tomek:
I noticed that PG 9.5 support was dropped so I left SET statments (as max_parallel_workers_per_gather is supported since 9.6) but jit_above_cost is changed only if PG version is above 11

thomas-hansen-grindr:
It seems like this change didn't make it into the 4.2.1 release available [here](https://nominatim.org/release/Nominatim-4.2.1.tar.bz2)

PR #2862, opened 2022-11-03
Install scripts: remove version from /var/run/php-fpm filenames

mtmail:
This is to match closer with https://nominatim.org/release-docs/latest/admin/Deployment/#nominatim-with-nginx where users can get confused (e.g. https://github.com/osm-search/Nominatim/discussions/2861#discussioncomment-4051544)

There was also a typo in the Ubuntu-22 script listing two different versions in the same config file.

### DIFF
--- a/vagrant/Install-on-Ubuntu-22.sh
+++ b/vagrant/Install-on-Ubuntu-22.sh
@@ -197,7 +197,7 @@ if [ "x$2" == "xinstall-nginx" ]; then #DOCS:
 sudo tee /etc/php/8.1/fpm/pool.d/www.conf << EOF_PHP_FPM_CONF
 [www]
 ; Replace the tcp listener and add the unix socket
-listen = /var/run/php8.1-fpm.sock
+listen = /var/run/php-fpm-nominatim.sock
 
 ; Ensure that the daemon runs as the correct user
 listen.owner = www-data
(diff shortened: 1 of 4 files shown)


vagrant/Install-on-Ubuntu-22.sh
```diff
@@ -197,7 +197,7 @@ if [ "x$2" == "xinstall-nginx" ]; then #DOCS:
 sudo tee /etc/php/8.1/fpm/pool.d/www.conf << EOF_PHP_FPM_CONF
 [www]
 ; Replace the tcp listener and add the unix socket
-listen = /var/run/php8.1-fpm.sock
+listen = /var/run/php-fpm.sock
```
lonvia, line 200: This change is missing in the two other scripts.

lonvia (commented):
I wonder if the socket name shouldn't better be something less general like `php-fpm-nominatim-www.sock` to ensure it does not clash with anything else running PHP on the machine.

vagrant/Install-on-Ubuntu-22.sh
mtmail: Good idea, I changed that now but without the `-www` part. You can doublecheck with `grep -r '\.sock' docs/ vagrant/`

PR #2908, opened 2022-11-29
Update Vagrantfile

n-timofeev:
Added support for Hyper-V. Removed missing CentOS configs.


lonvia:
Thank you. This update was long overdue.

PR #2919, opened 2022-12-08
Fix typo in NOMINATIM_LOG_FILE

woodpeck:
execution time is in ms, not s


lonvia:
The time is indeed in 's'. It's the other part of the docs that is wrong.

Issue #2754, opened 2022-06-20
Indicate that a search result is from TIGER data


simonpoole:
**Is your feature request related to a problem? Please describe.**
For a "normal" user the difference between results from TIGER and OSM data are not clear and can be surprising. In particular clicking on the result will select and display the street without any further explanation.

See for example https://www.openstreetmap.org/search?query=16%20Slayton%20Road%20melrose%20Massachusetts.#map=19/42.45596/-71.04467

**Describe the solution you'd like**

An indication in the results display that this is not a "proper" result from OSM.

lonvia:
I think we use to return a 'T' type for Tiger house numbers in `osm_type` but changed that when going to spelled out OSM types.  @mtmail do you happen to have a reference to the discussion back then? If I remember correctly the reasoning was that `osm_type` should not contain surprising values. And that still holds. Putting a new value in there is a sure way to break clients. 

Possible solutions:

1. Leave out `osm_type` completely to indicate a non-OSM source. This is already done with artificial postcodes, so clients must be able to handle that. We'd loose the parent information. osm.org uses it to go to the parent street when clicking on the result.
2. Add an additional source field. This could be generalized to any external data and combined with #1699.

Note that none of this will resolve that the source is shown on osm.org. Additional changes would be needed over there no matter what we do.

mtmail:
>  @mtmail do you happen to have a reference to the discussion back then?

That might have been via email. It came up when https://github.com/osm-search/nominatim-ui was developed because while the old PHP HTML templates were able to infer a place came from TIGER data via internal variables the JSON/XML templates where missing the information. It remained an unsolved issue.

mtmail:
Original test case no longer works because OSM mappers have added OSM house numbers around the location. New test case from another nominatim-docker discussion

1724 9th Avenue North, Bessemer, AL 35020, United States
https://nominatim.openstreetmap.org/ui/search.html?q=1724+9th+Avenue+North%2C+Bessemer%2C+AL+35020%2C+United+States

PR #2970, opened 2023-02-06
Python implementation of details endpoint

### LINKED ISSUE #1671: Improve details page for non-OSM objects
Nominatim sometimes presents results that are not directly related to an OSM object. These are:

* postcode (country + postcode + a parent OSM object)
* interpolations (OSM way + a housenumber)
* Tiger house numbers (OSM way + Tiger interpolations + housenumber)
* auxillary data (Currently unused)
* countries/country codes (osm_country_name table + location_country table for geometry)

These objects don't have good details pages at the moment. They usually simply redirect to their OSM parent object. This can be confusing at times. It would be nice if the available data for these objects is also presented in the details page.

Todo:
- [x] document auxiliary data and how it is returned to the user
- [ ] extend parameters of details page, so that auxiliary data can be uniquely linked (see also #1651)
- [ ] extend details API output
- [ ] nominatim-ui: extend linking to details of special objects
- [ ] nominatim-ui: extend details page to show additional information about the auxiliary object

lonvia:
This adds a new API function `lookup()` and the `/details` endpoint for the python server and CLI.

The PR contains a lot of code for handling search results: the datatype class itself, formatting, retrieval of extra information. I briefly considered simply returning result rows from SQLAlchemy but discarded the idea because it would make the Nominatim library API depended on internals of a library we use. I prefer to decouple the API from the implementation details. The disadvantage is that there is now quite a bit of code for building the results. Quite a lot of the code will be reusable for the `/revers` and `/search` endpoints but I do fully expect some refactoring as we might not want to expose the full contents of the database for ordinary searches.

The Python implementation comes with some minor changes in behaviour compared to the PHP endpoint:
* Drop `pretty=1` parameter. This should be done on the client side. Browsers usually do that automatically and for other uses there are plenty of client libraries. The CLI version of `details` will produce always produce pretty-printed output.
* Details for Tiger data, postcodes and interpolations can now be returned directly. The PHP version would only return information about the parent places. Interpolations can be addressed with OSM type + ID (and optionally housenumber in the class field), all others via place_id only right now. So there is still a bit of work to be done to be able to close #1671. Also still open is a proper return of the source (see #2754). The API result already has a `source` field for that purpose but that is not yet reflected in what the endpoint returns.
* Empty fields are now dropped in the response. The PHP implementation did return the fields with a value of `null`. This saves a tiny bit of bandwidth.

Also contains various adaptions because new versions pf pylint and PHPUnit have been released.


lonvia:
Github actions is hopelessly broken today. For a successful run of this branch see https://github.com/lonvia/Nominatim/actions/runs/4093277375

Issue #2953, opened 2023-01-22
New sanitizer: delete-tags


lonvia:
In order to filter out odd tagging practises like the postbox refs in Germany
(see #2652) we could do with a sanitizer that removes some names and address parts from
the list of indexed names. It should work along those lines:

### Sanitizer 'delete-tags'

Removes a name or address name from the list, when it matches all properties
given in the parameter list. There are two kinds of properties. Some refer to
the name and some refer to the place being indexed. If you specify only
place properties, then all names and address parts will be deleted and the
place will effectively not be searchable.

Where parameters contain regular expressions, they are always matched against
the full string. Add an exclamation mark (!) in front of the expression to
negate the match, i.e. delete the tag only if the property does not match.

When a parameter contains a list of conditions, a single match is sufficient.

Parameters:

```
 type - Either 'name' or 'address'.
 kind - A single string or a list of strings containing regular
        expressions that are matched against the 'kind' property of the name.
 suffix - A single string or a list of strings containing regular
          expressions that are matched against the 'suffix' property.
 name - A single string or a list of strings containing regular
        expressions that are matched against 'name' property of the name.
 country_code - A single string or list of strings containing two-letter
                lower-case country codes. Only places within one of the listed
                countries will be deleted.
 rank_address - A single string or list of strings containing either a single
                number referring to a rank or a range of the form <from>-<to>.
```

lonvia:
If somebody wants to take this up, here are some hints to get you started:

* there is a tutorial on creating sanitizers at https://nominatim.org/tutorials/customize-transliteration.html
* the reference for sanitizers can be found at https://nominatim.org/release-docs/latest/develop/ICU-Tokenizer-Modules/#custom-sanitizer-modules
*  have a look at the existing sanitizers in https://github.com/osm-search/Nominatim/tree/master/nominatim/tokenizer/sanitizers. This is also where the new sanitizer should go.
* unit tests for existing sanitizers are in https://github.com/osm-search/Nominatim/tree/master/test/python/tokenizer/sanitizers. Add a new test for the new sanitizer.

SainiAditya1:
assign this to me , I would like to work on this ..

lonvia:
We do not assign tasks. Just pick what interests you. Please make sure to read the [GSOC instructions](https://wiki.openstreetmap.org/wiki/User:Lonvia/GSoC_2021_Nominatim_Projects).

biswajit-k:
[...]

Issue #2652, opened 2022-03-30
Parenthesis in City names in germany leads to empty results


c-holtermann:
## What did you search for?

Different cities in germany with a Parenthesis in the name.

1) "21423 Winsen (Luhe) Bahnhofstraße 2".

https://www.openstreetmap.org/search?query=21423%20Winsen%20(Luhe)%20Bahnhofstra%C3%9Fe%202#map=19/53.36028/10.20633

2) "29456 Hitzacker (Elbe) Gelderländer Straße 12"

https://www.openstreetmap.org/search?query=29456%20Hitzacker%20(Elbe)%20Gelderl%C3%A4nder%20Stra%C3%9Fe%2012#map=19/53.36028/10.20633

## What result did you get?

None in both cases.

## What result did you expect?

1) https://www.openstreetmap.org/way/96068117
2) https://www.openstreetmap.org/way/277522980

## Further details

1) addr:city of the mentioned result actually contains the full name of the city which is "Winsen (Luhe)".
2) The region (https://www.openstreetmap.org/relation/1821294) has the name Hitzacker with a name:suffix (Elbe)

lonvia:
Paranthesis have nothing to do with it.

The first issue is easily fixed by placing strategic commas, e.g.:
https://nominatim.openstreetmap.org/ui/search.html?q=21423%2C+Winsen+%28Luhe%29+Bahnhofstra%C3%9Fe+2
(For reference, the reason Nominatim is thrown off is the odd ref mapping of postboxes in Winsen(Luhe), e.g. https://www.openstreetmap.org/node/2948710055. In the long run, Nominatim needs to detect those bogus refs and drop them.)

The second issue is a data issue: for [Hitzacker](https://nominatim.openstreetmap.org/ui/details.html?osmtype=R&osmid=1821294&class=boundary), the name with suffix "(Elbe)" is missing. I see it is there on the place node, so  #2637 will have fixed the issue already. Just needed a data update.

c-holtermann:
Thank you for your reply @lonvia!

1) Unfortunately I cannot change the way the adress gets searched just now. I'd be happy to set a comma but I don't have access to the source code.

Could you elaborate on the ref problem?

2) Actually works now

lonvia:
> Could you elaborate on the ref problem?

[...]

PR #2993, opened 2023-03-02
Adds sanitizer for preventing certain tags to enter search index based on parameters

biswajit-k:
References #2953.
It would prevent certain tags which are only used for referencing some place to enter the search index, as they are not part of the current object also these tags wrongly populate the search index because of which sometimes actual places are pushed to lower rankings(or even removed) in search. For Example: #2652.

### DIFF
--- a/nominatim/tokenizer/sanitizers/delete_tags.py
+++ b/nominatim/tokenizer/sanitizers/delete_tags.py
@@ -1,0 +1,144 @@
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# This file is part of Nominatim. (https://nominatim.org)
+#
+# Copyright (C) 2023 by the Nominatim developer community.
+# For a full list of authors see the git log.
+"""
+Sanitizer which prevents certain tags from getting into the search index.
+It remove tags which matches all properties given below.
+
+
+Arguments:
+    type: Define which type of tags should be considered for removal.
+          There are two types of tags 'name' and 'address' tags.
+          Takes a string 'name' or 'address'. (default: 'name')
+
+    filter-kind: Define which 'kind' of tags should be removed.
+                 Takes a string or list of strings where each
+                 string is a regular expression. A tag is considered
+                 to be a candidate for removal if its 'kind' property
+                 fully matches any of the given regular expressions.
+                 Note that by default all 'kind' of tags are considered.
+
+    suffix: Define the 'suffix' property of the tags which should be
+            removed. Takes a string or list of strings where each
+            string is a regular expression. A tag is considered to be a
+            candidate for removal if its 'suffix' property fully
+            matches any of the given regular expressions. Note that by
+            default tags with any suffix value are considered including
+            those which don't have a suffix at all.
+
+    name: Define the 'name' property corresponding to the 'kind' property
+          of the tag. Takes a string or list of strings where each string
+          is a regular expression. A tag is considered to be a candidate
+          for removal if its name fully matches any of the given regular
+          expressions. Note that by default tags with any 'name' are
+          considered.
+
+    country_code: Define the country code of places whose tags should be
+                  considered for removed. Takes a string or list of strings
+                  where each string is a two-letter lower-case country code.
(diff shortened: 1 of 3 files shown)


biswajit-k:
Hi @lonvia, I have made the required changes, please have a look.

nominatim/tokenizer/sanitizers/delete_tags.py
```diff
@@ -0,0 +1,146 @@
...
+
+        filtered_tags: List[PlaceName] = []
+
+        for tag in tags:
+            keep_tag = True
+
+            if self.filter_kind(tag.kind):
+
+                if (obj.place.country_code in self.country_codes and
+                    self._in_rank_addresses(obj.place.rank_address)):
```
lonvia, line 88: These two conditions apply to the whole object, not to the individual tags. That means, if they are not fulfilled, none of the tags will filtered. You can therefore have a shortcut and check those up in line 77 already.

nominatim/tokenizer/sanitizers/delete_tags.py
```diff
@@ -0,0 +1,146 @@
...
+                            else (intv[0], intv[1]))
+
+        return intervals
+
+
+    def _in_rank_addresses(self, rank: int) -> bool:
+        """ Return True if the given rank address lies in any of the
+            rank address intervals. Otherwise, returns False.
+         """
+        return any(l <= rank <= r for l, r in self.rank_address_intervals)
```
lonvia, line 127: You can do the check for rank more efficiently here. You just need to do a bit more preprocessing: we know that there are only ranks 0-30. So during preprocessing of the incoming configuration you can create a set of allowed rank numbers. That means spell out every single number to be accepted. For example, from the default configuration of '0-30' you can create a `allowed_ranks = set(range(0, 31))` and then simply check `obj.place.rank_address in allowed_ranks`.

(Even more advanced and faster: An alternative to a set would be a tuple of 31 booleans, so that eventually the lookup is `allowed_ranks[obj.place.rank_address]`.)

lonvia (commented):
This looks quite good already.

Your tests do not yet cover the use case, where only some of the arguments are given. For example, if the `country_code` parameter is not specified, then the filter should not check for country codes at all.

And thanks for fixing all my bad spelling. If you want to get it out of the way while working on the code comments, feel free to split them out in a separate PR. That can be merged quickly then.

As a general comment: you are welcome to rebase the branch for the PR at any time. So if you want to fix up or squash commits, feel free to do so.

biswajit-k:
Thanks for the review @lonvia. As suggested, I have raised a separate PR which fixes typos [here](https://github.com/osm-search/Nominatim/pull/2999). 
I have refactored the logic for  `__call__` method and also for rank address. Also, I have added default parameters to the sanitizer configuration and made changes in the tests accordingly. Please have a look.

lonvia (commented):
Looks mostly good now.

One more tiny thing: I did not mean to completely replace the test which has all parameters set but rather add to it. So if you could bring back the test which has all parameters set and checks that things still go well, this PR is good to go.

biswajit-k:
Just to confirm, now we need to add tests where all parameters are set in the sanitizer configuration.

lonvia:
Yes. You now have unit tests that make sure input for each single parameter is handled correctly. That is good. It gives a good code coverage for the tests. However, just because we know the single parts work, doesn't mean that they still do the right thing when combined. Testing all possible combinations of possible parameters is too much. So the second best thing is to have a test with all parameters set, tested against a couple of meaningful data points.

biswajit-k:
Hi @lonvia, I have added the tests. Please have a look.

lonvia:
Thank you!

If you are interested in a bit of a follow-up: The `_matches` function you have added is very similar to what `config.get_filter_kind()` does. If you generalise `get_filter_kind()`, so that it takes the name of the config parameter as an argument, it could be reused for the `name` and `suffix` checks and you could get rid of `_matches`.

biswajit-k:
That's a great idea! I will create a new PR for this.

PR #3006, opened 2023-03-11
generalize filter function for sanitizers

biswajit-k:
As discussed [here](https://github.com/osm-search/Nominatim/pull/2993) filter functions are commonly used by sanitizers to process only selected tags of places.

So, instead of having separate filters for each parameter of sanitizer configuration, we can have a generalized filter function `get_filter()` which takes the parameter name as one of the arguments and creates a filter function for it.


biswajit-k:
Hi @lonvia, could you please review the PR? Thanks.

test/python/tokenizer/sanitizers/test_sanitizer_config.py
```diff
@@ -78,36 +74,61 @@ def test_create_split_regex_empty_delimiter():
...
+            "name": "abc",
+            "filter-kind": "lang.*",
+        }
+    ).get_filter("filter-kind", [".*fr"])
 
     assert filt(kind)
 
 
-@pytest.mark.parametrize('kind', ('de ', '123', '', 'bedece'))
+@pytest.mark.parametrize("kind", ("de ", "123", "", "bedece", None))
```
lonvia, line 109: Please go with the existing quoting style.

The convention for quotes is:
 * use single quotes for all simple strings unless the string contains a single string (so, it should be 'London' but "London's subway")
 * use double quotes for format strings (i.e. `f"London {foo}"`)

I know that this convention is not followed everywhere but we should at least get it consistent in new code.

test/python/tokenizer/sanitizers/test_sanitizer_config.py
```diff
@@ -78,36 +74,61 @@ def test_create_split_regex_empty_delimiter():
...
+
+    assert not filt(kind)
+
+
+@pytest.mark.parametrize("kind", ("lang", "lang:de", "langxx"))
 def test_create_kind_filter_custom_regex_positive(kind):
-    filt = SanitizerConfig({'filter-kind': '.*de'}).get_filter_kind()
+    filt = SanitizerConfig(
+        {
+            "name": "abc",
```
lonvia, line 101: Why is this additional parameter needed?

biswajit-k:
Hi @lonvia, I have made the required changes. Also, noted the suggestions for the future. Thanks!

test/python/tokenizer/sanitizers/test_delete_tags.py
```diff
@@ -289,7 +289,7 @@ def run_sanitizer_on(self, country_code, rank_addr, suffix, **kwargs):
 
     def test_string_arguments_pass(self):
         res = self.run_sanitizer_on('de', '25-30', r'[\s\S]*',
-                                    name='foo', ref='foo', name_abc='bar', ref_abc='baz')
+                                    name_xyz='foo', ref_pqr='foo', name_abc='bar', ref_abc='baz')
```
lonvia, line 292: Can you please explain why this needed to be changed?

nominatim/tokenizer/sanitizers/config.py
```diff
@@ -102,30 +106,41 @@ def get_delimiter(self, default: str = ',;') -> Pattern[str]:
...
-                True when the item passes the filter.
+                A filter function that takes a target string as the argument and
+                returns True if the it fully matches any of the regular expressions
+                otherwise returns False.
+                Note that if instead of target string, None type is provided then
+                it simply returns False.
         """
-        filters = self.get_string_list('filter-kind', default)
+        values = self.data.get(param, None)
+        filters = self.get_string_list(param) if values is not None else default
```
lonvia, line 137: What is the motivation to change the semantics of `get_string_list()`? It just makes the code here a lot more complicated. Without the change, you can simply leave the code here untouched and it works.

lonvia (commented):
Sorry, two more small questions.

nominatim/tokenizer/sanitizers/config.py
biswajit-k: Actually, here we want to check if both `values` and `default` are `none` so I am explicitly checking that. With the current definition of `get_string_list`, it should not return `none`, because type checking would complain.

 And in case we allow it to also return `none`, we would explicitly have to check for `none` values wherever we use it. Also, I think it would be good idea to let the function only return a list as it would make the function return value be more predictable.

test/python/tokenizer/sanitizers/test_delete_tags.py
biswajit-k: This is because of the functionality of `get_filter` function. If we want all suffixes to pass, we should ideally not provide the suffix parameter. If we provide a suffix parameter, it will not allow empty suffixes. This is because the function treats `none` and empty string as different.

nominatim/tokenizer/sanitizers/config.py
```diff
@@ -102,30 +106,41 @@ def get_delimiter(self, default: str = ',;') -> Pattern[str]:
...
 
-        if not filters:
+        if filters is None:
```
lonvia, line 139: This changes the behaviour when the parameter is present but empty. It's an odd edge case but the idea was to have it behave as if the parameter was not present at all.

test/python/tokenizer/sanitizers/test_delete_tags.py
lonvia: Right. The `None` case is really hard to wrap your head around. I think for non-programmers it is easier to understand if an absent suffix is handled like the empty string. It would also make the code in `get_filter` easier because the function can go back to accepting `str` only.

biswajit-k: I agree with this. For these scenarios, we can consider `None` value as an empty string safely. I'll add the changes.

nominatim/tokenizer/sanitizers/config.py
biswajit-k: That could be done, but I would prefer to be it like this.

nominatim/tokenizer/sanitizers/config.py
biswajit-k: I see. This is less intuitive unless explicitly mentioned. What I can think of is - we can do either of two things-
1. Go with these changes and let `[]` keep its meaning i.e allow nothing. Because to allow anything to pass we can just skip the parameter, which ideally makes sense.
2. We can revert back to the original code and explicitly mention that `[]` is just like skipping the parameter. Also with that, if I allow `get_string_list` to take `None` default and return an empty list for it, we could bring back the `get_string_list(param, default)` thing.

Please let me know what you think and then I will give it a go.

biswajit-k:
Hi @lonvia, I have made the following changes- 
1. Treat `suffix` as an empty string when it is `None` 
2. Treat parameter value as missing when it is omitted or its value is missing in yaml
3. Add explicit values to the `default` parameter

Request you to please have a look.

PR #3020, opened 2023-03-30
Python implementation of reverse

lonvia:
This adds a new function `reverse()` to the Python API and the `/reverse` endpoint for server and CLI.

The implementation largely follows the PHP implementation with two exceptions:
1. There is a new `layer` parameter which allows to restrict the kind of features to be returned. The layer parameter replaces the hard-coded exceptions in the PHP lookup code.
2. The long deprecated lookup by osm ID is now gone. Use `/lookup` instead.

The output type of the `details()` function is changed to a `DetailedResult`. It turned out that handling is much easier, when each function has its dedicated returned type.

A large part of the new code contains the output formatting for all the different formats we offer. I've tried to port the output formats with all their particularities and inconsistencies to ensure there are few visible changes for the end user. The code is written, so that the same output formatting can be used for search later, making that part of the port a bit more lightweight.


ianvernon:
@lonvia nice work! Out of curiosity, what motivated creation of aPython version ?

lonvia:
You can find some of the background in this blog post: https://nominatim.org/2022/12/08/frontend-renovation.html

Issue #2938, opened 2023-01-04
Update Taginfo Projects


danieldegroot2:
Show the OpenStreetMap community what [data](https://wiki.openstreetmap.org/wiki/Tags) you are using. Update your project on the [list](https://taginfo.openstreetmap.org/projects) of [Taginfo Projects](https://github.com/taginfo/taginfo-projects).

lonvia:
Somebody would have to write a converter from the [lua import style](https://nominatim.org/release-docs/develop/customize/Import-Styles/) to a taginfo project description. The original one is not up to date any longer.

biswajit-k:
I would like to work on this issue if it's available. 
I had gone through the above lua import style guide and the format required for the project description. Also, could you please provide the source of the old converter? I searched for it but couldn't find it. It would be quite helpful for reference.

mtmail:
The output file should be https://nominatim.openstreetmap.org/taginfo.json I'm not aware that the file was created from a script previously, I think it was manual.

lonvia:
Indeed. This file was generate by hand from the old [import styles](https://github.com/osm-search/Nominatim/blob/4.2.x/settings/import-extratags.style).

My idea here was to write a small lua script, that imports a Nominatim lua style and then outputs the taginfo json file by just printing to stdout. There is a dirt trick, how to make that work. Write your own flex-base.lua, which implements their own variant of the convenience functions described [here](https://nominatim.org/release-docs/develop/customize/Import-Styles/#changing-the-recognized-tags). The parameters for these functions have all the relevant tags in handy lists, so they can be easily collected and afterwards printed.

biswajit-k:
Thanks, @mtmail, @lonvia for the information and suggestion. I had gone for [this issue](https://github.com/osm-search/Nominatim/issues/2953) and would come to this one when its done.

lonvia:
The taginfo.json generated by @biswajit-k's code is now live on https://nominatim.openstreetmap.org/taginfo.json.

PR #3045, opened 2023-04-26
Add script to generate taginfo project description

biswajit-k:
As mentioned in #2938, Nominatim now uses [flex style](https://nominatim.org/release-docs/develop/customize/Import-Styles/) to import places into the database. The flex-style files are present in the [settings](https://github.com/osm-search/Nominatim/tree/master/settings) folder. These files contain the OSM tags which are being used by Nominatim.
The script collects these tags and prints the Taginfo project description in the required [format](https://wiki.openstreetmap.org/wiki/Taginfo/Projects#Project_Files).

It should be noted that `delete` and `extra` tags are not considered in the description because these tags are not part of the search process. These tags are either completely ignored or are just used for displaying extra information with the search results.

### DIFF
--- a/.github/actions/build-nominatim/action.yml
+++ b/.github/actions/build-nominatim/action.yml
@@ -25,7 +25,7 @@ runs:
           shell: bash
         - name: Install${{ matrix.flavour }} prerequisites
           run: |
-            sudo apt-get install -y -qq libboost-system-dev libboost-filesystem-dev libexpat1-dev zlib1g-dev libbz2-dev libpq-dev libproj-dev libicu-dev liblua${LUA_VERSION}-dev lua${LUA_VERSION}
+            sudo apt-get install -y -qq libboost-system-dev libboost-filesystem-dev libexpat1-dev zlib1g-dev libbz2-dev libpq-dev libproj-dev libicu-dev liblua${LUA_VERSION}-dev lua${LUA_VERSION} lua-dkjson
             if [ "$FLAVOUR" == "oldstuff" ]; then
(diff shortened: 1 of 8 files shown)


biswajit-k:
Hi @lonvia, request you to please review the PR.

lonvia:
Sorry, I'm really busy at the moment but hope to get to this before the end of the week.

lonvia:
The trick with the shell script is very clever but it can break too easily when the code in flex-base.lua changes. I would be perfectly happy with a solution, where the `MAIN`,  `NAME_KEYS` and `ADDRESS_KEYS` structures are created in the `flex-base.lua` file directly and are made available from the module. The extra code will only have an effect on the setup of the module and that is not performance critical. Once you have that, you can simply execute the original lua script at the beginning of your script and then run your code without additional shell scripting around.

The dkjson library is already packaged in Debian/Ubuntu, so please use the package instead of adding it to the Nominatim source code.

lonvia:
Looks good in general now, I'll have a bit of a closer look later. In the meantime, could you add the taginfo file generation for one of the default styles to the CI tests around [here](https://github.com/osm-search/Nominatim/blob/d69411f414c6e93e558e01e1736f1628fcec67ab/.github/workflows/ci-tests.yml#L257).

biswajit-k:
Hi @lonvia, I have made the required changes. Please have a look.

lonvia:
Thank you, works nicely.

I noticed that the project information is severely out of date. But that has nothing to do with your code, so I'll add a commit on top for that.

PR #3090, opened 2023-06-19
check-database on frozen db shouldnt recommend indexing

mtmail:
Not beautiful output I have to admit. Should we not a print a warning if un-indexed places is below a certain percentage? Even with whole planet I rarely see over 500.

```
Checking database connection ... OK
Checking for placex table ... OK
Checking for placex content ... OK
Checking that tokenizer works ... OK
Checking for wikipedia/wikidata data ... not applicable
Checking indexing status ... WARNING

 The indexing didn't finish. 4 entries are not yet indexed.

 To index the remaining entries, run:               Database is marked frozen, it cannot be updated.
Low counts of unindexed places are fine.

Checking that database indexes are complete ... OK
Checking that all database indexes are valid ... OK
Checking TIGER external data table. ... not applicable
```


lonvia:
I wonder if we should simply rerun indexing even on import to make sure the count is at 0 before freezing the database.

lonvia:
Please rebase on master. The mypy issues should be fixed there.

mtmail:
> I wonder if we should simply rerun indexing even on import to make sure the count is at 0 before freezing the database.

Is there a quiet indexing method? The default one goes through all ranks and prints 100s of lines to the output. When you know you index only a couple of places it's quite verbose.

lonvia:
You can run indexing in quiet mode (`-q`).

PR #3096, opened 2023-06-27
Parameterise the search only within countries

alfmarcua:
If, despite not finding a point within the static grid of countries, it finds a geometry of a region, do not return the geometry. Return "Unable to geocode" instead. By default this parameter is False for Nominating to work as before.

### DIFF
--- a/lib-php/ReverseGeocode.php
+++ b/lib-php/ReverseGeocode.php
@@ -85,6 +85,11 @@ class ReverseGeocode
 
     protected function lookupLargeArea($sPointSQL, $iMaxRank)
     {
+        if(CONST_Search_WithinCountries
+            and $this->lookupInCountry($sPointSQL, $iMaxRank) == null){
+                return  null;
+        }
+
         if ($iMaxRank > 4) {
             $aPlace = $this->lookupPolygon($sPointSQL, $iMaxRank);
             if ($aPlace) {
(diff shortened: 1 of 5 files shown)


lib-php/ReverseGeocode.php
```diff
@@ -85,6 +85,11 @@ protected function lookupInterpolation($sPointSQL, $fSearchDiam)
 
     protected function lookupLargeArea($sPointSQL, $iMaxRank)
     {
+        if(CONST_Search_WithinCountries
+            and $this->lookupInCountry($sPointSQL, $iMaxRank) == null){
+                return  null;
+        }
```
lonvia, line 91: This code is nicely readable but unfortunately leads to 1 - 2 additional unnecessary SQL queries to the database for every reverse query.

Split `lookupInCountry` in two functions, one which looks up the country code in the country grid (lines 107-120), one that looks up country information (lines 122ff). Then call here the first part, remembering the country code. And in line 102 call the second part using the remembered country code.

alfmarcua: Thank you so much. I added a commit that splits `lookupInCountry` in two functions and document the parameter.

lonvia:
Looks good now. Do you mind rebasing and resolving the resulting conflicts?

PR #3108, opened 2023-07-07
These days the OSM wikipedia tab no longer contains URLs

mtmail:
I believe supporting Wikipedia values containing URLs is no longer needed.

Out of 1.7m values in OSM planet only 68 contain either 'http' or 'https'. 3 contain 'wikipedia.org', 21 '/wiki/', 517 don't contain any colon. Together less than 0.03%. I'd consider them all data errors. For example language prefix forgotten, ; instead of :, wikidata ids.

The `lower(trim(` on the `wiki_article_language` is probably also not needed. I found only 5 values where the language was uppercase, none with spaces.


lonvia:
I was planning to get rid of this with the [switch to CSV format](https://github.com/osm-search/Nominatim/compare/master...lonvia:Nominatim:new-wikimedia-tables).

PR #3112, opened 2023-07-17
fix polgyon polygon typo

jenkin:
I got this error while working on https://github.com/sparkfabrik/nominatim-openapi and I find two places with this little typo, both in error messages for the user.


lonvia:
Thank you!

PR #3139, opened 2023-08-04
Search.md - update XML and JSON example output

mtmail:
- Slash request format (http://domain/city/street) is no longer supported
- the request URL for KML request contained polygon=geojson
- the JSON examples didn't list the outer array
- JSON also returns `addresstype` and `place_rank` fields
- JSON no longer returns `icon` field

### DIFF
--- a/docs/api/Search.md
+++ b/docs/api/Search.md
@@ -174,116 +174,136 @@ This overrides the specified machine readable format. (Default: 0)
 ## Examples
 
 
-##### XML with kml polygon
+##### XML with KML polygon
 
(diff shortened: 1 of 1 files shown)


lonvia:
The XML format will go back to double quotes very soon. That's what Python produces.

mtmail:
I see double quotes now and will update the PR.

mtmail:
@lonvia Updated now. Note the `importance="9.999999994736442e-08"` string value in XML. I'm not sure if that's new behaviour.

PR #3158, opened 2023-08-19
Added a module to split Japanese words

miku0:
In these codes, Japanese addresses are divided into three categories based on administrative divisions: cities, municipalities, and below.
Nominatim uses ICU (International Components for Unicode) transliteration for user-entered addresses to split them into meaningful words. Here is an example of debugging. There are many candidates.
image
Fig. 1 The example of debugging.

To help make this division more accurate, when there are large administrative divisions (prefecture and city) in the string, we pre-separate them in the algorithm and put "," markers between the split words. 
This "," is set to BreakType.SOFT_PHRASE in the program and words with this node are penalized with a lower search priority.
The node relationship is as follows
(1)--da->(2)--ban->(3)--shi->(4)--da->(5)--ban->(6)
||                          　　　　^^                                  ||
|+------大阪市--------------+ +-------大阪--------+|
+-------------------大阪市大阪---------------------+

As a result of this change, "大阪市大阪" with SOFT_PHRASE is penalized more and given lower search priority than "大阪市", the name of a city (the fifth value from the left is the penalty value).
image
Fig. 2 Before the change.
image
Fig. 3 After the change.

### DIFF
--- a/nominatim/api/search/icu_tokenizer.py
+++ b/nominatim/api/search/icu_tokenizer.py
@@ -7,7 +7,7 @@
 """
 Implementation of query analysis for the ICU tokenizer.
 """
-from typing import Tuple, Dict, List, Optional, NamedTuple, Iterator, Any, cast
+from typing import Tuple, Dict, List, Optional, NamedTuple, Iterator, Any, cast, Set
 from copy import copy
 from collections import defaultdict
 import dataclasses
@@ -22,7 +22,7 @@ from nominatim.api.connection import SearchConnection
 from nominatim.api.logging import log
 from nominatim.api.search import query as qmod
 from nominatim.api.search.query_analyzer_factory import AbstractQueryAnalyzer
-
+from nominatim.api.search import icu_tokenizer_japanese
 
 DB_TO_TOKEN_TYPE = {
     'W': qmod.TokenType.WORD,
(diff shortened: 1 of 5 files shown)


lonvia:
The failing tests may not be related to you code. I have the same in an unrelated change. I'm investigating.

lonvia:
Can you please rebase your code on master? This should make the CI errors go away.

nominatim/api/search/icu_tokenizer.py
```diff
@@ -203,6 +207,25 @@ def normalize_text(self, text: str) -> str:
...
+        self, phrases: List[qmod.Phrase]
+    ) -> List[qmod.Phrase]:
+        """Normalize the phrases
+        """
+        normalized = list(filter(lambda p: p.text,
+                                 (qmod.Phrase(p.ptype, self.normalize_text(p.text))
+                                  for p in phrases)))
+        return normalized
+
+    def split_key_japanese_phrases(
```
lonvia, line 220: You probably forgot to delete the old code here.

nominatim/api/search/query.py
```diff
@@ -29,6 +29,7 @@ class BreakType(enum.Enum):
     """ Break created as a result of tokenization.
         This may happen in languages without spaces between words.
     """
+    SOFT_PHRASE = ':'
```
lonvia, line 32: Can you please add documentation for this new type, just like it is done in the lines above.

lonvia (commented):
Looks mostly good. Just two minor comments from my side.

nominatim/api/search/query.py
miku0: Thank you so much for your help and comments.
I added the documentation.

lonvia:
Superseded by #3629.

Issue #2719, opened 2022-05-22
Allow creating postgresql database ahead of time


Xiretza:
**Is your feature request related to a problem? Please describe.**
The current installation instructions require creating a postgresql superuser just for nominatim: https://nominatim.org/release-docs/latest/appendix/Install-on-Ubuntu-20/#setting-up-postgresql

While those permissions can probably be removed once the database is created, that step is easily forgotten, causing a possible security issue for all other users of the postgresql cluster in case the nominatim user is compromised.

Trying to create the database before running `nominatim import` immediately errors with `permission denied to create database`, even if the database already exists.

**Describe the solution you'd like**
The administrator should be able to set up nominatim's database structure without giving superuser permissions to nominatim.

lonvia:
There probably will be a non-supervisor import mode at some point but it's not exactly high on the priority list at the moment.

lonvia:
Here is how this could work:

* add a new import mode `nominatim import --prepare-database`, which executes all commands that need privileged mode (create database, create extensions, anything else?)
* add a new import node `nominatim import --no-superuser` which executes the import as usual except for the priviledged commands. This mode should still fail when it detects that Nominatim is already installed in the database (e.g. look if table `place` exists).

The new modes should be tested by changing the [CI import runs](https://github.com/osm-search/Nominatim/blob/5751686fdce33f6b69a93dd1641ceacb2239791f/.github/workflows/ci-tests.yml#L189) to run first `--prepare-database` and then `--no-superuser` import.

contiyo:
Hello. 

Is this being worked by someone ? 
It seems like a really good first issue and I would like to contribute.

lonvia:
Please go ahead.

lonvia:
This is now implemented on master since #3204.

PR #3204, opened 2023-09-19
Added --no-superuser flag

robbe-haesendonck:
Adding the --no-superuser flag to allow imports into already existing databases.
Refs: #2719

Note: This is currently still a draft, there is no validation of required plugins, ... yet.

### DIFF
--- a/nominatim/clicmd/setup.py
+++ b/nominatim/clicmd/setup.py
@@ -65,6 +65,8 @@ class SetupAll:
                            help='Continue import even when errors in SQL are present')
         group3.add_argument('--index-noanalyse', action='store_true',
                            help='Do not perform analyse operations during index (expert only)')
+        group3.add_argument('--no-superuser', action='store_true',
+                            help='Do not attempt to create the database')
 
 
     def run(self, args: NominatimArgs) -> int: # pylint: disable=too-many-statements
@@ -81,9 +83,10 @@ class SetupAll:
             if not files:
                 raise UsageError("No input files (use --osm-file).")
 
-            LOG.warning('Creating database')
-            database_import.setup_database_skeleton(args.config.get_libpq_dsn(),
-                                                    rouser=args.config.DATABASE_WEBUSER)
+            if not args.no_superuser:
+                LOG.warning('Creating database')
+                database_import.setup_database_skeleton(args.config.get_libpq_dsn(),
+                                                        rouser=args.config.DATABASE_WEBUSER)
 
             LOG.warning('Setting up country tables')
             country_info.setup_country_tables(args.config.get_libpq_dsn(),
--- a/nominatim/clicmd/args.py
+++ b/nominatim/clicmd/args.py
@@ -86,6 +86,7 @@ class NominatimArgs:
     offline: bool
     ignore_errors: bool
     index_noanalyse: bool
+    no_superuser: bool
 
     # Arguments to 'index'
     boundaries_only: bool


robbe-haesendonck:
Some TODO's left:

- [ ] Check whether there is data already in the database
If there is data in the database, we need to exit prematurely as importing 2 times might cause problems
- [ ] Add some form of validation to see whether all required plugins are installed/available in postgres.

robbe-haesendonck:
Normally I implemented everything as described in the linked ticket.
If I forgot anything please let me know!

Thanks in advance
~ Robbe

.github/workflows/ci-tests.yml
```diff
@@ -303,7 +303,9 @@ jobs:
               working-directory: /home/nominatim/nominatim-project
 
             - name: Import
-              run: nominatim import --osm-file ../test.pbf
+              run: |
+                  nominatim import --prepare-database --osm-file ../test.pbf
```
Xiretza, line 307: `--prepare-database` shouldn't need the OSM files.

nominatim/clicmd/setup.py
```diff
@@ -81,36 +85,46 @@ def run(self, args: NominatimArgs) -> int: # pylint: disable=too-many-statements
...
-                          'use Wikipedia importance data.')
-
-            LOG.warning('Importing secondary importance raster data')
-            if refresh.import_secondary_importance(args.config.get_libpq_dsn(),
-                                                   args.project_dir) != 0:
-                LOG.error('Secondary importance file not imported. '
-                          'Falling back to default ranking.')
-
-            self._setup_tables(args.config, args.reverse_only)
+            if args.no_superuser and args.prepare_database:
```
Xiretza, line 88: The logic here is quite confusing, maybe renaming the options to `--only-prepare-database` and `--only-import-data` would help? Alternatively, split the database preparation completely out of `import` so you first do `nominatim prepare-db`, then `nominatim import --osm-file ...`. This would also more easily allow passing different postgres connection parameters to the prepare step, doing only that part as the superuser is the whole point after all.

Xiretza (changes requested):
Thanks for working on this! I've yet to try this out, but the main use case for me would be running the preparation using the postgres superuser, then running the import as the unprivileged nominatim user. AFAICT this isn't possible yet without changing the config file inbetween, ideally the preparation step would accept different postgresql connection settings on the commandline.

nominatim/clicmd/setup.py
robbe-haesendonck: I agree, I'll start with implementing the different naming of the flags in favor of keeping backwards compatibility.

robbe-haesendonck:
@Xiretza Thanks for taking a look! I'll implement the changes as soon as possible. Splitting it up into different commands altogether will break compatibility I presume, so I'll first look at renaming the flags hoping that will clear the logic up (which indeed the logic for is kinda messy at the moment).

.github/workflows/ci-tests.yml
robbe-haesendonck: Added an extra check in the cli and changed the ci-tests.

lonvia:
I find the options still confusing, especially as they now in competition with `continue`. I would prefer:

```
nominatim import --prepare-database
nominatim import --continue import-from-file --osm-file planet.pbf
```

The first command can be prefixed with `NOMINATIM_DATABASE_DSN=whatever` to modify the user if necessary.

This should make much of the logic easier.

.github/workflows/ci-tests.yml
```diff
@@ -303,7 +303,9 @@ jobs:
               working-directory: /home/nominatim/nominatim-project
 
             - name: Import
-              run: nominatim import --osm-file ../test.pbf
+              run: |
+                  nominatim import --prepare-database
+                  nominatim import --no-superuser --osm-file ../test.pbf
```
lonvia, line 308: Instead of modifying the existing CI run, add another, where you set up a privileged and non-privileged user and really run the second command in non-privileged mode.

nominatim/tools/database_import.py
```diff
@@ -38,6 +40,20 @@ def _require_version(module: str, actual: Tuple[int, int], expected: Tuple[int,
...
+def check_existing_database_plugins(dsn: str):
+    """ Check that the database has the required plugins installed."""
+    with connect(dsn) as conn:
+        _require_version('PostgreSQL server',
+                         conn.server_version_tuple(),
+                         POSTGRESQL_REQUIRED_VERSION)
+        _require_version('PostGIS',
+                         conn.postgis_version_tuple(),
+                         POSTGIS_REQUIRED_VERSION)
+        _require_version('hstore',
```
lonvia, line 52: I don't think we require any specific minimal version of hstore. Simply check that the extension is loaded.

.github/workflows/ci-tests.yml
robbe-haesendonck: I looked at this yesterday but I'm noticing I'm a bit out of my depth in that regard. Can you give me a pointer to where the users are created? Once I have that it should be pretty easy.

robbe-haesendonck:
@lonvia I Implemented the options using continue, but I kept the legacy option also.
Do I remove the legacy import command altogether (as in the user needs to run the 2 commands)?

I wasn't sure so I opted to do it like this first.

nominatim/clicmd/setup.py
```diff
@@ -157,6 +173,11 @@ def run(self, args: NominatimArgs) -> int: # pylint: disable=too-many-statements
 
         return 0
 
+    def _is_complete_import(self, args: NominatimArgs) -> bool:
```
lonvia, line 188: You don't really need this extra function. Everywhere where it is used, a simple check for either `continue` or `prepare_database` will do.

nominatim/clicmd/setup.py
```diff
@@ -76,43 +80,66 @@ def run(self, args: NominatimArgs) -> int: # pylint: disable=too-many-statements
...
+                LOG.warning('Creating database')
+                database_import.setup_database_skeleton(args.config.get_libpq_dsn(),
+                                                        rouser=args.config.DATABASE_WEBUSER)
+
+                if not self._is_complete_import(args):
+                    return 0
+
+            if not args.prepare_database or \
+                    args.continue_at == 'import-from-file' or \
+                    self._is_complete_import(args):
```
lonvia, line 112: This looks incorrect. If I read the code right, then `args.prepare_database` must be `false` at this point because otherwise the return in line 108 would have triggered. That would mean the statement is always true.

nominatim/db/connection.py
```diff
@@ -174,6 +174,15 @@ def postgis_version_tuple(self) -> Tuple[int, int]:
 
         return (int(version_parts[0]), int(version_parts[1]))
 
+
+    def extension_loaded(self, extension_name: str) -> bool:
+        """ Return True if the hstore extension is loaded in the database.
+        """
+        with self.cursor() as cur:
+            cur.execute('SELECT extname FROM pg_extension WHERE extname = %s', (extension_name))
```
lonvia, line 182: Fun with Python tuples: needs to be `(extension_name, )`.

lonvia:
> Do I remove the legacy import command altogether (as in the user needs to run the 2 commands)?

No. The logic right now is what I had in mind. Makes it a reasonably simple change.

Regarding github actions, I think you are almost there. Simply work from a copy of the install action of these lines: https://github.com/osm-search/Nominatim/blob/master/.github/workflows/ci-tests.yml#L222-L291 instead of modifying the existing one. Once you add the code lines you already have, that should already be it. Maybe at the call to `check-database` for paranoia. Also, it is enough to run this in Ubuntu-22. You can delete `Ubuntu-20` from the matrix names.

robbe-haesendonck:
@lonvia So, the CI tests are failing. I fixed the create import user (-P option doesn't accept a string which was a wrong assumption on my end).

However there are also some fails on the linting, but I simply do not know how to reduce the branching any further. Do I disable that test? Could also probably create a function which does some of the sanity checks (first three if statements).

lonvia:
> However there are also some fails on the linting, but I simply do not know how to reduce the branching any further. Do I disable that test?

Yes, for this particular case you can simply add another exception in line 74 for the 'too-many-branches' check. That entire function will need some refactoring eventually but that is something for a different PR.

robbe-haesendonck:
I looked at the errors in the CI log, the failing postgres login is probably because it doesn't have a password set yet, which can be fixed easily.
The other issue is with the static type checker, but AFAIK I did not touch that part at all. I guess it should work when I change line 110 in `database_import.py` to make the 2 signatures match.

I've tested a non-superuser import on Nomad, and everything seems to work just fine. So I assume changing the signature could be ok?

I'll take a more in-depth look tomorrow, today I was preoccupied with some other work.

nominatim/clicmd/setup.py
```diff
@@ -76,43 +80,62 @@ def run(self, args: NominatimArgs) -> int: # pylint: disable=too-many-statements
...
-
-            self._setup_tables(args.config, args.reverse_only)
-
-        if args.continue_at is None or args.continue_at == 'load-data':
+            if args.prepare_database or args.continue_at is None:
+                LOG.warning('Creating database')
+                database_import.setup_database_skeleton(args.config.get_libpq_dsn(),
+                                                        rouser=args.config.DATABASE_WEBUSER)
+                if args.prepare_database:
+                    return 0
```
lonvia, line 106: This part in lines 101-106 should really be before line 96. As the code is written now, it gives the impression that database preparation happens as part of `--continue-from-file` but then it doesn't.

nominatim/clicmd/setup.py
```diff
@@ -76,43 +80,62 @@ def run(self, args: NominatimArgs) -> int: # pylint: disable=too-many-statements
...
+            if args.continue_at in ('import-from-file', None):
+                # Check if the correct plugins are installed
+                database_import.check_existing_database_plugins(args.config.get_libpq_dsn())
+                LOG.warning('Setting up country tables')
+                country_info.setup_country_tables(args.config.get_libpq_dsn(),
+                                                args.config.lib_dir.data,
+                                                args.no_partitions)
+
+                LOG.warning('Importing OSM data file')
+                database_import.import_osm_data(files,
```
lonvia, line 117: The type checker complains here because `files` can be None and it it is not able to determine that it cannot be None from the more complicated condition in line 98. Probably easiest to go back to the simple condition of `if not files` in line 98. We won't catch `--prepare-database` and `--osm-file` being used together but that doesn't do any harm.

robbe-haesendonck: After implementing the above comment, I ran the test locally and it seems it doesn't complain about it anymore.
I'll leave this open until I know from the CI env that it's actually resolved.

robbe-haesendonck:
@lonvia My apologies for the really slow implementation of the change requests.

robbe-haesendonck:
Static type checking seems to pass, only thing left is getting the import user correct in the CI tests.

.github/workflows/ci-tests.yml
```diff
@@ -349,3 +349,102 @@ jobs:
...
+                export DEBIAN_FRONTEND=noninteractive
+                apt-get update -qq
+                apt-get install -y git sudo wget
+                ln -snf /usr/share/zoneinfo/$CONTAINER_TIMEZONE /etc/localtime && echo $CONTAINER_TIMEZONE > /etc/timezone
+            shell: bash
+
+          - name: Setup database users
+            run: |
+                useradd -m nominatim
+                useradd -m osm-import
```
lonvia, line 388: You don't need this additional unix user. Simply run the no-superuser import as user 'nominatim' as well. Then the permission issues go away.

lonvia: Password authentication should work when you connect to 127.0.0.1 instead of the socket, .i.e. NOMINATIM_DATABASE_DSN="pgsql:dbname=nominatim;host=127.0.0.1;user=osm-import;password=osm-import"

.github/workflows/ci-tests.yml
```diff
@@ -349,3 +349,102 @@ jobs:
...
+
+          - name: Create import user
+            run: |
+                sudo -u postgres createuser -S osm-import
+                sudo -u postgres psql -c "ALTER USER \"osm-import\" WITH PASSWORD 'osm-import';"
+            working-directory: /home/nominatim/nominatim-project
+
+          - name: Grant import user rights
+            run: |
+                sudo -u postgres psql -c "GRANT INSERT, UPDATE ON ALL TABLES IN SCHEMA public TO \"osm-import\";"
```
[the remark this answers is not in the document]
lonvia: I'm thinking of a function similar to the '--owner' parameter of 'createdb'. The flag would specify "please prepare the database for the non-privileged user foo and make them owner".

.github/workflows/ci-tests.yml
robbe-haesendonck: That was the hint I was looking for!
Thanks!

.github/workflows/ci-tests.yml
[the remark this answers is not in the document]
robbe-haesendonck: I guess you would want this after the prepare database part is done, and if set, make the user defined by that flag the owner? I understand correctly right?

--database-owner does sound clear.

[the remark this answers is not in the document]
lonvia: It's more convenient to do it directly when creating the database.

Let's leave this for a follow-up.

lonvia:
Almost good to go. Could I kindly ask you to rebase your code on master and force push to the branch? If you prefer to squash your commits, you are welcome to do so as well.

lonvia:
The failing check should be fixed on master. I've manually merged this now doing a rebase. Thank you for your work and for patiently addressing all comments.

It would be good to have some documentation for this new feature in https://nominatim.org/release-docs/latest/admin/Advanced-Installations/. If you like doing a follow-up commit, it would mean adding a section in https://github.com/osm-search/Nominatim/blob/master/docs/admin/Advanced-Installations.md

PR #3226, opened 2023-10-14
Proposal for documentation extension

dbt-lucka:
This PR refers to further analysis I spent based on [this](https://github.com/osm-search/Nominatim/discussions/3225) discussion.

I have `pg_restore`'ed from a 64 GB machine to a 32 GB machine or so, and it went fine, seemingly. Nominatim works as expected and there is no reason to believe something is off. However, there was 1 subtle line in the PostgreSQL logs that I overread: 

```
ERROR:  temporary file size exceeds temp_file_limit (30852179kB)
```

This has most likely resulted in skipping index creation on `planet_osm_nodes` (containing 3.2B entries for Europe since I don't use a flatnode file) and thus, `nominatim replication` taking literally forever because of slow query performance on that table. It was very time consuming to find out the root cause of this and its so subtle that I found it useful to extend the documentation.

Adding the equal amount of memory to the target SQL instance at least during the restore process is essential.

### DIFF
--- a/docs/admin/Advanced-Installations.md
+++ b/docs/admin/Advanced-Installations.md
@@ -197,7 +197,10 @@ target machine.
     of a full database.
 
 Next install Nominatim on the target machine by following the standard installation
-instructions. Again, make sure to use the same version as the source machine.
+instructions. Again, make sure to use the same version as the source machine and ensure 
+the target machine has equal system requirements than the local source system. As 
+`pg_restore` is running actual SQL commands, you need identical hardware settings at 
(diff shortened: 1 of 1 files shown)


lonvia:
I'm not sure that this is a good advice to add. Importing on a large machine (to have a fast import) and then restoring on a smaller production machine is one of the main use-cases for moving the database.

`pg_restore` should have reported an error when creating an index. Given that errors are usually not fatal for pg_restore that tends to be overseen. So maybe the better advice here is: carefully inspect the output of pg_restore.

PR #3228, opened 2023-10-16
Properly validate postcodes with country code

### LINKED ISSUE #3227: Some valid postcodes with country code in the name do not pass validation
## What did you search for?

All postcodes from Isle of Man (IM) and Jersey (JE) do not pass postcode validation. As a result, a search for a valid postcode IM9 4LY returns nothing:
https://www.openstreetmap.org/search?query=IM9%204LY
even though the postcode exists in OSM data:
https://www.openstreetmap.org/way/396520136

The issue is caused by postcode normalisation function which strips the country code from the postcode. It does not work correctly when a valid postcode is prefixed by its country's country code. This can be fixed by removing the country code from the postcode pattern in country_settings.yaml config file, as is done for Anguilla, Andorra, Cayman Islands, Saint Lucia, Latvia, Moldova, Saint Vincent and British Virgin Islands. However, such approach makes the validation less accurate, as it lets the invalid postcodes (without the two letter prefix) be accepted as valid ones. Additionally, there are at least two countries in which postcodes can start with country code, but don't have to:
- Malta, postcodes in Mtarfa town start with MTF
- Brunei - some postcodes start with BN: https://postal-codes.cybo.com/brunei/
These cannot be treated by fixing the output format, though I can't tell if these cases currently exist in OSM data or not. 

Strangely, the postcode QA map does not show the affected postcodes for Isle of Man and Jersey, even though they must have been rejected from search index as illegal:
https://nominatim.org/qa/#map=9.43/54.17/-4.32&layer=postcode_bad_format

## What result did you get?

Nothing

Valid postcode

**When the result missing completely:**

I expect the postcode from these nodes to be searchable: 
https://www.openstreetmap.org/way/396520136
https://www.openstreetmap.org/way/493311642
https://www.openstreetmap.org/way/591317991

## Further details

I prepared a generic fix for this issue, by including the country regex in normalization regex. This way the country code is stripped when it is part of the postcode pattern only in case it is duplicated. (e.g. IM IM9 4LY)

pawel-wroniszewski:
Include postcode pattern in postcode normalisation regex, instead of removing it from postcode pattern in config.

It properly handles postcode validation and normalization when country code is part of the postcode, e.g. for Isle of Man, Jersey, Anguilla, Andorra, Cayman Islands and more.

Fixes #3227.

### DIFF
--- a/nominatim/data/postcode_format.py
+++ b/nominatim/data/postcode_format.py
@@ -25,7 +25,7 @@ class CountryPostcodeMatcher:
 
         pc_pattern = config['pattern'].replace('d', '[0-9]').replace('l', '[A-Z]')
 
-        self.norm_pattern = re.compile(f'\\s*(?:{country_code.upper()}[ -]?)?(.*)\\s*')
+        self.norm_pattern = re.compile(f'\\s*(?:{country_code.upper()}[ -]?)?({pc_pattern})\\s*')
         self.pattern = re.compile(pc_pattern)
 
         self.output = config.get('output', r'\g<0>')
--- a/settings/country_settings.yaml
+++ b/settings/country_settings.yaml
@@ -4,7 +4,8 @@ ad:
     languages: ca
     names: !include country-names/ad.yaml
     postcode:
-      pattern: "ADddd"
+      pattern: "(ddd)"
+      output: AD\1
 
 
 # United Arab Emirates (الإمارات العربية المتحدة)
@@ -38,7 +39,8 @@ ai:
     languages: en
     names: !include country-names/ai.yaml
     postcode:
-      pattern: "AI-2640"
+      pattern: "2640"
+      output: AI-2640
 
 
 # Albania (Shqipëria)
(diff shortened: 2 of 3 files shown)


lonvia (changes requested):
The fix in postcode_format.py is okay.

Getting rid of the optional country codes is not acceptable. Both forms are used in these countries and there is no reason to reject one form or the other.

pawel-wroniszewski:
@lonvia The optional country codes are still supported, there are tests for it - I actually added some more, like "je JE5 0LA" will now be accepted as a legal post code "JE5 0LA" which was not the case previously. Obviously the preferred format is "JE5 0LA", but it is just to show that I did not touch the existing functionality.

pawel-wroniszewski:
@lonvia I think I see your concern - you meant I shouldn't have made the country settings stricter. I briefly checked the postcode format usage in OSM data and noticed that indeed, in most of in the countries affected in my first commit, the usage was mixed (a lot of postcodes not following the official format). As an exception, Latvia seemed to have all of the postcodes valid, while Andorra and  Cayman Islands had most of them valid. Nevertheless I don't want to mix two things in the PR and I reverted country settings change. I aligned unit tests with the current state of things.

lonvia (approved):
Indeed, that's what I meant. No need to be overly strict with OSM data here.

PR #3283, opened 2023-12-29
Solve cast error when building nominatim with pg module

williancolognesitrimble:


mtmail:
@williancolognesitrimble Just to clarify, this is an error you saw with postgresql 16? We test against version s9.6, 13 and 15 currently. It's possible you're the first user who tried version 16.

williancolognesitrimble:
> @williancolognesitrimble Just to clarify, this is an error you saw with postgresql 16? We test against version s9.6, 13 and 15 currently. It's possible you're the first user who tried version 16.

Not sure if the problem is specific with postgres 16, the problem is that the postgres module was not compiling due to syntax error.

lonvia:
Confirmed that this is a new error for PostgreSQL 16, however the fix is the wrong one. Postgresql 16 requires a new include `"varatt.h"`.

PR #3320, opened 2024-01-27
Fix returned HTTP error when query runs too long

### LINKED ISSUE #3303: Python frontend takes long time to processes certain queries and returns 500 error to the user
## What did you search for?
9 k https://nominatim.openstreetmap.org/search.php?q=9+k&format=jsonv2
9 al https://nominatim.openstreetmap.org/search.php?q=9+al&format=jsonv2
6 al https://nominatim.openstreetmap.org/search.php?q=6+al&format=jsonv2
10 ju https://nominatim.openstreetmap.org/search.php?q=10+ju&format=jsonv2
10 ji https://nominatim.openstreetmap.org/search.php?q=10+ji&format=jsonv2
1 h https://nominatim.openstreetmap.org/search.php?q=1+h&format=jsonv2
1 al https://nominatim.openstreetmap.org/search.php?q=1+al&format=jsonv2
1 s https://nominatim.openstreetmap.org/search.php?q=1+s&format=jsonv2
12 r https://nominatim.openstreetmap.org/search.php?q=12+r&format=jsonv2
5 ba https://nominatim.openstreetmap.org/search.php?q=5+ba&format=jsonv2

## What result did you get?
500 Internal Server Error

A proper result or an empty list. If queries take too long, I'd expect them to be terminated and a non-complete result set be returned to the user, according to the NOMINATIM_REQUEST_TIMEOUT docs. 

## Further details

I think the 500 error is the QUERY_TIMEOUT error being propagated to the user. 
I was playing with Python frontend compiled from current master (bfc7acb) and I get the same. They work when I increase the QUERY_TIMEOUT. The queries can take up to a minute on my machine. Surprisingly, they take that long every time I run them - I'd expect them to be cached in Postgres once complete.

lonvia:
The code was only catching the generic TimeoutError. In Python <= 3.10 asyncio.TimeoutError is still something distinct, so catch them both.

Fixes #3303.

### DIFF
--- a/nominatim/server/falcon/server.py
+++ b/nominatim/server/falcon/server.py
@@ -10,6 +10,7 @@ Server implementation using the falcon webserver framework.
 from typing import Optional, Mapping, cast, Any, List
 from pathlib import Path
 import datetime as dt
+import asyncio
 
 from falcon.asgi import App, Request, Response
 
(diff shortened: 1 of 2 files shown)


MinhazAhmed24:
how do remove parent security and parent connenction

PR #3342, opened 2024-02-26
Correct some typos

mtmail:
As reported by https://github.com/codespell-project/codespell

It suggests `./nominatim/api/search/db_search_builder.py:8: Convertion ==> Conversion` but I wasn't sure if it should be `Converting` instead.

iD editor uses it in Github Actions with `only_warn` https://github.com/openstreetmap/iD/blob/develop/.github/workflows/codespell.yml


lonvia:
Thank you! I have a few pending PRs and will merge this manually afterwards. I will also fix the new linting complaint.

lonvia:
Surprisingly it was mergable without conflicts.

> iD editor uses it in Github Actions with only_warn

Looks interesting. Do you want to set that up for Nominatim?

mtmail:
> Do you want to set that up for Nominatim?

I'm in favor. I'll add it to my list unless you want to open a github issue to track it.

PR #3375, opened 2024-03-26
add missing space in taginfo listing

matkoniecz:
say https://taginfo.openstreetmap.org/tags/shop=guns#projects was listing description

`POI/feature in the search database(except for values: no).`

BTW - would it be useful to notify about retagging such as `shop=guns` to `shop=gun` before they happen (or after they happen?). It was suggested to me in https://www.openstreetmap.org/changeset/148979715 that doing it would be useful but I am not entirely sure about it.

Or would it be useful in cases where specific tag value is reported as supported, and not useful when key in general is supported?


lonvia:
The entry is for `shop=<anything>` . A change from 'gun' to 'guns' is not relevant.

Issue #3374, opened 2024-03-25
Does --reverse-only param still work for import process ?


felix-geovelo:
**Describe the bug**
I can't see any difference in the import process with and without --reverse-only. In both case I have the same logs output.

**To Reproduce**
I tried to import a small PBF file with and without the --reverse-only param to see the difference in the import logs. But there is apparently none.
I was expecting that with --reverse-only, all logs looking like "Rank ... ETA" would disappear but that was not the case (i.e that all indexing stage would be skipped)
Does this feature still works ? Is it supposed to save a lot of time during the import process for large PBF file ?

**Software Environment (please complete the following information):**
- Nominatim version: 4.4.0
- Postgresql version: 16
- Postgis version: 3.4.2
- OS: Ubuntu 22.04.4 LTS

**Hardware Configuration (please complete the following information):**
- RAM: 126 Go
- number of CPUs: 16
- type and size of disks: 2T of NVME

**Postgresql Configuration:**

RAS

**Nominatim Configuration:**

Variable set in .env :
NOMINATIM_DATABASE_DSN & NOMINATIM_DATABASE_WEBUSER

**More info:**

I'm trying to create a nominatim DB to import into photon after, so --reverse-only could help me to save some time during import process by skipping indexing stages

mtmail:
Indexing by rank is still needed for a reverse-only database. It calcuzwhich streets belong to which city, which city to which state etc. Only the search index creation is skipped. That’s maybe 10% of total import time.

felix-geovelo:
Thanks for the quick and clear reply, maybe it could be a useful add to the doc saying that --reverse-only save up to 10% of import time on behalf of 5% disk space.

Thanks again

PR #3385, opened 2024-04-02
Import documentation: reverse-only only a little faster import

mtmail:
As discussed in https://github.com/osm-search/Nominatim/issues/3374#issuecomment-2019548650

10% might even be optimistic. The 'post-process tables' step on a planet import can take as little as 1 hour only.


lonvia:
Yes, likely optimistic. Timewise you save writing to the search_name table and creating indexes for it. It's probably not noticable in the grand scheme of things. Should we put it like that?

mtmail:
Changed now to 'not significant faster' now

PR #3389, opened 2024-04-09
CLI: get valid --format values via autodiscover

mtmail:
When adding a new format to `nominatim/api/v1/format.py` one also needed to add the format into the CLI. Now the format gets found automatically:

Before
```
nominatim search -h
nominatim reverse -h
nominatim lookup -h
  --format {xml,json,jsonv2,geojson,geocodejson,debug,myformat}
```

Now
```
  --format {xml,geojson,geocodejson,json,jsonv2,myformat,debug}
```


lonvia:
Nice!

Issue #2785, opened 2022-07-31
Reverse geocoding does not return amenities, only adresses


Juanjo-MC:
## What did you search for?

https://nominatim.openstreetmap.org/reverse?lat=50.8973333&lon=4.3578048

## What result did you get?

The address of those coordinates

## What result did you expect?

I expected to see, in addition to the address, the name and info of the amenity located there (B Event caterer), see here: https://nominatim.openstreetmap.org/ui/search.html?q=B+Event
It seems to me that it worked that way before.

## Further details

May it be related to #2683, which is now solved?

lonvia:
The house (with only the address) and the craft node are exactly the same distance from the lookup point. Which one of the two is returned is pretty much random. As soon as you are slightly away from the location of the craft node, the building polygon will always be closer.

Juanjo-MC:
> The house (with only the address) and the craft node are exactly the same distance from the lookup point.

Not sure I understand. 
If you look at the reverse geocoding request I made, I passed to it the exact same coordinates than the forward one gave me for that POI. In that case the distance between the craft POI and the lookup point should be zero, no?

lonvia:
> In that case the distance between the craft POI and the lookup point should be zero, no?

Yes it is. But so is the distance between the lookup point and https://www.openstreetmap.org/way/252972065. They are the same, namely 0. Thus whatever PostgreSQL decides to return first wins. We only order by distance and there is nothing in the query that would help decide which result should be preferred.

lonvia:
I've given this some more thought and can't come up with any particular reason to chose one order over another. So lets approach this from another direction: @Juanjo-MC could you explain, why exactly it is a problem for you that the house is returned instead of the caterer? What were you trying to achieve?

Juanjo-MC:
>  What were you trying to achieve?

Nothing in particular

> could you explain, why exactly it is a problem for you that the house is returned instead of the caterer? 

Well, it is not really a problem. I seem to remember that, in previous versions, if there was an amenity at the same address Nominatim was returning it. But it might be that I am wrong.

lonvia:
We might have ordered by distance to the center of the building previously instead of the full building outline. But doing that comes with its own issues. One could argue that a node or a way should be preferred over an area when the distance is equal but that would only solve the case, where the location hits the amenity node exactly. Not too good either.

[...]

PR #3390, opened 2024-04-11
Handle POIs in POIs during reverse lookup

lonvia:
The closest object for reverse lookups is determined via the distance between query point and object. This automatically makes area POIs win when the lookup point is inside because then the distance is 0.

This PR tries to somewhat improve the situation by looking for the two closest objects and returning the second if the first is an area and the second is a POI node that is reasonably close.

There are some corner cases that this hack doesn't cover, for example when another street happens to be closer than the point POI. These cases are anyways tricky and probably better solved by returning multiple results. The generic case of a large POI area like a shopping mall should work better with this change.

See also #2785 and #3387.


mtmail:
I can confirm this fixes the issue I saw in https://github.com/osm-search/Nominatim/discussions/3387

Issue #3382, opened 2024-04-01
geocodejson format can contain duplicate admin level keys


mtmail:
https://nominatim.openstreetmap.org/reverse?accept-language=en&addressdetails=1&format=geocodejson&lat=41.00488&lon=28.98016

returns level4 twice

```
  "label": "Agora Guesthouse, 10, Amiral Tafdil Sokağı, Cankurtaran Mahallesi, Istanbul, Fatih, Istanbul, Marmara Region, 34122, Turkey",
  "admin": {
    "level8": "Cankurtaran Mahallesi",
    "level4": "Istanbul",
    "level6": "Fatih",
    "level4": "Istanbul",
    "level3": "Marmara Region"
  }
```

It's not obvious because tools like JSON browser plugin, [jq](https://jqlang.github.io/jq/) or some parsers ignore the second occurance silently. In fact duplicate keys seem to be allowed https://stackoverflow.com/questions/21832701/does-json-syntax-allow-duplicate-keys-in-an-object . The parser I'm using fails when it encounters a duplicate key.

I'll provide a PR for the output formatting but I wonder if the issue is deeper or if it's possible the values for the same level could also be different.

lonvia:
Place nodes with `admin_level` again. They tend to be set for capitals and then refer to admin level of the administration they are the capital for. The geocodejson formatter should ignore everything that isn't a `boundary=administrative`.

PR #3400, opened 2024-04-23
geocodejson output: make sure a level is only printed once

mtmail:
Another fix to https://github.com/osm-search/Nominatim/issues/3382

Example:
`curl 'https://nominatim.osm.org/reverse?lat=51.01762378&lon=-3.1520338&format=geocodejson'` prints
`[...]"admin":{"level10":"Bishop's Hull","level10":"Bradford-on-Tone","level6":"Som[...]`

Browsers and Nominatim CLI handle duplicate keys by selecting the last occurrence ("Bradford-on-Tone") while the fix I added selects the first ("Bishop's Hull"). 

https://nominatim.openstreetmap.org/ui/details.html?osmtype=W&osmid=6011967
image

### DIFF
--- a/nominatim/api/v1/format_json.py
+++ b/nominatim/api/v1/format_json.py
@@ -246,10 +246,13 @@ def format_base_geocodejson(results: Union[napi.ReverseResults, napi.SearchResul
 
             out.key('admin').start_object()
             if result.address_rows:
+                seen = {}
                 for line in result.address_rows:
                     if line.isaddress and (line.admin_level or 15) < 15 and line.local_name \
-                       and line.category[0] == 'boundary' and line.category[1] == 'administrative':
+                      and line.category[0] == 'boundary' and line.category[1] == 'administrative' \
+                      and line.admin_level not in seen:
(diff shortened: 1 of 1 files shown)


lonvia:
I don't think this is about first/last occurrence. We have a street here that goes through two admin_level=10 areas. You'd probably want to select one were the majority of the road is. Or for a house number, you'd want to select the one, it is actually in. It would be relatively expensive to do such a computation on admin_level level.

While I see that there is potentially a problem here, I'd file this particular example under data issue. Bishop's Hull is mapped as a suburb and even tagged with `is_in=Tauton` but the [administrative boundary for Tauton](https://www.openstreetmap.org/relation/15836117) disagrees and excludes Bishop's Hull. The result is that Bishop's Hull is a suburb without a city to be in. Upgrade Bishop's Hull to village and all is well.

mtmail:
Thanks for pointers. Fixing the OSM data worked.

I saw a lot more places where this is happening but I'll keep the code fix in my local installation. We can review when somebody else reports the same issue. Probably my JSON parser (Perl) isn't modern enough.

PR #3414, opened 2024-05-07
Add USPS Standard Suffix Abbreviation

mhsrn21:
Added USPS's Standard Suffix Abbreviation for postal addressing ([https://pe.usps.com/text/pub28/28apc_002.htm](https://pe.usps.com/text/pub28/28apc_002.htm))


mhsr21:
Some of the duplicates were present before my contribution--should I remove them altogether?

mtmail:
@mhsr21 You're right, there's already 42 duplicates. I will cleanup the others later.

```bash
cat settings/icu-rules/variants-* | perl -ne '/^\s+-\s+(.+?)\s+->\s+(.+)/ && $1 eq $2 && print' | wc -l
      42
```

mtmail:
@mhsr21 Would be great if you can remove the other duplicates, too. I see 42, and 41 of those are in the `variants-en.yaml` file.

```bash
cat settings/icu-rules/variants-* | perl -ne '/^\s+-\s+(.+?)\s+->\s+(.+)/ && $1 eq $2 && print' | wc -l
      42
```

mhsr21:
@mtmail Got rid of all duplicates in `variants-en.yaml`, plus the other one in `variants-fr.yaml`

settings/icu-rules/variants-en.yaml
```diff
@@ -344,25 +409,29 @@
...
     -  Slope -> Slpe
     -  Sound -> Snd
     -  South -> S
     -  South -> Sth
     -  Southeast -> SE
     -  Southwest -> SW
-    -  Spur -> Spur
+    -  Spring -> Spg
+    -  Springs -> Spgs
+    -  SPurs -> Spur
```
lonvia, line 432: Minor typo here.

lonvia (commented):
Thanks for going through this. I agree that we should have official abbreviations in this list.

On a more general note, the US abbreviation list has always had the problem that it is far too long. In particular, it has the problem that it proposes sometimes 3 or 4 variants for the same word. This has a negative effect on the size of the index. Would it make sense to restrict ourselves to the official abbreviations only or are the other ones are just as frequently used?

mhsr21:
> Thanks for going through this. I agree that we should have official abbreviations in this list.
> 
> On a more general note, the US abbreviation list has always had the problem that it is far too long. In particular, it has the problem that it proposes sometimes 3 or 4 variants for the same word. This has a negative effect on the size of the index. Would it make sense to restrict ourselves to the official abbreviations only or are the other ones are just as frequently used?

I only added the official abbreviations (the rightmost column).

lonvia:
Merged as part of #3652.

PR #3422, opened 2024-05-14
Drop address interpolations when no parent can be found

### LINKED ISSUE #3421: Interpolated address doesn't return city, state
## What did you search for?

Somewhat remote location in Canada

https://nominatim.openstreetmap.org/ui/reverse.html?lat=48.66510&lon=-71.22887&zoom=18

## What result did you get?

https://nominatim.openstreetmap.org/ui/details.html?osmtype=W&osmid=143579633&class=place

```json
  "category": "place",
  "type": "house",
  "addresstype": "place",
  "display_name": "3, Canada",
  "address": {
    "house_number": "3",
    "country": "Canada",
    "country_code": "ca"
  },
```

A street name, city, region, state. A nearby search on the same street returns house number 1
https://nominatim.openstreetmap.org/ui/reverse.html?lat=48.66611&lon=-71.22907&zoom=18

and contains

```json
  "display_name": "1, Chemin du Lac-à-l'Ours, Saint-David-de-Falardeau, Le Fjord-du-Saguenay, Saguenay–Lac-Saint-Jean, Quebec, Canada",
  "address": {
    "house_number": "1",
    "road": "Chemin du Lac-à-l'Ours",
    "village": "Saint-David-de-Falardeau",
    "county": "Le Fjord-du-Saguenay",
    "region": "Saguenay–Lac-Saint-Jean",
    "state": "Quebec",
    "ISO3166-2-lvl4": "CA-QC",
    "country": "Canada",
    "country_code": "ca"
  },
```

lonvia:
When an address interpolation has no parent, then results within these interpolations are displayed as "<housenumber>, <country>". That is not very helpful. Address interpolations only look for streets for parenting. We could extend this to parent to the next relevant administrative area but this would change the display name only to something like "<housenumber>, <city>, <country>". This is better but still confusing.

There are currently about 500 interpolations without parent (of about 3million in total). That's a very small number and points rather to data issues that should be fixed. On the Nominatim side, just drop those to avoid odd results during reverse geocoding.

Fixes #3421.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -13,6 +13,6 @@ ignored-classes=NominatimArgs,closing
 # 'too-many-ancestors' is triggered already by deriving from UserDict
 # 'not-context-manager' disabled because it causes false positives once
 #   typed Python is enabled. See also https://github.com/PyCQA/pylint/issues/5273
-disable=too-few-public-methods,duplicate-code,too-many-ancestors,bad-option-value,no-self-use,not-context-manager,use-dict-literal,chained-comparison,attribute-defined-outside-init,too-many-boolean-expressions
+disable=too-few-public-methods,duplicate-code,too-many-ancestors,bad-option-value,no-self-use,not-context-manager,use-dict-literal,chained-comparison,attribute-defined-outside-init,too-many-boolean-expressions,contextmanager-generator-missing-cleanup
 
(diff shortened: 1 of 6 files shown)


mtmail:
> There are currently about 500 interpolations without parent (of about 3million in total)

I fixed many, should be less than 150 now.

PR #3423, opened 2024-05-14
Update Output place_id documentation to match implementation

mudnug:
The current documentation represents that place_id is output as a string, but in fact the implementation provides it as a number


lonvia:
We have the same issue with "osm_id".

PR #3424, opened 2024-05-15
Implement a lighter format for wikipedia importance tables

lonvia:
Adds support for the new simpler CSV format for wikipedia importance values. This also comes with a much simplified table structure: redirects and articles are now in the same table and all unnecessary information has been dropped leaving only wikipedia article, wikidata ID and importance.

Support for the old-style wikipedia importance dumps remains in place for now. There will be official CSV dumps once we have removed the last obstacles in the generation process in https://github.com/osm-search/wikipedia-wikidata.

### DIFF
--- a/lib-sql/functions/importance.sql
+++ b/lib-sql/functions/importance.sql
@@ -65,7 +65,7 @@ BEGIN
   RETURN NULL;
 END;
 $$
-LANGUAGE plpgsql IMMUTABLE STRICT;
+LANGUAGE plpgsql IMMUTABLE;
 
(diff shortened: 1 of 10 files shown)


mtmail:
I think `nominatim/tools/check_database.py` needs to check for the new database table, too.

lonvia:
The [latest commit](https://github.com/osm-search/Nominatim/pull/3424) should be implementing that. Or do you have a different place in mind?

mtmail:
ah I see it now

PR #3466, opened 2024-07-08
Vagrant: add config for Apple Silicon

mtmail:
Apple Silicon systems (arm64) require a different image. Virtualbox (emulation) doesn't work, but Parallels (virtualization) does.


lonvia:
I just hope this doesn't create the false impression that we officially support MacOS.

PR #3472, opened 2024-07-09
Documentation: python frontend doesnt need website directory

mtmail:
The `Install-on-Ubuntu-*.sh` scripts setup a Nominatim python frontend. The python frontend doesn't create a `website` directory. The nginx configuration works fine either way but it's confusing to users (e.g. https://github.com/osm-search/Nominatim/discussions/3471)

### DIFF
--- a/vagrant/Install-on-Ubuntu-22.sh
+++ b/vagrant/Install-on-Ubuntu-22.sh
@@ -271,7 +271,7 @@ server {
     listen 80 default_server;
     listen [::]:80 default_server;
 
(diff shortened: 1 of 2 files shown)


lonvia:
Setting root to the project directory isn't a good idea. There might be files in this directory you don't want accidentally exposed to the internet. Just use nginx' default on Ubuntu `/var/www/html`?

mtmail:
replaced by https://github.com/osm-search/Nominatim/pull/3490

PR #3479, opened 2024-07-16
fix docs link in update interval error message

shumailxyz:
This PR updates the link for the docs in the error message that is throw for a low update interval.


mtmail:
thank you

PR #3490, opened 2024-07-29
After install point user to '/status' not '/' path

mtmail:
At https://nominatim.org/release-docs/latest/admin/Deployment-Python/ (very last line) we already point users to the status endpoint because there is no default redirect if no endpoint is selected. Do the same in the Ubuntu install instructions. Avoids confusion like https://github.com/osm-search/Nominatim/discussions/3471#discussioncomment-10000220 had.

### DIFF
--- a/vagrant/Install-on-Ubuntu-22.sh
+++ b/vagrant/Install-on-Ubuntu-22.sh
@@ -252,7 +252,8 @@ else                                  #DOCS:
     sudo systemctl restart apache2
 fi                                    #DOCS:
 
-# The Nominatim API is now available at `http://localhost/nominatim/`.
+# The Nominatim API is now available at `http://localhost/nominatim/`. Point your browser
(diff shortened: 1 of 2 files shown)


mtmail:
Good idea, changed now.

vagrant/Install-on-Ubuntu-22.sh
```diff
@@ -248,7 +248,8 @@ else                                  #DOCS:
     sudo systemctl restart apache2
 fi                                    #DOCS:
 
-# The Nominatim API is now available at `http://localhost/nominatim/`.
+# The Nominatim API is now available at `http://localhost/nominatim/`. To test if everything
+# works, point your browser to the status page at http://localhost/nominatim/status."
```
lonvia, line 252: You copied a '"' too much.

And let's get rid of the final full stop as well. I can already see the bug reports complaining about the 404 on `status.`. Or should we simply make it a markdown link to avoid copy mistakes?

PR #3523, opened 2024-08-23
Import-Styles.md - clarify country_code key

mtmail:
I'm also not sure about the example

``` lua
    local flex = require('import-full')

    function osm2pgsql.process_relation(object)
        if object.tags.boundary ~= 'administrative' or object.tags.admin_level ~= '2' then
          flex.process_relation(object)
        end
    end
```

When I added a debug statement to the function it was never called even though the input had relations. Should it be `flex.process_relation`?


lonvia:
Now, how do I convert this into an issue?

lonvia:
Retested the example replacing `process_relation` and it works as expected for me.

PR #2854, opened 2022-10-31
Expand Russian abbreviation list

shoorick:
There is a lot of abbeviations in Russian maps — in old paper maps and in OpenStreetMap too.
Previous version of list based on https://wiki.openstreetmap.org/wiki/Name_finder:Abbreviations#.D0.A0.D1.83.D1.81.D1.81.D0.BA.D0.B8.D0.B9_-_Russian contains items belongs mostly to roads, but there are many other shortened items on the maps.

PR #3555, opened 2024-10-06
Fixed Russian abbreviation list 

IvanShift:


IvanShift:
Hello!

I fixed the problem with https://github.com/osm-search/Nominatim/pull/2854
by changing the order of  "школа -> шк"  and adding the missing  "дом -> д"

PR #3592, opened 2024-11-18
Remove cmake building and bundled osm2pgsql

lonvia:
This leaves the pip installation as the only installation method.

Ubuntu 22.04 doesn't have a osm2pgsql that is recent enough, so that still needs to be compiled from source. Otherwise the action scripts become quite a bit simpler.


mtmail:
`Vagrantfile` and `Contributing` also mention `git clone --recursive`. That might no longer be needed.

PR #3615, opened 2024-12-16
Overhaul osm2pgsql style

### LINKED ISSUE #3365: Support for key "lock_name"
### Problem
Locks are tagged as regular waterways with an additional tag `lock=yes`. This means that their `name` tag is often simply the name of the river or canal, while the name of the lock itself is tagged in `lock_name`, as recommended in [Key:lock](https://wiki.openstreetmap.org/wiki/Key:lock). Nominatim doesn't see this tag, so these locks cannot be searched for by name. 

### Solution
Add the key `lock_name` to the list of tags that are considered names.

### Alternatives
Couldn't think of any. Hopefully this is a straightforward solution.

### Additional context
Searching for the name of a lock is useful not just for boaters, but also for regular folks who live in cities that have many of them, where they are often gathering places. The name of locks are also mentioned in the news and blogs posts, for instance the [Weerdsluis](https://www.openstreetmap.org/way/107194897) which is often talked about internationally due to its 'fish doorbell'.

### LINKED ISSUE #3603: Unexpected objects in search results for a building address
### Discussed in https://github.com/osm-search/Nominatim/discussions/3602

Searching for some building addresses with `layer=ADDRESS` in a DB imported with `NOMINATIM_IMPORT_STYLE=address` yields, on top of the building at the address, other unexpected POI-like objects.

Examples:
N5266350421, N5266417321 for [address](https://nominatim.openstreetmap.org/ui/search.html?q=%D0%9C%D0%BE%D1%81%D0%BA%D0%B2%D0%B0+%2C++%D0%BF%D1%80%D0%BE%D1%81%D0%BF%D0%B5%D0%BA%D1%82+%D0%97%D0%B0%D1%89%D0%B8%D1%82%D0%BD%D0%B8%D0%BA%D0%BE%D0%B2+%D0%9C%D0%BE%D1%81%D0%BA%D0%B2%D1%8B+%2C++1) (now removed).
N4682629046 for [address](https://nominatim.openstreetmap.org/ui/search.html?q=%D0%9C%D0%BE%D1%81%D0%BA%D0%B2%D0%B0%2C++%D0%90%D0%BA%D0%B0%D0%B4%D0%B5%D0%BC%D0%B8%D0%BA%D0%B0+%D0%90%D0%BD%D0%BE%D1%85%D0%B8%D0%BD%D0%B0+%2C+12%D0%BA3&layer=address).

The expected behavior is for POIs to be absent in a DB imported with `NOMINATIM_IMPORT_STYLE=address`, or if they has to be present, not be returned alongside the exact address match when there is one, or if they has to be returned, to be practically distinguishable from the exact address match.

lonvia:
This started out as a simple task to make the Nominatim osm2pgsql import style compatible with [osm2pgsql-themepark](https://osm2pgsql.org/themepark/) and ended up to be a complete refresh of the style implementation.

The philosophy of the configuration script stays largely the same but there are some significant changes in the implementation:

* The actual tag configuration, which was previously directly in the style files, has been moved to _preset_ snippets. The style files just apply the appropriate presets. This removes a lot of configuration duplication and makes it easier for you to reuse parts of the default configuration in your own scripts.
* The functions for setting the various tag types now have companion functions to modify an already existing configuration. This was necessary to make the preset snippets work but also means that you can now base your own style on an existing one and just modify the bits you don't like. You don't need to manual keep your configuration up to date with the default one anymore.
* The prefilter settings (delete tags or move them to extratags) have been merged with the main tag setting. These two settings have always been mutually exclusive. In fact, prefiltering was implicitly used to exclude some values for a certain key. The new `delete` and `extra` main tag settings, make the exclusion more explicit.
* The [main tag types](https://nominatim.org/release-docs/4.5/customize/Import-Styles/#set_main_tags-principal-tags) now translate to functions which do the final formatting/filtering of an entry before it is added to the database. That is about 20% faster than the nested if conditions we had before. It also has the added bonus that it is now possible to further customize at this point and do much more complex checks and transformations.

The changes are mostly backwards compatible, i.e. your custom styles should still work as before. The only exception is customization of the `process_tags()` function. This function is no longer considered public and neither are the helper functions used in it. They currently still work but will be removed at some point.

Finally, the actual goal, running under themepark, is now possible as well. Nominatim will export the appropriate variables to ensure that the theme is found. Using themepark it becomes easy to run Nominatim alongside other projects like openstreetmap-carto in the same database. I will update the cookbook after the next release.

This change also includes some improvements to the style which are now possible because of the new structure:
* exclude `highway=footway` when they are sidewalks and crossings
* use `information=*` for the class/type for `tourism=information` objects, when available
* use `water=*` for the class/type for `natural=water` when available and exclude water areas for rivers, streams and the like where there is already a waterway
* support for `lock_name` (closes #3365)
* do not import names when falling back to importing the address only (closes #3603)
* exclude unnamed swimming pools (most of those are private)


lonvia:
CI failures are due to OSM outage.

PR #3617, opened 2024-12-17
Slight wording changes for Import-Styles.md

mtmail:
Given the huge PR and number of code and configuration changes, these wording suggestions are minuscule. Can't wait to test the new logic and adapt my local custom filters.

### DIFF
--- a/docs/customize/Import-Styles.md
+++ b/docs/customize/Import-Styles.md
@@ -60,7 +60,7 @@ The _main tags_ classify what kind of place the OSM object represents. One
 OSM object can have more than one main tag. In such case one database entry
 is created for each main tag. _Name tags_ represent searchable names of the
 place. _Address tags_ are used to compute the address hierarchy of the place.
-Address are used for searching and for creating a display name of the place.
+Address tags are used for searching and for creating a display name of the place.
 _Extra tags_ are any tags that are not directly related to search but
 contain interesting additional information.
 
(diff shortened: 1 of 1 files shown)


docs/customize/Import-Styles.md
```diff
@@ -1,4 +1,4 @@
-# Configuring the Import of OSM data
+\# Configuring the Import of OSM data
```
lonvia, line 1: What does that do?

docs/customize/Import-Styles.md
```diff
@@ -23,7 +23,7 @@ customize it.
 
 The core of Nominatim's flex import configuration is the `flex-base` module.
 It defines the table layout used by Nominatim and provides standard
-implementations for the import callbacks that help with customizing
+implementations for the import callbacks that help with customising
```
lonvia, line 26: I was trying to settle on American spelling on the z vs. s front. In line with variable naming in the code.

docs/customize/Import-Styles.md
mtmail: It shows how easy it is to hit the \ key by accident. (removed now)

docs/customize/Import-Styles.md
mtmail: Ah, ok, I saw "recognise" and "initalise" later in the document and thought it was vice-versa.

PR #3648, opened 2025-02-16
Enable output of extratags for geocodejson format

lonvia:
This add support for the `extratags` parameter for the geocodejson format. The data is available in the additional `extra` field. This follows the standard already [set by Photon](https://github.com/komoot/photon/pull/576) a while ago.


lonvia:
This is now live on nominatim.osm.org @vkrause

vkrause:
> This is now live on nominatim.osm.org @vkrause

 Thank you, this is now in use in Itinerary :)

PR #3666, opened 2025-03-05
Replace custom Almost with stdlib math.isclose

eumiro:
`math.isclose` has per default relative tolerance of 10e-9, so if it matters, I can change it to the absolute tolerance of 0.00001 as it was used in the original code.


lonvia:
I remember that there were cases, where a 10e-9 tolerance was to small. However, as long as the tests are passing, lets just leave it.

Thanks.

PR #3667, opened 2025-03-05
Simplify  int/float manipulation

eumiro:
Some independent code simplifications. If needed, I can split them into individual commits.


lonvia:
Looks also good minus that small complaint from flake.

lonvia:
Thank you!

PR #3675, opened 2025-03-13
Add generic preprocessors

TuringVerified:
For now, I've added rules to filter out IPv4, IPv6 address & URLs starting with http/https in the ICU file. Most of regex_replace.py is the same as split_japanese_phrases.py.

I haven't added any tests yet. I was working using the code for the preprocessor test suite and when using QueryConfig inside a test file, the ICU-related settings (such as replacements) do not seem to be loaded. Is that supposed to happen? Are there any functions that can allow me to load the ICU file? Or is this something I'd have to make myself.

### DIFF
--- a/docs/customize/Tokenizers.md
+++ b/docs/customize/Tokenizers.md
@@ -67,8 +67,9 @@ Here is an example configuration file:
 
 ``` yaml
 query-preprocessing:
+    - step: split_japanese_phrases
     - step: regex_replace
-    replacements:
+      replacements:
         - pattern: https?://[^\s]* # Filter URLs starting with http or https
           replace: ''
     - step: normalize
@@ -110,6 +111,8 @@ The following is a list of preprocessors that are shipped with Nominatim.
         members: False
         heading_level: 6
         docstring_section_style: spacy
+
+##### regex-replace
 
 ::: nominatim_api.query_preprocessing.regex_replace
     options:
--- a/docs/customize/Tokenizers.md
+++ b/docs/customize/Tokenizers.md
@@ -67,7 +67,12 @@ Here is an example configuration file:
 
 ``` yaml
 query-preprocessing:
-    - normalize
+    - step: regex_replace
+    replacements:
+        - pattern: https?://[^\s]* # Filter URLs starting with http or https
+          replace: ''
+    - step: normalize
+
 normalization:
     - ":: lower ()"
     - "ß > 'ss'" # German szet is unambiguously equal to double ss
@@ -88,8 +93,8 @@ token-analysis:
             replacements: ['ä', 'ae']
 ```
 
-The configuration file contains four sections:
-`normalization`, `transliteration`, `sanitizers` and `token-analysis`.
+The configuration file contains five sections:
+`query-preprocessing`, `normalization`, `transliteration`, `sanitizers` and `token-analysis`.
 
 #### Query preprocessing
 
@@ -105,6 +110,17 @@ The following is a list of preprocessors that are shipped with Nominatim.
         members: False
         heading_level: 6
         docstring_section_style: spacy
+
+::: nominatim_api.query_preprocessing.regex_replace
+    options:
+        members: False
+        heading_level: 6
+        docstring_section_style: spacy
+    description: 
+        This option runs any given regex pattern on the input and replaces values accordingly
+    replacements:
+        - pattern: regex pattern
+          replace: string to replace with
 
 
 #### Normalization and Transliteration
--- a/settings/icu_tokenizer.yaml
+++ b/settings/icu_tokenizer.yaml
@@ -1,13 +1,5 @@
 query-preprocessing:
     - step: split_japanese_phrases
-    - step: regex_replace
-      replacements:
-        - pattern: \b(?:25[0-5]|2[0-4][0-9]|1?[0-9]{1,2})(?:\.(?:25[0-5]|2[0-4][0-9]|1?[0-9]{1,2})){3}\b # Filter for IPv4 addresses
-          replace: ''
-        - pattern: \b(?:(?:[A-Fa-f0-9]{1,4}:){1,7}|:)(?:[A-Fa-f0-9]{1,4})?\b # Filter for IPv6 addresses
-          replace: ''
-        - pattern: https?://[^\s]* # Filter URLs starting with http or https
-          replace: ''
     - step: normalize
 normalization:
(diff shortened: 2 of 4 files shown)


otbutz (changes requested):

settings/icu_tokenizer.yaml
```diff
@@ -1,5 +1,13 @@
 query-preprocessing:
     - step: split_japanese_phrases
+    - step: regex_replace
```
lonvia, line 3: As said in the issue, please don't add anything here in the default config. You can use the IPv4/6 example in the tests and in the [documentation](https://github.com/osm-search/Nominatim/blob/master/docs/customize/Tokenizers.md#query-preprocessing) that still needs to be written.

src/nominatim_api/query_preprocessing/regex_replace.py
```diff
@@ -0,0 +1,46 @@
...
+        match_patterns = self.config.get('replacements', 'Key not found')
+        for item in match_patterns:
+            phrase.text = re.sub(item['pattern'], item['replace'], phrase.text)
+
+        return phrase
+
+    def __call__(self, phrases: List[Phrase]) -> List[Phrase]:
+        """Apply regex replacements to the given addresses.
+        """
+        return [self.split_phrase(p) for p in phrases]
```
lonvia, line 40: This is not yet implementing the requirement that phrases are dropped completely when they are empty.

src/nominatim_api/query_preprocessing/regex_replace.py
```diff
@@ -0,0 +1,46 @@
...
+        """
+        This function performs replacements on the given text using regex patterns.
+        """
+
+        if phrase.text is None:
+            return phrase
+
+        match_patterns = self.config.get('replacements', 'Key not found')
+        for item in match_patterns:
+            phrase.text = re.sub(item['pattern'], item['replace'], phrase.text)
```
lonvia, line 33: This can be sped up by pre-compiling the regular expression.

lonvia (commented):
Regarding the tests: these are supposed to work independently from the default configuration. Simply initialize the QueryConfig with a dictionary containing the configuration (e.g. `{'step': 'regex_replace', 'replacements', ...}`)

When writing the tests, do not only think about successful use cases but also about possible errors. Please also test that the different variants of backreferences work as expected. See the [Python documentation](https://docs.python.org/3/library/re.html#re.sub) for more info.

TuringVerified:
I implemented the requested changes

settings/icu_tokenizer.yaml
lonvia: The code is still here.

docs/customize/Tokenizers.md
```diff
@@ -106,6 +111,17 @@ The following is a list of preprocessors that are shipped with Nominatim.
...
+    options:
+        members: False
+        heading_level: 6
+        docstring_section_style: spacy
+    description: 
+        This option runs any given regex pattern on the input and replaces values accordingly
+    replacements:
+        - pattern: regex pattern
+          replace: string to replace with
+
```
lonvia, line 124: Have you compiled the documentation and looked at the result of your changes?

test/python/api/query_processing/test_regex_replace.py
```diff
@@ -0,0 +1,48 @@
...
+    ('Check the map at https://www.openstreetmap.org', 'Check the map at '),
+    ('Use 203.0.113.255 for routing', 'Use  for routing'),
+    ('Find maps at https://osm.org and http://openstreetmap.org', 'Find maps at  and '),
+])
+def test_split_phrases(inp, outp):
+    query = [qmod.Phrase(qmod.PHRASE_ANY, inp)]
+
+    out = run_preprocessor_on(query)
+
+    assert out == [qmod.Phrase(qmod.PHRASE_ANY, outp)]
```
lonvia, line 48: This test clearly shows that the output if the preprocessors contains Phrases with the empty string.

lonvia (changes requested):

test/python/api/query_processing/test_regex_replace.py
TuringVerified: I was a bit confused about what "dropping phrases when they're empty" meant. Do I just return a list if the end output is empty?

docs/customize/Tokenizers.md
lonvia: The documentation is built using mkdocs, see the [Developer's documentation](https://nominatim.org/release-docs/develop/develop/Development-Environment/#documentation-pages). This page also has a list of packages you need to install to make it work.

test/python/api/query_processing/test_regex_replace.py
lonvia: Yes, if there was just one phrase, then the list should be empty in the end. If there are other Phrases, then they need to be kept, of course.

TuringVerified:
I got a bit confused due to the symlinks and thought that the config file was updated, but it wasn't. Fixed that now.

I compiled the documentation and fixed the indentation issue. Also, when there's a singular empty phrase, we should now just get an empty list straight. I've updated the tests to reflect this too.

TuringVerified:
I added docstrings in the .py file and I set members to True here so that it should all visible on the page at once now. But that also shows private functions. Is that something I should include? Or do I just add docstrings in the Python file itself?

lonvia:
You don't need the `members: True`. The documentation is taken from the module documentation, i.e. the docstring [here](https://github.com/osm-search/Nominatim/pull/3675/files#diff-a3919314ceca21964ea312baa1dfb83627556796d8f325e0363b0280af83e2a6R8).

TuringVerified:
Hi, I realised that I have a merge in my commit history and it seems to be complicating rebasing. For this PR, would you prefer a single squashed commit for simplicity? Or should I preserve the commit history?

lonvia:
The history should not have any merge commits. `git fetch origin; git rebase origin/master` usually does the trick.

Apart from that, it is your choice if you want to squash all into one commit or split up in multiple commits. It would be nice if the commits are somewhat meaningful. If there are "fixup" commits, squash them together with the commit they are fixing. (It takes a bit of getting used to but `git rebase -i master` is a really good tool to sort your commits.)

Feel free to force-push to the PR branch any time after rebasing.

lonvia (approved):
Rebase worked like a charm. Thank you.

Issue #2714, opened 2022-05-19
Linked place names need to include the default language


lonvia:
If a place is linked to a boundary and has different names, these names are added to the boundary. However, Nominatim doesn't do the expansion to the default language of the country there.

Example: the [boundary of Popayan](https://nominatim.openstreetmap.org/ui/details.html?osmtype=R&osmid=13111838&class=boundary) gets linked to the [town node](https://nominatim.openstreetmap.org/ui/details.html?osmtype=N&osmid=299736407&class=place). The boundary only has an ugly descriptive name while the node has the proper town name. English speakers get the `_place_name:en` while for Spanish speakers the `name:es` is preferred. From the `_place_name`, the additional `_place_name:es` (the default language of Colombia) needs to be inferred.

lonvia:
See https://github.com/osm-search/Nominatim/blob/5751686fdce33f6b69a93dd1641ceacb2239791f/lib-sql/functions/placex_triggers.sql#L1034 to learn where local names are added. See https://github.com/osm-search/Nominatim/blob/5751686fdce33f6b69a93dd1641ceacb2239791f/lib-sql/functions/placex_triggers.sql#L91 for where the place names from the linked place are linked in.

divyanshuagarwal-23:
Hello I want to pick this up

jayaddison:
I've begun exploring this and have installed the Nominatim application locally (although not as a managed service).  I'm able to get the test case added in #3687 (and only that test case) to fail, as expected, by removing the `@skip` decorator on it, and have learned a few things about behaviour-driven design along the way.

I _think_ the implementation for this may be best-placed into the `placex_triggers.sql` script, with some conditional logic to check for administrative boundaries, and then one or two joins to the parent object (I'm not _that_ familiar with the database schema yet, but it's beginning to make some sense).

Using `nomatim refresh ...` and `nominatim index ...` seem to be useful utilities to examine the effect of local/development changes on an imported dataset (for example, changing a single boundary at a time and then reindexing it).

A couple of questions:

* Is a join to the country table to obtain the default language required for this, or should the `name` from the parent node be used?
* If the parent-node name is identical to, or already included in the extratags for, the boundary, is it A) a requirement to include it, B) a requirement to exclude it, or C) whatever makes the implentation cleanest/easiest-to-maintain?

lonvia:
You need to do a lookup in the country table for the default languages. The translations of `name` are often user-supplied data and cannot be used to infer the default language.

[...]

PR #3687, opened 2025-03-28
test: linked places expand default language names

asharmalik19:
Added a failing BDD test for issue #2714 to verify default language expansion

### DIFF
--- a/test/bdd/db/import/linking.feature
+++ b/test/bdd/db/import/linking.feature
@@ -311,15 +311,16 @@ Feature: Linking of places
          | R1         | LabelPlace |
 
 
+    @Fail
     Scenario: Linked places expand default language names
         Given the grid
             | 1 |   | 2 |
             |   | 9 |   |
             | 4 |   | 3 |
         Given the places
-            | osm | class    | type   | name+name | name+name:en | name+name:ru | geometry |
-            | N9  | place    | city           | Popayán | Popayán | Попаян | 9 |
-            | R1  | boundary | administrative | Perímetro Urbano Popayán | | | (1,2,3,4,1) |
+            | osm | class    | type           | name+name                | geometry    |
+            | N9  | place    | city           | Popayán                  | 9           |
+            | R1  | boundary | administrative | Perímetro Urbano Popayán | (1,2,3,4,1) |
         And the relations
             | id | members  |
             | 1  | N9:label |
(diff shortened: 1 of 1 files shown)


test/bdd/db/import/linking.feature
```diff
@@ -310,3 +310,21 @@ Feature: Linking of places
...
 
+
+    Scenario: Linked places expand default language names
+        Given the grid
+            | 1 |   | 2 |
+            |   | 9 |   |
+            | 4 |   | 3 |
+        Given the places
+            | osm | class    | type   | name+name | name+name:en | name+name:ru | geometry |
+            | N9  | place    | city           | Popayán | Popayán | Попаян | 9 |
```
lonvia, line 321: You can even leave out the English and Russian names. They are not relevant for the example. (But they also do no harm.)

Can you please make sure the table columns are properly aligned? Makes it easier to read the tests.

lonvia (commented):
This looks almost good. If you add a `@Fail` above the scenario, like [here](https://github.com/osm-search/Nominatim/blob/master/test/bdd/db/import/postcodes.feature#L183), to mark that the test currently doesn't work, then we can directly merge this.

lonvia:
The current CI failure is not your fault. I'll commit a fix in a bit.

lonvia (approved):
Looks good now.

asharmalik19:
Thanks @lonvia

test/bdd/db/import/linking.feature
```diff
@@ -310,3 +310,22 @@ Feature: Linking of places
...
+            | osm | class    | type           | name+name                | geometry    |
+            | N9  | place    | city           | Popayán                  | 9           |
+            | R1  | boundary | administrative | Perímetro Urbano Popayán | (1,2,3,4,1) |
+        And the relations
+            | id | members  |
+            | 1  | N9:label |
+        When importing
+        Then placex contains
+            | object | name+_place_name | name+_place_name:es |
+            | R1     | Popayán          | Popayán             |
```
jayaddison, line 330: @asharmalik19 @lonvia maybe I have done something wrong, but I think the places that this test case creates are not linked places.  This test passes if I substitute this BDD assertion clause by using:

```diff
         When importing
         Then placex contains
-            | object | name+_place_name | name+_place_name:es |
-            | R1     | Popayán          | Popayán             |
-
+            | object      | linked_place_id |
+            | N9:place    | -               |
+            | R1:boundary | -               |
```

(in other words: neither the `N9` city node nor the `R1` boundary relation seem to have linked place IDs -- I think that's a prerequisite before the linked place names can be expanded?)

PR #3694, opened 2025-03-31
Committed changes in 'cursor.py'

AradhyaRay05:
1. Improved the error messages in 'scalar()' and 'row_set()' for making the debugging more easy.
2. Changes 'table_rows()' to use 'sql.SQL' and 'sql.Identifier' for minimalizing SQL Injection risks.
3. Added a where_params parameter to the table_rows() to properly handle special characters and prevent SQL Injection

### DIFF
--- a/test/python/cursor.py
+++ b/test/python/cursor.py
@@ -8,7 +8,7 @@
 Specialised psycopg cursor with shortcut functions useful for testing.
 """
 import psycopg
-
+from psycopg import sql
 
 class CursorForTesting(psycopg.Cursor):
     """ Extension to the DictCursor class that provides execution
@@ -20,7 +20,7 @@ class CursorForTesting(psycopg.Cursor):
             Raises an assertion when not exactly one row is returned.
         """
         self.execute(sql, params)
-        assert self.rowcount == 1
+        assert self.rowcount == 1, 'Expected exactly one row, got {}'.format(self.rowcount)
         return self.fetchone()[0]
 
     def row_set(self, sql, params=None):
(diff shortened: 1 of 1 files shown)


mtmail:
Can you run the linter? https://github.com/osm-search/Nominatim/blob/master/CONTRIBUTING.md#coding-style It will likely warn about missing space after comma and double-quotes.

lonvia:
Please give us more context on what the PR is supposed to do. In particular, a more meaningful PR title would be very helpful. Similarly, the commit messages should express what the commit changes (and if necessary why).

> 1.  Improved the error messages in 'scalar()' and 'row_set()' for making the debugging more easy.

Can you explain a bit more what prompted you to improve the messages? I presume you played around with the test and got confused by the error messages. Can you tell us a bit more what you were doing?

> 3.  Added a where_params parameter to the table_rows() to properly handle special characters and prevent SQL Injection

Can you give concrete examples where you see SQL injection issues and explain a bit more how your code change fixes them? Right now the `where_params` parameter isn't used anywhere, so it is difficult to judge how the code will function.

AradhyaRay05:
> > 1. Improved the error messages in `'scalar()' `and `'row_set()' `for making the debugging more easy.

>Can you explain a bit more what prompted you to improve the messages? I presume you played around with the test and got confused by the error messages. Can you tell us a bit more what you were doing?

The original error messages were very minimal. For example, with the original `scalar() `method: `assert self.rowcount == 1 `, if this assertion fails, the error would simply be "AssertionError" without any context. A developer would need to use a debugger to figure out how many rows were actually returned. With the improved message it will show for example, `Expected exactly 1 row, got 3`. It makes debugging much faster because anyone can immediately understand what's wrong without adding using a debugger.

> > 3. Added a `where_params `parameter to the `table_rows() `to properly handle special characters and prevent SQL Injection

>Can you give concrete examples where you see SQL injection issues and explain a bit more how your code change fixes them? Right now the where_params parameter isn't used anywhere, so it is difficult to judge how the code will function.

You're right that the `where_params` parameter isn't being used anywhere, so it is difficult to judge how the code will function.
But in the original code, if someone passed malicious text in the `where` clause, it would be directly inserted into the SQL statement and potentially executed. This can be dangerous because a malicious user could give an input which can change the meaning of query or executes harmful commands. With the `where_params` parameter. It separates what you want to search for from the actual values you're checking against. It first tells the database what you want to search and then handing the value separately, so database can't execute the harmful part.

lonvia:
> The original error messages were very minimal. For example, with the original scalar() method: assert self.rowcount == 1 , if this assertion fails, the error would simply be "AssertionError" without any context. 

I wonder why [pytest's extentended assert report](https://docs.pytest.org/en/stable/how-to/assert.html#asserting-with-the-assert-statement) isn't being used here. That's a pretty awsome feature of pytest and I very much prefer it over having to write custom assert messages.

> You're right that the where_params parameter isn't being used anywhere,

Probably better to just get rid of the where completely then.

lonvia:
This needs a different solution. And better commit messages.

PR #3709, opened 2025-04-07
Improve README formatting and add install steps

anqixxx:
I added a general cloning step before the virtual environment. This would have been helpful for me during Nominatim setup, and would have liked to have this in terms of ordering beforehand!

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,18 +27,25 @@ can be found at nominatim.org as well.
 
 A quick summary of the necessary steps:
 
-1. Create a Python virtualenv and install the packages:
+
+1. Clone this git repository and download the country grid
+
+        git clone https://github.com/osm-search/Nominatim.git
+        wget -O Nominatim/data/country_osm_grid.sql.gz https://nominatim.org/data/country_grid.sql.gz
+
+2. Create a Python virtualenv and install the packages:
 
         python3 -m venv nominatim-venv
         ./nominatim-venv/bin/pip install packaging/nominatim-{api,db}
 
-2. Create a project directory, get OSM data and import:
+3. Create a project directory, get OSM data and import:
 
         mkdir nominatim-project
         cd nominatim-project
-        ../nominatim-venv/bin/nominatim import --osm-file <your planet file>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,18 +27,25 @@ can be found at nominatim.org as well.
 
 A quick summary of the necessary steps:
 
-1. Create a Python virtualenv and install the packages:
+
+1. Clone this git repository and download the country grid
+      git clone https://github.com/osm-search/Nominatim.git
```
lonvia, line 32: Please add an empty line before this line so we get a code block.

lonvia (commented):
Except for the small formatting issue, this looks good to me.

README.md
anqixxx: Done!

lonvia: The formatting is still wrong. Can you please look at https://github.com/anqixxx/Nominatim/tree/update-readme?tab=readme-ov-file#installation and make sure it looks okay?

README.md
```diff
@@ -27,18 +27,26 @@ can be found at nominatim.org as well.
...
         python3 -m venv nominatim-venv
         ./nominatim-venv/bin/pip install packaging/nominatim-{api,db}
 
-2. Create a project directory, get OSM data and import:
+3. Create a project directory, get OSM data and import:
 
         mkdir nominatim-project
         cd nominatim-project
         ../nominatim-venv/bin/nominatim import --osm-file <your planet file>
+        <!-- Alternative command with logging: ../nominatim-venv/bin/nominatim import --osm-file <your planet file> 2>&1 | tee setup.log -->
```
lonvia, line 46: You can't use HTML syntax here. Also, this is getting too long for a quick setup. Lets just add the logging to the normal command and be done with it. I see that the full manual already does the same.

lonvia:
Not sure if my comment last week made it through but the formatting still needs a bit of fixing here. That's why it still isn't merged.

anqixxx:
> Not sure if my comment last week made it through but the formatting still needs a bit of fixing here. That's why it still isn't merged.

Fixed! Double checked on my branch to make sure it looks correct!

image

lonvia:
CI error to be fixed on the other branch.

PR #3710, opened 2025-04-08
Fix special phrases filtering

### LINKED ISSUE #235: Verify plausability of special phrase terms
The Special Phrase pages in the wiki get the occasional entry for tag combinations that don't really exist. Two things: a) make sure the key list is restricted to tags actually imported by osm2pgsql. b) Taginfo could be used to verify that a key/value combination actually exists.

anqixxx:
Fixes #235

Referencing https://github.com/osm-search/Nominatim/issues/235#issuecomment-1441967793 and https://github.com/osm-search/Nominatim/issues/235#issuecomment-805173363 I created two new functions to get valid special filters from both the database and the style file. The former only uses class/type combinations with more than 100 occurrences, while the latter only uses only class/type combinations with the main property.

For the style file, https://nominatim.org/release-docs/latest/develop/Import/#configuring-the-import is a depreciated link. I took this to assume that the style files may potentially be depreciated. However, on the chance that it was not, I still integrated into my code using https://github.com/osm-search/Nominatim/blob/28b4fb12b684a5c76ca24002ba1ae19ce2e22c87/settings/import-extratags.style .

These were combined into get_sp(), which returns a set of string, string tuples. What this does is essentially creates a new set with only elements from both get_sp_style() and get_sp_db(). I also created a testing file in /test for sp_importer.py. Documentation for the three new functions was included.

This code was integrated after the [index on classes and types has been created](https://github.com/osm-search/Nominatim/blob/8191c747b964530f67b064b43954e13a5b1a0791/nominatim/tools/special_phrases/sp_importer.py#L180) within _create_classtype_table_and_indexes(). When iterating through, I check to see if the phrase class and phrase type tuple is in the allowed set, and if not, they are removed.

Further testing using a more advanced knowledge of the database may be nessecary, as the database testing was done using [MagicMock](https://docs.python.org/3/library/unittest.mock.html).

### DIFF
--- a/src/nominatim_db/tools/special_phrases/sp_importer.py
+++ b/src/nominatim_db/tools/special_phrases/sp_importer.py
@@ -16,7 +16,7 @@
 from typing import Iterable, Tuple, Mapping, Sequence, Optional, Set
 import logging
 import re
-
+import json 
 from psycopg.sql import Identifier, SQL
 
 from ...typing import Protocol
@@ -65,6 +65,52 @@ class SPImporter():
         # special phrases class/type on the wiki.
         self.table_phrases_to_delete: Set[str] = set()
 
+    def get_classtype_pairs_style(self) ->  Set[Tuple[str, str]]:
+        """
+            Returns list of allowed special phrases from the the style file, 
+            restricting to a list of combinations of classes and types 
+            which have a 'main' property
+
+            Note: This requirement was from 2021 and I am a bit unsure if it is still relevant
+        """
+        style_file = self.config.get_import_style_file() # this gives the path, so i will import it as a json
+        with open(style_file, 'r') as file:
+            style_data = json.loads(f'[{file.read()}]')
+
+        style_combinations = set()
+        for _map in style_data: # following ../settings/import-extratags.style
+            classes = _map.get("keys", [])
+            values = _map.get("values", {})
+
+            for _type, properties in values.items():
+                if "main" in properties and _type: # make sure the tag is not an empty string. since type is the value of the main tag
+                    for _class in classes:
+                        style_combinations.add((_class, _type)) 
+
+        return style_combinations
+    
+    def get_classtype_pairs(self) ->  Set[Tuple[str, str]]:
+        """
+            Returns list of allowed special phrases from the database, 
+            restricting to a list of combinations of classes and types 
+            whic occur more than 100 times
+        """
+        db_combinations = set() 
+        query = """
+        SELECT class AS CLS, type AS typ
+        FROM placex
+        GROUP BY class, type
+        HAVING COUNT(*) > 100
+        """
+
+        with self.db_connection.cursor() as db_cursor:
+            db_cursor.execute(SQL(query))   
+            for row in db_cursor.fetchall():
+                db_combinations.add((row[0], row[1]))
+
+        return  db_combinations
+
+    
     def import_phrases(self, tokenizer: AbstractTokenizer, should_replace: bool) -> None:
         """
             Iterate through all SpecialPhrases extracted from the
@@ -85,9 +131,11 @@ class SPImporter():
             if result:
                 class_type_pairs.add(result)
 
-        self._create_classtype_table_and_indexes(class_type_pairs)
+        self._create_classtype_table_and_indexes(class_type_pairs) 
         if should_replace:
             self._remove_non_existent_tables_from_db()
+
+        
         self.db_connection.commit()
 
         with tokenizer.name_analyzer() as analyzer:
@@ -177,10 +225,17 @@ class SPImporter():
         with self.db_connection.cursor() as db_cursor:
             db_cursor.execute("CREATE INDEX idx_placex_classtype ON placex (class, type)")
 
+        allowed_special_phrases = self.get_classtype_pairs()
+
         for pair in class_type_pairs:
             phrase_class = pair[0]
             phrase_type = pair[1]
 
+            if (phrase_class, phrase_type) not in allowed_special_phrases:
+                LOG.warning("Skipping phrase %s=%s: not in allowed special phrases",
+                            phrase_class, phrase_type)
+                continue
+            
             table_name = _classtype_table(phrase_class, phrase_type)
 
             if table_name in self.table_phrases_to_delete:
@@ -212,8 +267,8 @@ class SPImporter():
             if doesn't exit.
         """
         table_name = _classtype_table(phrase_class, phrase_type)
-        with self.db_connection.cursor() as cur:
-            cur.execute(SQL("""CREATE TABLE IF NOT EXISTS {} {} AS
+        with self.db_connection.cursor() as db_cursor:
+            db_cursor.execute(SQL("""CREATE TABLE IF NOT EXISTS {} {} AS
                                  SELECT place_id AS place_id,
                                         st_centroid(geometry) AS centroid
                                  FROM placex
@@ -266,3 +321,4 @@ class SPImporter():
         drop_tables(self.db_connection, *self.table_phrases_to_delete)
         for _ in self.table_phrases_to_delete:
             self.statistics_handler.notify_one_table_deleted()
+
--- a/src/nominatim_db/tools/special_phrases/sp_importer.py
+++ b/src/nominatim_db/tools/special_phrases/sp_importer.py
@@ -64,23 +64,25 @@ class SPImporter():
         # special phrases class/type on the wiki.
         self.table_phrases_to_delete: Set[str] = set()
 
-    def get_classtype_pairs(self) -> Set[Tuple[str, str]]:
+    def get_classtype_pairs(self, min: int = 0) -> Set[Tuple[str, str]]:
         """
             Returns list of allowed special phrases from the database,
             restricting to a list of combinations of classes and types
-            which occur more than 100 times
+            which occur more than a specified amount of times.
+
+            Default value for this, if not specified, is at least once.
         """
         db_combinations = set()
-        query = """
+        query = f"""
         SELECT class AS CLS, type AS typ
         FROM placex
         GROUP BY class, type
-        HAVING COUNT(*) > 100
+        HAVING COUNT(*) > {min}
         """
 
         with self.db_connection.cursor() as db_cursor:
             db_cursor.execute(SQL(query))
-            for row in db_cursor.fetchall():
+            for row in db_cursor:
                 db_combinations.add((row[0], row[1]))
 
         return db_combinations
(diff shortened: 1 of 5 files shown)


src/nominatim_db/tools/special_phrases/sp_importer.py
```diff
@@ -177,10 +232,17 @@ def _create_classtype_table_and_indexes(self,
         with self.db_connection.cursor() as db_cursor:
             db_cursor.execute("CREATE INDEX idx_placex_classtype ON placex (class, type)")
 
+        allowed_special_phrases = self.get_sp()
```
lonvia, line 235: You don't really need to get the intersection of database and style class/types here. It is enough to fetch the class/types from the database. The rest of the code takes care of reading from the style and the actual intersection happens then through the condition in line 241.

src/nominatim_db/tools/special_phrases/sp_importer.py
```diff
@@ -85,9 +138,11 @@ def import_phrases(self, tokenizer: AbstractTokenizer, should_replace: bool) ->
             if result:
                 class_type_pairs.add(result)
 
-        self._create_classtype_table_and_indexes(class_type_pairs)
+        self._create_classtype_table_and_indexes(class_type_pairs) # from the issue, this said to filter this after, which doesnt make sense
```
lonvia, line 141: What the comment in the issue meant: at this point we create a special index on the placex table that helps looking up class/type pairs. Without that index, fetching the pairs from a planet-wide database will take a very long time.

Can you please remove the comment for the final version. It will likely be more confusing than helpful for somebody reading the code in a couple of months time because they won't know what issue you refer to.

src/nominatim_db/tools/special_phrases/sp_importer.py
```diff
@@ -65,6 +65,59 @@ def __init__(self, config: Configuration, conn: Connection,
...
+            values = _map.get("values", {})
+
+            for _type, properties in values.items():
+                if "main" in properties and _type: # make sure the tag is not an empty string. since type is the value of the main tag
+                    for _class in classes:
+                        style_combinations.add((_class, _type)) 
+
+        return style_combinations
+    
+    def get_sp_db(self) ->  Set[Tuple[str, str]]:
```
lonvia, line 92: It's better to spell out such function names to make it easier for a future code reader. And you are not really getting special phrases, you get class/type pairs.

lonvia (commented):
Can you please remove the commit with the changes to the README. That is another PR.

Some general remarks to the tests:

pytest has a convenient fixture for mocking, see https://docs.pytest.org/en/stable/how-to/monkeypatch.html. You should use that instead of MagicMock.

Mocking is fine for most of the test functions but you should have at least one test that ensures that `get_sp_db()` works correctly with a real database. There is a [fixture for creating a placex table](https://github.com/osm-search/Nominatim/blob/50d4b0a386c98a1046e19db4fa61e3fe9f0ca9b5/test/python/conftest.py#L183) that you can use. Simply use it and fill the table with some interesting data:

```
def test_foo(placex_table):
    placex_table.add(cls='amenity', typ='prison')
    placex_table.add(cls='tourism', typ='hotel')
    ...
```

The `add` function will conveniently fill all the other columns of the table that don't really interest you with reasonable values.

src/nominatim_db/tools/special_phrases/sp_importer.py
```diff
@@ -65,6 +65,59 @@ def __init__(self, config: Configuration, conn: Connection,
         # special phrases class/type on the wiki.
         self.table_phrases_to_delete: Set[str] = set()
 
+    def get_sp_style(self) ->  Set[Tuple[str, str]]:
+        """
+            Returns list of allowed special phrases from the the style file, 
+            restricting to a list of combinations of classes and types 
+            which have a 'main' property
+
+            Note: This requirement was from 2021 and I am a bit unsure if it is still relevant
```
lonvia, line 74: Let's leave it out for now. Having items in database is a good enough approximation.

src/nominatim_db/tools/special_phrases/sp_importer.py
anqixxx: I removed the comment for the final version!

src/nominatim_db/tools/special_phrases/sp_importer.py
anqixxx: I changed the function names to get_classtype_pairs and get_classtype_pairs_style, for the database and style versions respectively! I removed the overaching old get_sp function in light of the below comment, and just used the database version as the default function used in _create_classtype_table_and_indexes !

src/nominatim_db/tools/special_phrases/sp_importer.py
anqixxx: I removed it from the main code flow but left the function. Should I remove the function also to avoid future confusion?

anqixxx:
> Can you please remove the commit with the changes to the README. That is another PR.
> 
> Some general remarks to the tests:
> 
> pytest has a convenient fixture for mocking, see https://docs.pytest.org/en/stable/how-to/monkeypatch.html. You should use that instead of MagicMock.
> 
> Mocking is fine for most of the test functions but you should have at least one test that ensures that `get_sp_db()` works correctly with a real database. There is a [fixture for creating a placex table](https://github.com/osm-search/Nominatim/blob/50d4b0a386c98a1046e19db4fa61e3fe9f0ca9b5/test/python/conftest.py#L183) that you can use. Simply use it and fill the table with some interesting data:
> 
> ```
> def test_foo(placex_table):
>     placex_table.add(cls='amenity', typ='prison')
>     placex_table.add(cls='tourism', typ='hotel')
>     ...
> ```
> 
> The `add` function will conveniently fill all the other columns of the table that don't really interest you with reasonable values.

I also updated test_sp_importer to use monkeypatching instead of MagicMock, and created the mock real database as well!

src/nominatim_db/tools/special_phrases/sp_importer.py
lonvia: Yes, please remove it. Also anything related to it in the tests.

src/nominatim_db/tools/special_phrases/sp_importer.py
```diff
@@ -65,6 +65,52 @@ def __init__(self, config: Configuration, conn: Connection,
...
+                    for _class in classes:
+                        style_combinations.add((_class, _type)) 
+
+        return style_combinations
+    
+    def get_classtype_pairs(self) ->  Set[Tuple[str, str]]:
+        """
+            Returns list of allowed special phrases from the database, 
+            restricting to a list of combinations of classes and types 
+            whic occur more than 100 times
```
lonvia, line 96: typo 'whic'

src/nominatim_db/tools/special_phrases/sp_importer.py
```diff
@@ -65,6 +65,52 @@ def __init__(self, config: Configuration, conn: Connection,
...
+        """
+        db_combinations = set() 
+        query = """
+        SELECT class AS CLS, type AS typ
+        FROM placex
+        GROUP BY class, type
+        HAVING COUNT(*) > 100
```
lonvia, line 103: I have a doubt about this limit because it is an arbitrary number but still hard-coded. If we'd leave it in, we need to make it somehow configurable. For the moment I would suggest we just allow everything that appears at least once. If necessary, a minimum occurrences switch can be added later.

anqixxx: Okay yes that makes sense! I might have taken the previous comments on the issue a bit too literally:)

anqixxx: I added an optional argument, min, which represents the minimum number of times it must appear! The default is set to 0, meaning that we currently just allow everything that appears at least once. I also created a new test to test this!

src/nominatim_db/tools/special_phrases/sp_importer.py
anqixxx: Fixed!

lonvia (commented):
Looks quite good now. I've just added a few minor style comments.

I think the current CI failures are there because you are not on the latest version of Nominatim. Rebase on latest master (something along the line of: `git fetch origin` and `git rebase origin/master`) and force push (`git push -f`) to your github branch.

One last test would be good to add to also test your changes to `import_phrases()`. Have a look at the [test for import_phrases()](https://github.com/osm-search/Nominatim/blob/478a8741dbd4262907566d5d82ec4d5285b91484/test/python/tools/test_import_special_phrases.py#L179) and see if you can either expand the test to verify skipping works or add another test that looks similar and explicitly tests the skipping.

src/nominatim_db/tools/special_phrases/sp_importer.py
```diff
@@ -212,14 +242,14 @@ def _create_place_classtype_table(self, sql_tablespace: str,
...
-                             """).format(Identifier(table_name), SQL(sql_tablespace)),
-                        (phrase_class, phrase_type))
+        with self.db_connection.cursor() as db_cursor:
+            db_cursor.execute(SQL(
+                """CREATE TABLE IF NOT EXISTS {} {} AS
+                SELECT place_id AS place_id,
+                st_centroid(geometry) AS centroid
+                FROM placex WHERE class = %s AND type = %s
+                """).format(Identifier(table_name), SQL(sql_tablespace)),
+                (phrase_class, phrase_type))
```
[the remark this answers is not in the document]
anqixxx: Okay! The old format was just failing under the linter, but I'll revert it back!

src/nominatim_db/tools/special_phrases/sp_importer.py
```diff
@@ -65,6 +64,29 @@ def __init__(self, config: Configuration, conn: Connection,
...
+        query = f"""
+        SELECT class AS CLS, type AS typ
+        FROM placex
+        GROUP BY class, type
+        HAVING COUNT(*) > {min}
+        """
+
+        with self.db_connection.cursor() as db_cursor:
+            db_cursor.execute(SQL(query))
+            for row in db_cursor.fetchall():
```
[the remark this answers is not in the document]
anqixxx: Done! Thank you!

src/nominatim_db/tools/special_phrases/sp_importer.py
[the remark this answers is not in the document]
lonvia: What did the linter exactly complain about? Note that you need to revert completely, including your initial change from `cur` to `db_cursor`.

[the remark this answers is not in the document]
anqixxx: Sorry for being so MIA! Just finished finals! Added tests in test_import_phases to test the new special phrases filter. For the minimum, instead of changing the logic to accept a new optional argument in import_phases, I just left the default minimum as 1, as I was worried about backwards compatibility!

As for the linter, before it was giving a **src/nominatim_db/tools/special_phrases/sp_importer.py:254:13: E128 continuation line under-indented for visual indent** issue when I ran it. However when I went to the main repository and just copied everything over, it worked!

PR #3719, opened 2025-04-14
Estimate query direction

lonvia:
Nominatim was always able to handle addresses going in left-to-right or right-to-left direction. It would more or less handle both directions equally, which means twice the queries to send to the database. This PR introduces an estimation of the query direction using the token counts from the database: for every token in the query compute the ratio of how often the token appears in the name of an objects vs. appearing in its address. Then compute the linear regression over these ratios in the query to figure out if towards the end of the query, the tokens are more or less likely to appear in the name. The regression is converted into a penalty.

Currently the direction penalty only applies to standard name searches and none of the special searches (e.g. for postcodes). They usually have already a direction bias anyway because of the query structure they require.

The PR also contains some cleanup and reworing of the internal query structures. See individual commits.

### DIFF
--- a/src/nominatim_api/search/query.py
+++ b/src/nominatim_api/search/query.py
@@ -12,7 +12,14 @@ from abc import ABC, abstractmethod
 from collections import defaultdict
 import dataclasses
 
-
+# Precomputed denominator for the computation of the linear regression slope
+# used to determine the query direction.
+# The x value for the regression computation will be the position of the
+# token in the query. Thus we know the x values will be [0, query length).
+# As the denominator only depends on the x values, we can pre-compute here
+# the denominatior to use for a given query length.
+# Note that query length of two or less is special cased and will not use
+# the values from this array. Thus it is not a problem that they are 0.
 LINFAC = [i * (sum(si * si for si in range(i)) - (i - 1) * i * (i - 1) / 4)
           for i in range(50)]
 
--- a/src/nominatim_api/search/query.py
+++ b/src/nominatim_api/search/query.py
@@ -183,10 +183,10 @@ class QueryNode:
     """ Penalty for the break at this node.
     """
     term_lookup: str
-    """ Transliterated term following this node.
+    """ Transliterated term ending at this node.
     """
     term_normalized: str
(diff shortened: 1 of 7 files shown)


src/nominatim_api/search/query.py
```diff
@@ -13,6 +13,10 @@
 import dataclasses
 
 
+LINFAC = [i * (sum(si * si for si in range(i)) - (i - 1) * i * (i - 1) / 4)
```
mtmail, line 16: This is a list `[0.0, 0.0, 1.0, 6.0, 20.0, 50.0, 105.0, 196.0, 336.0, 540.0, 825.0, 1210.0, 1716.0, 2366.0, 3185.0, 4200.0, 5440.0, 6936.0, 8721.0, 10830.0, 13300.0, 16170.0, 19481.0, 23276.0, 27600.0, 32500.0, 38025.0, 44226.0, 51156.0, 58870.0, 67425.0, 76880.0, 87296.0, 98736.0, 111265.0, 124950.0, 139860.0, 156066.0, 173641.0, 192660.0, 213200.0, 235340.0, 259161.0, 284746.0, 312180.0, 341550.0, 372945.0, 406456.0, 442176.0, 480200.0]` and it feels all magic how the list length or the maximum number was selected.

Add some basic tests on `compute_direction_penalty` could help as documentation
image

lonvia: It's not magic, it's the part of the [linear regression function](https://en.wikipedia.org/wiki/Simple_linear_regression#Numerical_example) that can be precomputed.

In [this line](https://github.com/osm-search/Nominatim/pull/3719/files#diff-9c80e660fec66de9d99fcfce36458fc58c283a348608fbf0785d2cd6be025c1fR327) we compute the slope of the linear regression with x being the position of the token in the query and y being the ratio of appearances in names vs. total appearances of the token. As the denominator of the formula only depends on x, it will be constant with respect to the number of tokens in the query. So that's what `LINFAC` precomputes.

And in case you are wondering about the zero values in the array: `LINFAC[0:2]` are never used because they are special cased in the formula below.

I'll add a comment to make that clearer.

PR #3722, opened 2025-04-17
resolve datetime deprecation warnings

emmanuel-ferdman:
# PR Summary
This PR fixes the `DeprecationWarning` that appears in Python 3.12+ when using `datetime.utcnow()`. It replaces it with a more future-proof approach: `datetime.now(timezone.utc).replace(tzinfo=None)`. This keeps the `datetime` object naive (without timezone info), which is important for parts of the code that expect it that way, but avoids the warning by using a timezone-aware method first:
```python
# Before (deprecated)  
from datetime import datetime  
naive_time = datetime.utcnow()  # Raises warning  

# Common Fix (timezone-aware)  
from datetime import datetime, timezone  
aware_time = datetime.now(timezone.utc)  # Breaks naive-dependent code  

# This PR's Fix (naive but compliant)  
from datetime import datetime, timezone  
compliant_time = datetime.now(timezone.utc).replace(tzinfo=None)  # No warning, backward-compatible  
```
You can see this warning in the [latest CI run](https://github.com/osm-search/Nominatim/actions/runs/14516596010/job/40727023541#step:15:122):
```
Thu, 17 Apr 2025 13:26:45 GMT
  /home/runner/work/Nominatim/Nominatim/Nominatim/src/nominatim_api/v1/format_xml.py:93: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
Thu, 17 Apr 2025 13:26:45 GMT
    root.set('timestamp', dt.datetime.utcnow().strftime('%a, %d %b %Y %H:%M:%S +00:00'))
```

### DIFF
--- a/src/nominatim_api/v1/format_xml.py
+++ b/src/nominatim_api/v1/format_xml.py
@@ -90,7 +90,7 @@ def format_base_xml(results: Union[ReverseResults, SearchResults],
         result will be output, otherwise a list.
     """
     root = ET.Element(xml_root_tag)
-    root.set('timestamp', dt.datetime.utcnow().strftime('%a, %d %b %Y %H:%M:%S +00:00'))
+    root.set('timestamp', dt.datetime.now(dt.timezone.utc).strftime('%a, %d %b %Y %H:%M:%S +00:00'))
     root.set('attribution', cl.OSM_ATTRIBUTION)
     for k, v in xml_extra_info.items():
         root.set(k, v)


lonvia:
Thanks, for fixing. I meant to look into that for ages.

PR #3781, opened 2025-07-14
Reduce number of tokens used for index lookups during search

lonvia:
When looking something up in the search_name table, we've so far used the full set of tokens from the query. The assumption here was that more tokens means better selectivity and fewer rows to load from the search_name table for further inspection. The strategy can be counterproductive when the query has a few tokens with relatively low frequency and other tokens with high number of occurrences. The frequent tokens will then take a long time to be looked up in the index while it is actually the low-frequency tokens that contribute most to reducing the number of candidates.

This changes the search strategy to order the search tokens by their frequency and then use at max the three least frequent tokens for the index lookup. All other tokens are used only for restricting results after index use. The new strategy is also able to switch to address token lookup when the tokens for the name are too frequent and less frequent terms are available in the address. This should speed up searches for queries like "Main St, Hogspogstown".

Also restricts postcode lookups to only match when they are in the same country as the place.

### DIFF
--- a/src/nominatim_api/search/db_search_builder.py
+++ b/src/nominatim_api/search/db_search_builder.py
@@ -226,10 +226,10 @@ class SearchBuilder:
                                 ) -> Iterator[Tuple[float, int, List[dbf.FieldLookup]]]:
         """ Yield the best lookup for a name-only search.
         """
-        split = partials.get_num_lookup_tokens(30000, 10)
+        split = partials.get_num_lookup_tokens(30000, 6)
 
         if split > 0:
-            yield 0.0, partials.expected_for_all_search(10), \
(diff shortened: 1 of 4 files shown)


mtmail:
> If a token count is one, then statistics are likely to be unavaible[sic] and a relatively high count is assumed instead.

What would make the statistics unavailable? Are they not collected in some queries for performance?

lonvia:
Word statistics are only computed once at the end of the import. While we've had mandatory name term statistics for quite a while, address term statistics only arrived with version 4.5. So it's quite possible that people still have a database without address statistics.

The other way, a term may end up without statistics unavailable is when it newly arrives in the database through updates. Although it can be assumed that such terms are relatively rare, the PR prefers to err on the side of caution and go with known statistics instead.

Issue #3794, opened 2025-07-24
Nominatim not creating settings database during import


saharri2:
**Describe the bug**

Nominatim is not creating the settings relation in the database.

**To Reproduce**

I just ran Nominatim import as normal, only changes to the .env file were for paths to files such as the flatnode file. After the import completed (at least, I think it completed? It didn't produce an error and returned to the prompt) I ran nominatim admin --check-database and it showed that the DB version (it showed as (None)) did not match the Nominatim software version. I opened a shell in the database to see if the settings relation was created, and it was not. I used a python venv to run the import, as recommended by the documentation.

**Software Environment (please complete the following information):**
- Nominatim version: 5.1.0-0
- Postgresql version: 17.5
- Postgis version: 3.5.2
- OS: Ubuntu 25.04
- Python: 3.9

**Hardware Configuration (please complete the following information):**
- RAM: 62 GB
- number of CPUs: 16
- type and size of disks: SSD, 500GB

**Nominatim Configuration:**

Only change was adding the path to the flatnode file. I didn't create a .env file, just used export before running the import.

**Additional context**

* Here is the output of --check-database:

2025-07-24 11:13:35: Using project directory: /home/sonny/nominatim-project
2025-07-24 11:13:35: Checking database
Checking database connection ... OK
Checking database_version matches Nominatim software version ... Failed
Database version (None) doesn't match Nominatim version (5.1.0-0)

Hints:
* Are you connecting to the correct database?

Check the Migration chapter of the Administration Guide.

Project directory: /home/sonny/nominatim-project
Current setting of NOMINATIM_DATABASE_DSN: pgsql:dbname=nominatim

* I also tried to migrate the DB, but it says that the DB version is older than 4.3.

* And finally, I had to use Python 3.9 because the icu tokenizer uses the datrie library, which does not support newer python versions.

lonvia:
You need to tell us how the normal import looked like and what the output was. If the import really was successful then it hasn't imported into the database you think it was importing to. Check if there is a database with another name around.

Nominatim 5.1 no longer has a [dependency on datrie](https://github.com/osm-search/Nominatim/pull/3659).

saharri2:
> You need to tell us how the normal import looked like and what the output was. If the import really was successful then it hasn't imported into the database you think it was importing to. Check if there is a database with another name around.
> 
> Nominatim 5.1 no longer has a [dependency on datrie](https://github.com/osm-search/Nominatim/pull/3659).

[...]

PR #3797, opened 2025-07-30
Better hint to user if database import didnt finish

mtmail:
Improve error message for when the database import hasn't finished (when `nominatim_properties` table is created, but the `database_version` record hasn't been written yet). For https://github.com/osm-search/Nominatim/issues/3794

### DIFF
--- a/src/nominatim_db/tools/check_database.py
+++ b/src/nominatim_db/tools/check_database.py
@@ -2,7 +2,7 @@
 #
 # This file is part of Nominatim. (https://nominatim.org)
 #
-# Copyright (C) 2024 by the Nominatim developer community.
+# Copyright (C) 2025 by the Nominatim developer community.
 # For a full list of authors see the git log.
 """
 Collection of functions that check if the database is complete and functional.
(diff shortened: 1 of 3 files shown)


src/nominatim_db/tools/check_database.py
```diff
@@ -163,37 +163,36 @@ def check_connection(conn: Any, config: Configuration) -> CheckResult:
...
-            else 'You need to upgrade the Nominatim software.'
-        )
-    else:
-        instruction = ''
+            instruction = (
+                'Run migrations: nominatim admin --migrate'
+                if db_version < NOMINATIM_VERSION
+                else 'You need to upgrade the Nominatim software.'
+
+                'Check the Migration chapter of the Administration Guide.'
```
lonvia: You want to add a space or better a newline in front of this line.

lonvia (commented):
Tiny formatting issue. Looks good otherwise.

mtmail:
Added now. But previously it only printed in the `else` case. I made sure it always prints.

Issue #2833, opened 2022-09-28
"Tor 3, REYHER, Hamburg" not found


westnordost:
## What did you search for?

[Tor 3, REYHER, Hamburg](https://nominatim.openstreetmap.org/ui/search.html?q=Tor+3%2C+Reyher%2C+Hamburg)

## What result did you get?

None

## What result did you expect?

https://www.openstreetmap.org/node/9396927629

## Further details

[REYHER, Hamburg](https://nominatim.openstreetmap.org/ui/search.html?q=Reyher%2C+Hamburg) returns the expected result: The company premises of Reyher.
So, Nominatim has the ability to find objects independent of `addr:*` tags and also has some notion of which objects are contained in other objects. Gates (on the outline of such objects) with either `name` or `ref` seem to not be regarded, however.

lonvia:
`barrier=*` objects are not indexed at the moment.

lonvia:
Note that this issue comes out of a discussion on the German forum: https://community.openstreetmap.org/t/pforte-wie-eintragen/3047 Unfortunately there wasn't any usable conclusion there.

One thing is indisputable: Nominatim currently does not index named entrances/gates. That should be changed.

Looking at the example and other named `barrier=gate` I wonder though, if the barrier object is the right one to have the name. Isn't it rather the overall entrance which is named.

So the question is: should `barrier=gate` be indexed (none of the other barrier values are really make sense to be named, so it would be the exception) or `entrance=*`?

westnordost:
> Isn't it rather the overall entrance which is named.

What do you mean? How would the overall entrance (to fenced company premises) be tagged if not as a barrier=gate?

I was under the impression that entrance=* is rather only used on building entrances, not on fences and walls. At least in German, an "entrance" through a fence would always be called a "gate", no matter the size.

lonvia:
> What do you mean? How would the overall entrance (to fenced company premises) be tagged if not as a barrier=gate?

A `barrier=gate` is strictly speaking only the 'section in a [wall](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dwall) or [fence](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dfence) which can be opened to allow or restrict access'.  An entrance (or the German 'Tor' you refer to here) may be comprised of more than just that movable part. The referenced discussion mentions a gate house as an example. Or think about the huge multi-gate entrances to zoos.

[...]

PR #3807, opened 2025-08-05
Index and return entrance coordinates for places

### LINKED ISSUE #536: return entrance point for large places like airports
on osrm-talk there is a discussion
https://www.mail-archive.com/osrm-talk@openstreetmap.org/msg01223.html
summary:
in a scenario when finding a route to airport/zoo, nominatim returns (?) centroid of the big feature. snapping the centroid to the nearest road network sometimes returns very wrong route (ending nowhere near the entry).

I guess this is a tagging issue, nominatim should know the "normal entry point" from somewhere. Common cases are big features (airport, zoo, train stations) or even houses/addresses (centroid can be closer to some service road at the back, where there is no door).

is there any existing tagging to solve this? Like putting entrance=main on the polygon of the address so that nominatim takes entrance instead of centroid.

emlove:
## Description
This PR adds logic to record the main entrance location for places. I'd love to get some feedback on the approach from maintainers on the approach here.

In this implementation a new `place_entrance` table, where entrance node details are stored.

Fixes #536 

Example output:
```json
[
    {
        "place_id": 534548,
        "licence": "Data © OpenStreetMap contributors, ODbL 1.0. http://osm.org/copyright",
        "osm_type": "way",
        "osm_id": 108240964,
        "lat": "41.8838293",
        "lon": "-87.6319547",
        "category": "amenity",
        "type": "townhall",
        "place_rank": 30,
        "importance": 9.99999999995449e-06,
        "addresstype": "amenity",
        "name": "Chicago City Hall",
        "display_name": "Chicago City Hall, 121, Pedway, Loop, Chicago, Cook County, 60602, United States",
        "entrances": [
            {
                "lat": 41.883881,
                "lon": -87.6322362,
                "type": "yes",
                "osm_node_id": 2391052802
            },
            {
                "lat": 41.8843544,
                "lon": -87.6317131,
                "type": "yes",
                "osm_node_id": 2391052808
            },
            {
                "lat": 41.8836391,
                "lon": -87.632232,
                "type": "yes",
                "osm_node_id": 2391057836
            },
            {
                "lat": 41.8833509,
                "lon": -87.6317131,
                "type": "yes",
                "osm_node_id": 10922694346
            }
        ],
        "boundingbox": [
            "41.8833431",
            "41.8843544",
            "-87.6322445",
            "-87.6316693"
        ]
    }
]
```

#### TODO:
- [x] Make entrances data opt-in
- [x] Implement other format outputs
- [x] Migration
- [x] Tests


lib-sql/functions/utils.sql
```diff
@@ -623,3 +623,30 @@ BEGIN
...
+DECLARE
+  entrance RECORD;
+BEGIN
+  IF osmtype != 'W' THEN
+    RETURN NULL;
+  END IF;
+
+  FOR entrance IN
+    SELECT ST_SetSRID(ST_MakePoint(lon / 1e7, lat / 1e7), 4326) as location
+      FROM planet_osm_nodes
```
lonvia: Alas, you cannot use the `planet_osm_nodes` table because it is not present when [flatnode store](https://nominatim.org/release-docs/develop/admin/Import/#flatnode-files) is used. The entrance data isn't stored in the place table either at the moment.

Two ways to work around this come to mind, a hacky way and the clean solution.

The hacky way would be to just add entrances to the [osm2pgsql lua style](https://nominatim.org/release-docs/develop/customize/Import-Styles/) and then filter entrances in [placex_insert](https://github.com/osm-search/Nominatim/blob/master/lib-sql/functions/placex_triggers.sql#L684) to prevent them from polluting the placex table. You could then join against the place table here.

The clean solution would be to create a special `planet_osm_entrances` table that contains lat/lon and tags for entrance nodes and query that one. It's been on the roadmap for a while to split up the huge `place` table into topic-specific sub tables as it would also simplify the triggers. This solution would require significantly more work because the helper functions for the osm2pgsql import style are currently not designed to distribute data over different tables. It would need some thinking how this can be done in a somewhat backward-compatible way.

lib-sql/functions/placex_triggers.sql
```diff
@@ -671,6 +671,7 @@ BEGIN
   NEW.indexed_status := 1; --STATUS_NEW
 
   NEW.centroid := get_center_point(NEW.geometry);
+  NEW.entrance := get_entrance(NEW.osm_type, NEW.osm_id);
```
lonvia: No need to compute that during insert already. `placex_update` will be called eventually and take care of filling the data.

lib-sql/functions/placex_triggers.sql
```diff
@@ -873,6 +874,7 @@ BEGIN
 
   -- Compute a preliminary centroid.
   NEW.centroid := get_center_point(NEW.geometry);
+  NEW.entrance := get_entrance(NEW.osm_type, NEW.osm_id);
```
lonvia: We should restrict the kinds of objects we compute entrances for. It is fairly expensive to compute. For example, right now it would try the algorithm on all the street ways as well. Lets find out what kind of objects have entrances and then we can decide what the best filter method is.

lonvia (commented):
Thanks for giving that a try. That's a lot less code than I expected and looks quite feasible to do.

So there are two major points that need some discussion here. One is the technical question where we get the entrance data from, see comment. The other one is about the design of the output.

OSM objects can have multiple entrances. Right now you are choosing one more or less at random. Nominatim could get some heuristic to cleverly chose the main entrance but the discussion in #2833 and #536 was already going in the direction that users might want to chose themselves. So we'd probably want a multivalue field. And in the output instaed of just returning lat/lon, have entrances be an array of objects with an extendable set of properties and a coordinate.

Technically speaking, we could just make the entrance column of JSONB type and then save all the entrance data. But if we go for the "clean solution" for finding entrances and have a planet_osm_entrance table, then entrances can be an array of OSM node IDs, which we just join against that entrance table.

emlove:
Thanks for the feedback! I'll give it another shot with some of these approaches.

emlove:
Pushed up a re-implementation! I updated the PR description.

src/nominatim_api/search/db_searches/base.py
```diff
@@ -47,7 +49,15 @@ def select_placex(t: SaFromClause) -> SaSelect:
...
-                     t.c.geometry.ST_Expand(0).label('bbox'))
+                     t.c.geometry.ST_Expand(0).label('bbox'),
+                     sa.select(sa.func.jsonb_agg(
+                         sa.func.jsonb_build_object(
+                             'osm_node_id', entrance_place.c.osm_id,
+                             'type', entrance_place.c.type,
+                             'lat', sa.func.ST_Y(entrance_place.c.geometry),
+                             'lon', sa.func.ST_X(entrance_place.c.geometry),
+                         )
+                     )).select_from(entrance_place).filter(sa.any_(t.c.entrance_osm_ids) == entrance_place.c.osm_id).label('entrances'))
```
lonvia: I should have been more explicit: the `place` table may also disappear, when the [database is frozen](https://nominatim.org/release-docs/develop/admin/Import/#dropping-data-required-for-dynamic-updates). That means in theory you would have to copy all the data over to `placex`. However, thinking more long term here: we eventually want to have a separate entrance source table and we wouldn't want to have to change the format of the placex table again.

So lets go with this solution for now and simply disable returning entrances for frozen database. Two things we should do for that:

* make querying of entrances here conditional to the existence of the place table. Please use [`get_cached_value()`](https://github.com/osm-search/Nominatim/blob/b7d77b9b43bc738b71f4b65b4ff1a69fade9c850/src/nominatim_api/connection.py#L103) to make sure that the existence of the table is only queried once.
* add a warning to the [documentation](https://nominatim.org/release-docs/develop/admin/Import/#dropping-data-required-for-dynamic-updates) that entrances won't work when the database is frozen

lib-sql/functions/utils.sql
```diff
@@ -623,3 +623,24 @@ BEGIN
...
+  node_ids BIGINT[];
+BEGIN
+  -- Get the list of OSM node IDs for the way
+  SELECT nodes FROM planet_osm_ways WHERE id=way_osm_id INTO node_ids;
+
+  -- Filter the list to only include places with the entrance class
+  SELECT array_agg(osm_id)
+    FROM (SELECT osm_id, class FROM place WHERE osm_id = ANY(node_ids))
+    WHERE class IN ('routing:entrance', 'entrance')
+    INTO node_ids;
```
lonvia: You can use a join to combine the two selects here.

This second part is going to need a special [partial index](https://www.postgresql.org/docs/current/indexes-partial.html) on place's osm_ids.

lib-sql/tables.sql
```diff
@@ -157,6 +157,7 @@ CREATE TABLE placex (
   country_code varchar(2),
   housenumber TEXT,
   postcode TEXT,
+  entrance_osm_ids BIGINT[] NOT NULL,
```
lonvia: `NOT NULL` is not ideal here. We'd want to use `NULL` to indicate something has no entrances because it takes significantly less space than an empty array.

lonvia:
One thing to add to the TODO list: the place table will eventually need a migration that adds the new column.

lib-sql/tables.sql
```diff
@@ -244,6 +244,23 @@ CREATE UNIQUE INDEX idx_postcode_id ON location_postcode USING BTREE (place_id)
...
 GRANT SELECT ON location_postcode TO "{{config.DATABASE_WEBUSER}}" ;
 
+-- Table to store location of entrance nodes
+DROP TABLE IF EXISTS place_entrance;
+CREATE TABLE place_entrance (
+  place_id BIGINT NOT NULL,
+  osm_node_id BIGINT NOT NULL,
+  type TEXT NOT NULL,
+  geometry GEOMETRY(Point, 4326) NOT NULL
+  );
```
lonvia, line 255: This may be  a stupid idea but if you structure the table like this instead:
```
   place_id BIGINT NOT NULL,
   entrance_info JSONB
```
and then store one line per place with the entrance information already processed in the way you need it in the API calls below, you might be able to avoid all the complications around json-compatibility for sqlite.

lonvia:
One more thought: if you consider the entrance data as an optional extra information similar to [address details](https://nominatim.org/release-docs/develop/api/Search/#output-details), then the user can choose if they need the extra information and you can greatly simply the code by implementing the query for the entrances only once in [add_result_details](https://github.com/osm-search/Nominatim/blob/master/src/nominatim_api/results.py#L451).

lib-sql/tables.sql
emlove: Yeah that's not a bad idea. We're not querying against the location data here anyway, so it might as well just live in a JSONB column. And yeah, this sqlite compatibility has grown into a mess.

emlove:
Alright, this is starting to look respectable!

lonvia:
Can you rebase on latest master? There are unfortunately some conflicts and I can't start the CI.

lib-sql/functions/placex_triggers.sql
```diff
@@ -874,6 +881,20 @@ BEGIN
   -- Compute a preliminary centroid.
   NEW.centroid := get_center_point(NEW.geometry);
 
+  -- Record the entrance node locations
+  IF NEW.osm_type = 'W' THEN
```
lonvia: The indexes might make this faster but you'd still would want to skip the 60 million highways if possible.

I would suggest a filter like this: `NEW.osm_type = 'W' and (NEW.rank_search > 27 or NEW.class IN ('landuse', 'leisure'))`. This will cover all POI-like objects (hotels, shops, zoos, etc) and some larger areas with possible entrances like parks and gated communities.

lib-sql/functions/placex_triggers.sql
```diff
@@ -874,6 +881,20 @@ BEGIN
...
+    SELECT jsonb_agg(jsonb_build_object('osm_id', osm_id, 'type', type, 'lat', ST_Y(geometry), 'lon', ST_X(geometry)))
```
lonvia: To be future-proof, I would suggest to also include the contents of 'extratags'. This might get us extra information later like "is this a public or private access point".

src/nominatim_api/results.py
```diff
@@ -466,6 +468,10 @@ async def add_result_details(conn: SearchConnection, results: List[BaseResultT],
             log().comment('Query parent places')
             for result in results:
                 await complete_parented_places(conn, result)
+        if details.entrances:
+            log().comment('Query entrances details')
+            for result in results:
+                await complete_entrances_details(conn, result)
```
lonvia: This is okay for now but eventually we are going to need to have a function here that sends only one SQL query for the complete set of result. `complete_address_details` is doing something along the way.

src/nominatim_api/types.py
```diff
@@ -548,3 +551,23 @@ def layer_enabled(self, layer: DataLayer) -> bool:
             true when layer restriction has been disabled completely.
         """
         return self.layers is None or bool(self.layers & layer)
+
+
+class EntranceDetails(NamedTuple):
```
lonvia: I'd prefer dataclasses over NamedTuple here, just for consistency with the other types.

src/nominatim_api/v1/format_xml.py
```diff
@@ -122,4 +122,12 @@ def format_base_xml(results: Union[ReverseResults, SearchResults],
                 for k, v in result.names.items():
                     ET.SubElement(eroot, 'name', attrib={'desc': k}).text = v
 
+        if options.get('entrances', False):
+            eroot = ET.SubElement(root if simple else place, 'entrances')
+            if result.entrances:
+                for entrance_detail in result.entrances:
+                    print(entrance_detail)
```
lonvia: leftover debug stuff

src/nominatim_api/v1/format.py
```diff
@@ -196,6 +196,9 @@ def _format_details_json(result: DetailedResult, options: Mapping[str, Any]) ->
         else:
             _add_address_rows(out, 'hierarchy', result.parented_rows, locales)
 
+    if result.entrances is not None:
+        out.keyval('entrances', [r._asdict() for r in result.entrances])
```
lonvia: Can also be left for later, but this should eventually use the json writer functions directly instead of converting into a dictionary first. Same for the outputs below, so this looks like it wants a entrance-json-output helper function.

src/nominatim_db/tools/migration.py
```diff
@@ -115,3 +116,27 @@ def create_postcode_parent_index(conn: Connection, **_: Any) -> None:
...
+  );
+CREATE UNIQUE INDEX idx_place_entrance_place_id ON place_entrance
+  USING BTREE (place_id) {{db.tablespace.search_index}};
+GRANT SELECT ON place_entrance TO "{{config.DATABASE_WEBUSER}}" ;
+
+-- Create an index on the place table for lookups to populate the entrance
+-- table
+CREATE INDEX IF NOT EXISTS idx_place_entrance_lookup ON place
+  USING BTREE (osm_id)
+  WHERE class IN ('routing:entrance', 'entrance');
```
lonvia: Nitpicking here: please indent by 8 spaces for better readability.

test/bdd/features/api/details/params.feature
```diff
@@ -30,6 +30,14 @@ Feature: Object details
         And the result is valid json
         And the result has attributes address
 
+    Scenario: Details with entrances
+        When sending v1/details
+          | osmtype | osmid     | entrances |
+          | W       | 429210603 | 1         |
+        Then a HTTP 200 is returned
+        And the result is valid json
+        And the result has attributes entrances
```
lonvia: I'd really like to have a check against the content here. That's going to be a bit tricky because there are no out-of-the-box checks for array attributes right now. It should be somehow possible to extend the ['result N contains in field'](https://github.com/osm-search/Nominatim/blob/master/test/bdd/conftest.py#L311) so that it can check all results of an array.

lib-sql/functions/placex_triggers.sql
emlove: Oh yeah, I was avoiding individual class matching because there ended up being quite a lot, but using the rank is clever.

test/bdd/features/db/update/entrances.feature
```diff
@@ -0,0 +1,82 @@
...
+          | 1  | 10,20,30,40 |
+        When importing
+        Then placex contains exactly
+         | object | place_id |
+         | W1     | 1        |
+        Then place_entrance contains exactly
+         | place_id | entrances |
+        When updating places
+          | osm  | class    | type  | geometry    |
+          | N10  | entrance | main  | 1           |
```
lonvia: Looking at these tests I realise that they may not automatically reflect correctly what actually happens during an update.

osm2pgsql has a mechanism where changes are propagated to dependent objects (see stage 1b in [osm2pgsql's how-it-works description](https://osm2pgsql.org/contribute/how-osm2pgsql-processing-works.html). That means that in practise a change of a node on a way like this will automatically mean that the way is updated as well.

The BDD tests will not automatically take these "forward dependencies" into account. You need to add an explicit "updating places" with the same content as lines 12-13 here.

The positive side is that the way will be presented to the `place_insert` trigger in all test cases here. It will just be dropped there because the trigger detects that nothing has changed. So you should get away with recomputing the entrance table content in `place_insert`.

test/bdd/features/db/update/entrances.feature
```diff
@@ -0,0 +1,82 @@
...
+    Scenario: A building with a newly tagged entrance
+        Given the grid
+          | 1 | 2 |
+          | 4 | 3 |
+        Given the places
+          | osm  | class    | type  | geometry    |
+          | W1   | building | yes   | (1,2,3,4,1) |
+        And the ways
+          | id | nodes       |
+          | 1  | 10,20,30,40 |
```
lonvia: You should use the same ids as in the grid here. The grid IDs do represent node IDs. Also, please close the building here as well, i.e. `1,2,3,4,1`.

lib-sql/tables.sql
```diff
@@ -244,6 +244,22 @@ CREATE UNIQUE INDEX idx_postcode_id ON location_postcode USING BTREE (place_id)
 CREATE INDEX idx_postcode_geometry ON location_postcode USING GIST (geometry) {{db.tablespace.address_index}};
 GRANT SELECT ON location_postcode TO "{{config.DATABASE_WEBUSER}}" ;
 
+-- Table to store location of entrance nodes
+DROP TABLE IF EXISTS place_entrance;
+CREATE TABLE place_entrance (
+  place_id BIGINT NOT NULL,
+  entrances JSONB NOT NULL
```
emlove: I wonder if at this point it makes sense to make this table more structured. Something like:
```sql
  place_id BIGINT NOT NULL,
  osm_id BIGINT NOT NULL,
  type TEXT NOT NULL,
  location GEOMETRY(Point, 4326) NOT NULL,
  extratags JSONB

Comment (lonvia) [reply lib-sql/tables.sql]: Possibly. It would very likely be more efficient to save on the PostgreSQL side. By moving the code that reads the table to the extra data section, you very likely no longer need the SQLite json function hell that prompted to switch to the compact table in the first place.

File: lib-sql/functions/utils.sql [hunk 19]
```diff
@@ -623,3 +623,31 @@ BEGIN
...
+  RETURNS INTEGER
+  AS $$
+DECLARE
+  entrance RECORD;
+  osm_ids BIGINT[];
+BEGIN
+  osm_ids := '{}';
+  FOR entrance in SELECT osm_id, type, geometry, extratags
+      FROM place
+      WHERE osm_id IN (SELECT unnest(nodes) FROM planet_osm_ways WHERE id=osmid) AND class IN ('routing:entrance', 'entrance') LOOP
```
lonvia: This is missing a `osm_type='N'` check.

lonvia:
Looks better now. Planet import running without errors so far. I'll report back once it is finished.

lib-lua/themes/nominatim/presets.lua
```diff
@@ -172,6 +172,10 @@ module.MAIN_TAGS_POIS = function (group)
                  no = group,
                  yes = group,
                  fire_hydrant = group},
+    entrance = {'always',
+                no = group},
+    ["routing:entrance"] = {'always',
```
lonvia: Lets do a little trick here and set this instead of 'always' to 

    exclude_when_key_present('entrance')

This will avoid that two rows are created when entrance and routing:entrance tags are both present in the same OSM object.

lib-sql/tables.sql
```diff
@@ -244,6 +244,25 @@ CREATE UNIQUE INDEX idx_postcode_id ON location_postcode USING BTREE (place_id)
 CREATE INDEX idx_postcode_geometry ON location_postcode USING GIST (geometry) {{db.tablespace.address_index}};
 GRANT SELECT ON location_postcode TO "{{config.DATABASE_WEBUSER}}" ;
 
+-- Table to store location of entrance nodes
+DROP TABLE IF EXISTS place_entrance;
+CREATE TABLE place_entrance (
```
lonvia: We are entering nitpicking territory here but I'd prefer to name this table `placex_entrance` now as it has come out as an extension to `placex` in the end.

src/nominatim_api/results.py
```diff
@@ -717,6 +722,28 @@ async def complete_linked_places(conn: SearchConnection, result: BaseResult) ->
         result.linked_rows.append(_result_row_to_address_row(row))
 
 
+async def complete_entrances_details(conn: SearchConnection, results: List[BaseResultT]) -> None:
+    """ Retrieve information about tagged entrances for this place.
+    """
+    t = conn.t.place_entrance
+    sql = sa.select(t.c.place_id, t.c.osm_id, t.c.type, t.c.location, t.c.extratags)\
+            .where(t.c.place_id.in_(r.place_id for r in results))
```
lonvia: You can filter results here for `result.source_table == SourceTable.PLACEX`. Results from other tables won't have entrances.

lonvia (commented):
Planet import done. There is no noticeable difference in import time. The entrance table has 200 MB for a total of 2.3 million entrances. `idx_place_entrance_lookup` is 95MB and `idx_place_entrance_place_id_osm_id` is 102MB. That's all within acceptable parameters.

I've added some final comments for small changes. It would also be good if you could squash the PR now into a handful of meaningful commits. And then this is good to go.

lib-sql/tables.sql
emlove: No worries. Better to get it right now before it gets rolled out everywhere.

lonvia:
Thanks for this nice PR. It was really a joy to review.

I'll roll out the change on the servers over at nominatim.openstreetmap.org right away but I won't backfill the entrance tables. So for the time being the entrance information will gradually emerge as changes to OSM are made. Might do a reimport once PostgreSQL 18 is out. Let see.

emlove:
Makes sense. Thanks for helping me through this! Also a big shout-out back for the incredible project docs. It helped out a ton spinning up on this! 💖

lonvia:
Turns out we have forgotten about documentation. Any chance you can add the new parameter to the [API documentation](https://nominatim.org/release-docs/latest/api/Overview/)? It would need to go into the 'Output details' section of each endpoint.

emlove:
> Turns out we have forgotten about documentation. Any chance you can add the new parameter to the [API documentation](https://nominatim.org/release-docs/latest/api/Overview/)? It would need to go into the 'Output details' section of each endpoint.

Ope. Yeah I can get that going tomorrow.

emlove:
https://github.com/osm-search/Nominatim/pull/3825

mint1goose:
Is it possible to expand the filter to include boundaries?
I added an entrance to https://nominatim.openstreetmap.org/ui/details.html?osmtype=W&osmid=1448345146&class=boundary
I could remove boundary:protected_area so that it would be a class of leisure (I believe?) but I am trying to follow https://wiki.openstreetmap.org/wiki/United_States/Public_lands#Tagging_of_IUCN_Protected_Areas
Tagged entrance node https://www.openstreetmap.org/node/13288475587#map=16/29.74502/-82.41343

PR #3813, opened 2025-08-13
Transliterate refactor WIP

anqixxx:


anqixxx:
Is there a way to add the new Python libraries as dependencies to github? I.e. https://pypi.org/project/cantonese-romanisation/ and https://pypi.org/project/Unidecode/ and https://pypi.org/project/OpenCC/

lonvia:
You need to add them in the [package description](https://github.com/osm-search/Nominatim/blob/196de9e974538ca6cd269a231cd769fbbb917406/packaging/nominatim-api/pyproject.toml#L17).

.gitignore
```diff
@@ -11,3 +11,4 @@ dist
 
 .vagrant
 data/country_osm_grid.sql.gz
+data/*
```
lonvia: This looks wrong. There are git-managed file in data/.

src/nominatim_api/localization/base.py
```diff
@@ -49,34 +41,60 @@ def _add_lang_tags(self, *tags: str) -> None:
                 self.name_tags.append(f"{tag}:{lang}")
                 self.name_tags.append(f"_place_{tag}:{lang}")
 
-    def display_name(self, names: Optional[Mapping[str, str]]) -> str:
+    def display_name(self, names: Optional[Dict[str, str]]) -> str:
```
lonvia: Is this change necessary?

src/nominatim_api/localization/base.py
```diff
@@ -49,34 +41,60 @@ def _add_lang_tags(self, *tags: str) -> None:
...
+            containing different name variants, as well as an identifier
+            with regards to what language used
+
+            If 'names' is null or empty, an empty tuple is returned. If no
+            appropriate localization is found, the first name is returned with
+            the 'default' marker, where afterwards iso is used, using country of origin.
+        """
         if not names:
-            return ''
+            return ("", "")
```
lonvia: Style nitpicking: please use simple quotes for simple strings. (Double quotes are reserved for format literals.)

(Also below in the strings for "default".)

src/nominatim_api/localization/base.py
```diff
@@ -49,34 +41,60 @@ def _add_lang_tags(self, *tags: str) -> None:
...
         if len(names) > 1:
             for tag in self.name_tags:
                 if tag in names:
-                    return names[tag]
+                    _, _, lang = tag.partition(':')
+                    return (names[tag], lang or "default")
 
         # Nothing? Return any of the other names as a default.
-        return next(iter(names.values()))
+        return (next(iter(names.values())), "default")
```
lonvia: You need to do the same partition trick here. The random choice of names can still contain a name tag.

src/nominatim_api/localization/transliterator.py
```diff
@@ -0,0 +1,211 @@
...
+
+            Only address parts that are marked as isaddress are localized.
+        """
+        if not result.address_rows:
+            return
+
+        local_languages = country_info.get_lang(str(result.country_code))
+        # would want to put cantonese here, i.e. use regions to detect
+        if len(local_languages) == 1:
+            result.region_lang = local_languages[0]
```
lonvia: This doesn't need to be a property of result. Just use a local variable here.

packaging/nominatim-api/pyproject.toml
```diff
@@ -19,7 +19,10 @@ dependencies = [
     "pyYAML>=5.1",
     "SQLAlchemy>=1.4.31",
     "psycopg",
-    "PyICU"
+    "PyICU",
+    "cantonese-romanisation",
+    "unidecode",
+    "opencc"
```
lonvia, line 30: These libraries should preferably be [optional dependencies](https://hatch.pypa.io/latest/config/metadata/#optional).

src/nominatim_db/data/lang_info.py
```diff
@@ -0,0 +1,79 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
```
lonvia: This file must be available in `nominatim_api` if it is to be used there. Move this file simply over there. For `country_info.py` create a smaller version that only contains the get functions you need.

test/python/api/test_normalization.py
```diff
@@ -0,0 +1,44 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
```
lonvia, line 1: File name is not ideal. `test_accept_language_parsing.py` or something along those lines would be better. Or just move the content into the transliterate_locales test below, which then would give us one test file testing all functionality of TransliterateLocales.

.gitignore
anqixxx: I'll revert this from the commited code, this was just an extra precaution to not commit my local files!

src/nominatim_api/localization/base.py
anqixxx: I think this was a byproduct of changes made from the locale refactor, will revert back to Mapping!

src/nominatim_api/localization/base.py
anqixxx: This code I believe is what is currently in the master branch! This was the logic used before, so I thought that this avoided the name tag issue, since the default is normally tag-less

https://github.com/osm-search/Nominatim/blob/196de9e974538ca6cd269a231cd769fbbb917406/src/nominatim_api/localization.py#L52

Issue #3810, opened 2025-08-06
Handling of braces when splitting names


lonvia:
Our [strip-braces sanitizer](https://github.com/osm-search/Nominatim/blob/master/src/nominatim_db/tokenizer/sanitizers/strip_brace_terms.py) creates variants for names with braces like "Halle (Saale)", assuming that the brace term is just an addendum.

The algorithm is doing the wrong thing when the brace term appears in the middle of the name. [Berlin (Ost) Hauptbahnhof](https://www.openstreetmap.org/node/3419894993) shouldn't create a variant "Berlin".

PR #3820, opened 2025-08-23
Sanetizer no longer strips name parts in brackets when more parts follow

mtmail:
fixes https://github.com/osm-search/Nominatim/issues/3810


lonvia:
Thanks!

PR #3825, opened 2025-09-02
Some docs for the entrances output

emlove:
I'd like to include an example in the outputs section if you happen to know a place that has entrances indexed on nominatim.openstreetmap.org.

See https://github.com/osm-search/Nominatim/pull/3807

### DIFF
--- a/docs/api/Output.md
+++ b/docs/api/Output.md
@@ -325,3 +325,62 @@ associated with the entrance node.
 
 They roughly correspond to the classification of the OpenStreetMap data
 according to either the `place` tag or the main key of the object.
+
+#### Example
+
+##### JSON
+
+[https://nominatim.openstreetmap.org/details?osmtype=W&osmid=32619803&entrances=1&format=json](https://nominatim.openstreetmap.org/details?osmtype=W&osmid=32619803&entrances=1&format=json)
+
+
+```json
+{
+  "place_id": 124325848,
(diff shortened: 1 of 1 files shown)


docs/api/Output.md
```diff
@@ -224,9 +230,9 @@ be more than one. The attributes of that element contain:
 When `addressdetails=1` is requested, the localised address parts appear
 as subelements with the type of the address part.
 
-Additional information requested with `extratags=1` and `namedetails=1` can
-be found in extra elements as sub-element of `extratags` and `namedetails`
-respectively.
+Additional information requested with `extratags=1`, `namedetails=1`, and
+`entrances=1` can be found in extra elements as sub-element of `extratags`,
+`namedetails`, and `entrance` respectively.
```
lonvia: Should be 'entrances'?

lonvia:
I missed the question for the example:

How about: https://nominatim.openstreetmap.org/ui/details.html?osmtype=W&osmid=32619803&class=shop It comes with four different types of entrances.

Or a simple one-entrance example: https://nominatim.openstreetmap.org/ui/details.html?osmtype=W&osmid=32259324&class=amenity

(Also opened https://github.com/osm-search/nominatim-ui/issues/300, so the entrances can actually be inspected in the UI.)

lonvia:
Thank you!

PR #3836, opened 2025-09-11
chore: update no,nb,nn country-names translation

Johannes-Andersen:
Update and add missing translations for: 
- Norwegian (no) 
- Norwegian Bokmål (nb) 
- Norwegian Nynorsk (NN)

Also removed nn and nb values where the "no" value was matching. Ref: https://wiki.openstreetmap.org/wiki/Multilingual_names#Norway

**Development process:**
- Initial batch done by a Node.JS script I wrote to read the yamls, then call the https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames API and add nb, no and nn
- Due to amount of mistakes in JS, I did a check against the norwegian lexicon https://snl.no/ for the norwegian name for listed countries.
- Finally I manually went over the changes and validated the names and made minor tweaks and adjustments.

### DIFF
--- a/settings/country-names/by.yaml
+++ b/settings/country-names/by.yaml
@@ -83,8 +83,7 @@ name:
     na: Berarut
     ne: बेलारुस
     nl: Wit-Rusland
-    nn: Kviterussland
-    "no": Hviterussland
+    "no": Belarus
     oc: Bielorussia
     or: ବେଲାଋଷ
     os: Белорусси
--- a/settings/country-names/fm.yaml
+++ b/settings/country-names/fm.yaml
@@ -31,7 +31,7 @@ name:
     lv: Mikronēzija
     mn: Микронези
     nl: Micronesia
-    "no": Mikronesia
+    "no": Mikronesiaføderasjonen
     oc: Estats Federats de Micronesia
     pl: Mikronezja
     ru: Федеративные Штаты Микронезии
--- a/settings/country-names/lc.yaml
+++ b/settings/country-names/lc.yaml
@@ -29,6 +29,7 @@ name:
     lv: Sentlūsija
     mn: Сент Люсиа
     nl: Saint Lucia
+    "no": Saint Lucia
     pl: Saint Lucia
     pt: Santa Lúcia
     ru: Сент-Люсия
--- a/settings/country-names/sh.yaml
+++ b/settings/country-names/sh.yaml
@@ -15,6 +15,7 @@ name:
     lt: Šventoji Elena, Dangun Žengimo ir Tristanas da Kunja
     mk: Света Елена, Успение и Тристан да Куња
     nl: Sint-Helena, Ascension en Tristan da Cunha
+    "no": Sankt Helena, Ascension og Tristan da Cunha
     pt: Santa Helena, Ascensão e Tristão da Cunha
     ru: Острова Святой Елены, Вознесения и Тристан-да-Кунья
     sk: Svätá Helena, Ascension a Tristan da Cunha
--- a/settings/country-names/vc.yaml
+++ b/settings/country-names/vc.yaml
@@ -30,7 +30,10 @@ name:
     li: Saint-Vincent
     lt: Sent Vinsentas ir Grenadinai
     mn: Сент-Винсент ба Гренадин
+    nb: Saint Vincent og Grenadinene
     nl: Saint Vincent en de Grenadines
+    nn: Saint Vincent og Grenadinane
+    "no": Saint Vincent og Grenadinene
     pl: Saint Vincent i Grenadyny
     pt: São Vicente e Granadinas
     ru: Сент-Винсент и Гренадины
--- a/settings/country-names/ad.yaml
+++ b/settings/country-names/ad.yaml
@@ -97,7 +97,6 @@ name:
     na: Andorra
     ne: एण्डोरा
     nl: Andorra
-    nn: Andorra
     "no": Andorra
     nv: Andówa
     oc: Andòrra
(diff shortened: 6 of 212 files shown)


settings/country-names/by.yaml
```diff
@@ -81,10 +81,11 @@ name:
     mt: Belarus
     my: ဘီလာရုဇ်နိုင်ငံ
     na: Berarut
+    nb: Belarus
     ne: बेलारुस
     nl: Wit-Rusland
-    nn: Kviterussland
-    "no": Hviterussland
+    nn: Belarus
```
Johannes-Andersen: https://www.nrk.no/nyheter/norge-endrer-navnet-pa-hviterussland-til-belarus-1.15983390

settings/country-names/fm.yaml
```diff
@@ -30,8 +30,10 @@ name:
     lt: Mikronezija
     lv: Mikronēzija
     mn: Микронези
+    nb: Mikronesiaføderasjonen
```
Johannes-Andersen: Mikronesia is the region,  Mikronesiaføderasjonen is the country.

lonvia:
Thanks for looking into keeping these names up to date.

For the country additions, see the comment in #3837.

Any keys "no" need to be surrounded by double qoutes or yaml will interpret them as "false". 

The rules around "nb" are apparently [complicated in OSM](https://wiki.openstreetmap.org/wiki/Multilingual_names#Norway). Basically it says to use 'name:nb'/'name:nn' only when the spellings differ, otherwise just use `name:no`. This seems a reasonable rule to follow. The result would rather mean removing a couple of the `name:nn` though.

Johannes-Andersen:
> For the country additions, see the comment in https://github.com/osm-search/Nominatim/issues/3837.

Agree! Removed them now 😁

> Any keys "no" need to be surrounded by double qoutes or yaml will interpret them as "false".

Fixed!

> The rules around "nb" are apparently [complicated in OSM](https://wiki.openstreetmap.org/wiki/Multilingual_names#Norway). Basically it says to use 'name:nb'/'name:nn' only when the spellings differ, otherwise just use name:no. This seems a reasonable rule to follow. The result would rather mean removing a couple of the name:nn though.

Hehe, yeah. nn, nb and no are always fun in every i18n project I have seen 😅

I could definitely update the configs to follow that schema! Just a quick question first; how does the fallback logic work? 
Example if someone where to just set nn or nb and not no, or if someone has they language set to nb and only nn is set

lonvia:
There currently is no fallback logic in Nominatim, somewhat under the assumptions that Norwegian speakers are aware that they better provide fallbacks in their language preferences. ;) I would take a PR to add a fallback. Somwhere around [here](https://github.com/osm-search/Nominatim/blob/2fb03cd1034224aa15d91a937bd3c3de5e75a8ee/src/nominatim_api/localization.py#L93) there should be there should be a special case to add 'no' if 'nn' or 'nb' appear without 'no' being in the list.

Johannes-Andersen:
> There currently is no fallback logic in Nominatim, somewhat under the assumptions that Norwegian speakers are aware that they better provide fallbacks in their language preferences. ;) I would take a PR to add a fallback. Somwhere around [here](https://github.com/osm-search/Nominatim/blob/2fb03cd1034224aa15d91a937bd3c3de5e75a8ee/src/nominatim_api/localization.py#L93) there should be there should be a special case to add 'no' if 'nn' or 'nb' appear without 'no' being in the list.

Sounds good!
Went ahead and ran a quick script to do some cleanup 😄

settings/country-names/lc.yaml
```diff
@@ -29,6 +29,7 @@ name:
     lv: Sentlūsija
     mn: Сент Люсиа
     nl: Saint Lucia
+    "no": St. Lucia
```
lonvia: Please use the unabbreviated for m here.

settings/country-names/sh.yaml
```diff
@@ -15,6 +15,7 @@ name:
     lt: Šventoji Elena, Dangun Žengimo ir Tristanas da Kunja
     mk: Света Елена, Успение и Тристан да Куња
     nl: Sint-Helena, Ascension en Tristan da Cunha
+    "no": St. Helena, Ascension og Tristan da Cunha
```
lonvia: dto. no abbreviations.

settings/country-names/vc.yaml
```diff
@@ -30,7 +30,10 @@ name:
     li: Saint-Vincent
     lt: Sent Vinsentas ir Grenadinai
     mn: Сент-Винсент ба Гренадин
+    nb: St. Vincent og Grenadinene
     nl: Saint Vincent en de Grenadines
+    nn: St. Vincent og Grenadinane
+    "no": St. Vincent og Grenadinene
```
lonvia: abbreviation

Johannes-Andersen:
> If Norwegian needs abbreviation forms like "saint" -> "st", then this would need to go into [`master`/settings/icu-rules/variants-no.yaml](https://github.com/osm-search/Nominatim/blob/master/settings/icu-rules/variants-no.yaml?rgh-link-date=2025-09-12T12%3A08%3A24Z) Something for another PR to look at.

Nice catch! Removed abbreviations 😄

> Also, please squash the commits into one and force push the result.

I think I already did that? 👀 Should just have been one commit I force pushed as I made changes.
Let me know if I need to change anything :D 

Thanks for the patience and the feedback!

settings/country-names/vc.yaml
```diff
@@ -30,7 +30,10 @@ name:
     li: Saint-Vincent
     lt: Sent Vinsentas ir Grenadinai
     mn: Сент-Винсент ба Гренадин
+    nb: St. Vincent og Grenadinene
     nl: Saint Vincent en de Grenadines
+    nn: St. Vincent og Grenadinane
```
lonvia: Sorry to be such a pain but there's still two abbreviations left here.

lonvia:
Thanks for patiently revising this PR. The CI failure is unrelated and already fixed on an upcoming branch of mine.

PR #3838, opened 2025-09-11
feat: add new countries

Johannes-Andersen:
This would need to be merged after https://github.com/osm-search/Nominatim/pull/3836.

I am not that familiar with how partition should be used, so just put them in a incremental order.
The data for language comes from https://wiki.openstreetmap.org/wiki/Nominatim/Country_Codes

Since I do not have time at the moment, none of the postalcode fiels are completed. So is blocked until I (or someone else) can finish it.

### DIFF
--- a/settings/country_settings.yaml
+++ b/settings/country_settings.yaml
@@ -1973,14 +1973,12 @@ bq:
     partition: 253
     languages: nl, pap, en
     names: !include country-names/bq.yaml
-    postcode: no
 
 # Bouvet Island (Bouvetøya)
 bv:
     partition: 254
     languages: no
     names: !include country-names/bv.yaml
-    postcode: no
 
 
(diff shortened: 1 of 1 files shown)


lonvia:
These countries have been removed intentionally in #2885 because they are not on country level in OSM and will never be used. Any particular reason you think they need to be brought back? If you have a custom installation with a different country base, then you can simply use a custom country name configuration as well.

Johannes-Andersen:
> These countries have been removed intentionally in #2885 because they are not on country level in OSM and will never be used. Any particular reason you think they need to be brought back? If you have a custom installation with a different country base, then you can simply use a custom country name configuration as well.

Hey! I can close this then :D
Just thought to contribute it upstream from my instance in case it could be useful. But as you mentioned, is not needed since it isn't going to be a country.

PR #3855, opened 2025-10-22
force layer to be address

hasandiwan:
appending the layer attribute and setting it to address allows building numbers to be shown,


lonvia:
The fact that Nominatim currently shows a garden instead of the address with that example URL is actually a bug that needs fixing.

That said, having an example of address-only reverse geocoding is probably a good idea.

PR #3862, opened 2025-10-30
Postcode sanetizer now skips values which are only zeros

mtmail:
followup to https://github.com/osm-search/Nominatim-Data-Analyser/pull/61#issuecomment-3456753621


src/nominatim_db/data/postcode_format.py
```diff
@@ -40,7 +43,10 @@ def match(self, postcode: str) -> Optional[Match[str]]:
         normalized = self.norm_pattern.fullmatch(postcode.upper())
 
         if normalized:
-            return self.pattern.fullmatch(normalized.group(1))
+            match = self.pattern.fullmatch(normalized.group(1))
+            if match and self.zero_pattern.match(match.string):
+                return None
```
lonvia: This looks wrong. If 'match' is truthy, then match should be returned.

PR #3874, opened 2025-11-10
Adapt type annotations to Falcon App type changes

vytas7:
See also: https://falcon.readthedocs.io/en/latest/api/typing.html#generic-app-types

### DIFF
--- a/src/nominatim_api/server/falcon/server.py
+++ b/src/nominatim_api/server/falcon/server.py
@@ -7,6 +7,8 @@
 """
 Server implementation using the falcon webserver framework.
 """
+from __future__ import annotations
+
 from typing import Optional, Mapping, Any, List, cast
 from pathlib import Path
 import asyncio


vytas7:
Ah, sorry, these errors in the BDD tests are introduced by my changes. One seems to install Falcon 3.1.1 from the Ubuntu repos which even predates the typing support in Falcon in general.

I can add `from __future__ import annotations` to the top of the file if that's OK?

lonvia:
Thanks a lot for the PR. That's really kind to help with changes in falcon.

So, the policy for Nominatim here is that pypy needs to work only with the latest release of the dependencies but the code needs to run with older versions.

I've usually solved this by quoting the types which only work with newer Python/falcon, e.g. [here](https://github.com/osm-search/Nominatim/blob/303ac42b479c93fac69160f9ca7c42bfcaa6da0c/src/nominatim_api/typing.py#L32). But using `__future__` is fine as well, if it doesn't complain when running against a falcon 3.x.

vytas7:
> Thanks a lot for the PR. That's really kind to help with changes in falcon.

You're welcome!

On that note (maybe it should have been a separate discussion), is it hard to set up the basic unit tests for the Falcon part of Nominatim?
We could potentially add them to our CI to make sure that we don't accidentally introduce breaking changes that affect downstream community projects. We already have such a CI gate for [GNU Mailman 3](https://github.com/falconry/falcon/blob/master/.github/workflows/test-mailman.yaml), but Nominatim would be interesting because, in contrast to Mailman, it exercises the ASGI (`async`) part of Falcon.

> I've usually solved this by quoting the types which only work with newer Python/falcon, e.g. [here](https://github.com/osm-search/Nominatim/blob/303ac42b479c93fac69160f9ca7c42bfcaa6da0c/src/nominatim_api/typing.py#L32). But using `__future__` is fine as well, if it doesn't complain when running against a falcon 3.x.

Using `__future__` should work fine, yes, as you can basically use any syntax you want until you actually try to evaluate the annotations. The aliases you referenced, on the other hand, unfortunately still need quoting, because they are not inside type annotations (unless they are under `TYPE_CHECKING`).

lonvia:
> On that note (maybe it should have been a separate discussion), is it hard to set up the basic unit tests for the Falcon part of Nominatim?

Not sure any of the tests we do are really useful for your context. Nominatim's tests work with the basic assumption that falcon is working as expected. The falcon bindings are only exercised in the [BDD tests](https://github.com/osm-search/Nominatim/tree/master/test/bdd/features/api) and even there we only use the [falcon test harness](https://github.com/osm-search/Nominatim/blob/e578c60ff48b66612e2d8f58d83aec46154e5926/test/bdd/utils/api_runner.py#L45) to speed up testing.

vytas7:
> Not sure any of the tests we do are really useful for your context. Nominatim's tests work with the basic assumption that **falcon is working as expected**. The falcon bindings are only exercised in the [BDD tests](https://github.com/osm-search/Nominatim/tree/master/test/bdd/features/api) and even there we only use the [falcon test harness](https://github.com/osm-search/Nominatim/blob/e578c60ff48b66612e2d8f58d83aec46154e5926/test/bdd/utils/api_runner.py#L45) to speed up testing.

Thanks for the quick walkthrough!
Well the idea of such tests is to ensure exactly this (emphasis mine) during [SemVer](https://falcon.readthedocs.io/en/stable/community/releases.html#semantic-versioning) minor releases (there cannot be any breaking changes in the API; does not apply to typing atm), however, looking at these tests, I see what you mean. But we'll still know there is an option.

Issue #3872, opened 2025-11-08
Entrance not found for boundary:protected_area


mint1goose:
## What did you search for?

https://nominatim.openstreetmap.org/ui/details.html?osmtype=W&osmid=1448345146

## What result did you get?

Returns center point of preserve, no entrance found despite tagged node at https://www.openstreetmap.org/node/13288475587#map=16/29.74502/-82.41343

## What result did you expect?

Should return the main entrance, as getting car directions to the center is meaningless.

## Further details

In the PR https://github.com/osm-search/Nominatim/pull/3807/commits/048d571e4618697eaa1301980503c0a0856ac0ae I noticed that there is a filter `NEW.osm_type = 'W' and (NEW.rank_search > 27 or NEW.class IN ('landuse', 'leisure'))` The filter should be expanded so that protected areas can have entrances but I don't know the best way to do this.
I could remove the boundary tag so that the class of the way would be leisure (I think?) But I am trying to follow these guidelines https://wiki.openstreetmap.org/wiki/United_States/Public_lands#Tagging_of_IUCN_Protected_Areas

lonvia:
`boundary` is a bit of a catch-all key. I'd be reluctant to add this to the filter. Probably the better solution here is to make sure that `leisure=nature_reserve` is used as the main class instead of `boundary=protected_area`. According to the wiki, the tag indicates "human usage", so more likely something with an entry.

mint1goose:
> `boundary` is a bit of a catch-all key. I'd be reluctant to add this to the filter. Probably the better solution here is to make sure that `leisure=nature_reserve` is used as the main class instead of `boundary=protected_area`. According to the wiki, the tag indicates "human usage", so more likely something with an entry.

Many parks in the US with trails/camping are an IUCN protected area with 'protect_class=*' that require a boundary tag. The wiki says that IUCN protected areas in the US must use:

- boundary=protected_area
- protect_class=*
- leisure=nature_reserve

Is it possible to tell Nominatim which main class to use between 'boundary' and 'leisure'? 
My understanding is that either IUCN areas should not have entrances tagged or the current IUCN areas with entrances should have 'boundary' and 'protect_class' removed. Or is the solution to have an unnamed relation with 'boundary' and 'protect_class' that contains an area of 'leisure'?

PR #3884, opened 2025-11-20
Allow any boundary to have an entrance

mint1goose:
There is no entrance for https://nominatim.openstreetmap.org/ui/details.html?osmtype=W&osmid=1448345146 because the search rank is less than 27 and this park is filtered out even though there is an entrance on the way. (The park is search rank 25)
More detail https://github.com/osm-search/Nominatim/issues/3872

It is also possible to set NEW.rank_search > 27 to be greater than 24, but I am not sure which option would be the most efficient.

### DIFF
--- a/lib-sql/functions/placex_triggers.sql
+++ b/lib-sql/functions/placex_triggers.sql
@@ -891,7 +891,7 @@ BEGIN
   NEW.centroid := get_center_point(NEW.geometry);
 
   -- Record the entrance node locations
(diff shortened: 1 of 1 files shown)


lonvia:
Thanks for giving this a try but as already mentioned in #3872 the issue with boundary is that it is a catch-all for anything remotely 'bounded'. I've now opened #3893 which will ensure that when an area is tagged with `boundary=protected_area` and `leisure=nature_reserve` that Nominatim will prefer the latter tag over the former for the main tag of the feature. This means that your area will fall under the `leisure` class and be checked for entrances. Please note that this will take only effect when data is modified in OSM.

PR #3892, opened 2025-11-29
Add success message to setup.log: related to #3891

daishu0000:
When the Nominatim import command is completed:

 `../nominatim-venv/bin/nominatim import --osm-file <your planet file> 2>&1 | tee setup.log`

a log entry will be added:

 `Import completed successfully in {elapsed:.2f} seconds.`

This makes it easier to track the import completion status in the logs.

### DIFF
--- a/src/nominatim_db/clicmd/setup.py
+++ b/src/nominatim_db/clicmd/setup.py
@@ -23,6 +23,7 @@ from ..tokenizer.base import AbstractTokenizer
 from ..version import NOMINATIM_VERSION
 from .args import NominatimArgs
 
+import time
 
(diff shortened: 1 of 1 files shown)


lonvia:
The CI fail is independent of your change. I'll fix that on master later. Can you please remove the latest test commit. Then this would be good to merge.

daishu0000:
> The CI fail is independent of your change. I'll fix that on master later. Can you please remove the latest test commit. Then this would be good to merge.


OK, I have removed the lastst test commit.

lonvia:
Thank you.

Issue #2463, opened 2021-10-09
Use search_name table for TIGER data imports on 'dropped' databases


lonvia:
When somebody has run the import with `--drop` or dropped the update information at a later point, it is no longer possible to import TIGER data because it relies on the `location_road_2` table. We could be using the `search_name`table instead in that case to look up the street. It will be quite a bit slower (maybe a temporary special index would help?) but at least it would work.

ChristophTito:
Hello, 
I am new to this project and would like to participate! As this is declared as a good first issue, I would like to try out if I can contribute something. Can you maybe provide some hints where to look first and some guidance how to tackle this problem?
Best regards,
Christoph

lonvia:
This concerns the implementation of the `nominatim add-data --tiger-data`.  The basic implementation can be found in[nominatim/tools/tiger_data.py](https://github.com/osm-search/Nominatim/blob/a7e048484b0f9066d1c6ba422087c2fd2c635b16/nominatim/tools/tiger_data.py#L78). However, to solve this issue, you have to dive a bit into the SQL functions. The main part of inserting the tiger data is done in [the tiger_line_import function](nominatim/tools/tiger_data.py). Starting in line 63 there are three functions used `getNearestNamedRoadPlaceId`, `getNearestParallelRoadFeature` and `getNearestRoadPlaceId`. These three functions determine the closest road that should be used as the parent road. These function can be replaced with equivalent queries of search_name (for `getNearestNamedRoadPlaceId`) and placex (for the other two functions which need a geometry).

pratham-saraf:
Hello 
As i am new here i thought i could start by contributing to good first issues. As per the information provided by lonvia the link to the tiger_line_import function is not functional so can it be re-provided so i can look at the guidance material provided
Thanks 
regards
Pratham Saraf

izzzabellay:
Hi, may I try to fix this issue as my first contribution on open-source project? 

I have successfully forked from the commit a7e0484 (the commit containing above-linked nominatim/tools/tiger_data.py) but met problems when trying to run CI test from that version. (claiming for no longer using version of Ubuntu, unable to install Centos-8, etc.)
So I tried to copy the ci_test.yml from the current master branch and ran the CI test, but again failed with the below summaries:
![image](https://github.com/user-attachments/assets/d2677291-1852-41d7-bffe-668ead91b1f1)

May I know what I can do to fix this?

[...]

PR #3900, opened 2025-12-06
wip: use search_name and placex for imports on dropped

Jamiller137:
- Replaced partitioned tables with `search_name` / `placex`.
  - `placex` queries are now filtered by `rank_address` 26–27 to align with the street rank.
  - Partition arguments are now unused.
- Commented out `is_frozen()` checks and related imports.
  - Dead code is not cleaned up yet.
- Adapted `test_add_tiger_data_database_frozen()` to test `--drop` imports
  - Renamed to `test_add_tiger_data_after_drop_updates()`

---

## Testing

Verified `make lint`, `make pytest`, and the recommended first steps in #2463:

```bash
./nominatim-cli.py import \
  --osm-file test/testdb/apidb-test-data.pbf \
  --no-updates

./nominatim-cli.py add-data \
  --tiger-data test/testdb/tiger/
```

now gives:

```plaintext
2025-12-05 23:21:43: Found 1 CSV files in path test/testdb/tiger/
......
2025-12-05 23:21:45: Creating indexes on Tiger data
```

### DIFF
--- a/src/nominatim_db/tools/freeze.py
+++ b/src/nominatim_db/tools/freeze.py
@@ -45,7 +45,7 @@ def drop_update_tables(conn: Connection) -> None:
     conn.commit()
 
 
-def _install_frozen_partition_functions(conn: Connection, config: Configuration) -> None:
+def install_frozen_partition_functions(conn: Connection, config: Configuration) -> None:
     """Frozen versions of partition-functions.sql"""
 
     frozen_functions_file = config.lib_dir.sql / 'functions' / 'frozen-db-functions.sql'
@@ -74,7 +74,7 @@ def freeze(config: Configuration) -> None:
     try:
         with connect(config.get_libpq_dsn()) as connection:
             drop_update_tables(connection)
-            _install_frozen_partition_functions(connection, config)
+            install_frozen_partition_functions(connection, config)
         drop_flatnode_file(config.get_path('FLATNODE_FILE'))
 
     except Exception as e:
--- a/src/nominatim_db/clicmd/setup.py
+++ b/src/nominatim_db/clicmd/setup.py
@@ -136,6 +136,7 @@ class SetupAll:
             if args.no_updates:
                 conn.autocommit = True
                 freeze.drop_update_tables(conn)
+                freeze.install_frozen_partition_functions(conn, args.config)
         tokenizer.finalize_import(args.config)
 
         LOG.warning('Recompute word counts')


lonvia:
I assume you are attacking #2463 here. The initial idea was to change only the TIGER import code but I really like your approach of changing the partition functions, so that even an `add-data` still works. The only drawback right now is that you are changing the functions globally for everything including the import. We don't want that. As long as the database isn't frozen, we'd want to use the partitioned helper tables.

So here is what I would suggest trying out: instead of modifying the partition functions in place, create a new file in `lib-sql/functions` that contains your modified partition functions. Then have the freeze function install your modified functions [around here](https://github.com/osm-search/Nominatim/blob/4cfc1792fb58f53a679ea9396cb5eb1d3b637082/src/nominatim_db/tools/freeze.py#L44). Subsequent calls to add tiger and other data then will automatically use the modified functions.

You'd also need to modifiy the tests for adding data and TIGER data to test against frozen and non-frozen databases.

AnderMiller:
@lonvia Thank you for the feedback and direction! It appears that #3906 also addresses this issue. Should this be closed? Otherwise, I am happy to implement these or any other changes you'd suggest.

lonvia:
As I said, I find your approach very interesting, but it's not clear if it will work out in the end. So I wouldn't mind if you both keep working on your respective approaches and we see in the end what works better. We will merge only one of the two eventually, of course, so I'll understand if you don't feel like continuing. For me, it's more the working on a PR the things to be learned that counts. If it gets merged or not, doesn't matter so much.

AnderMiller:
I still need to update tests but freeze now installs the modified partition functions. Currently, this does not handle `add-data` for databases imported with `--no-updates` without first freezing.

lib-sql/functions/frozen-db-functions.sql
```diff
@@ -0,0 +1,131 @@
+-- These functions query unified tables instead of partitions,
+-- which allows TIGER data imports on frozen databases.
+
+-- TODO: add reinstatiation of partition functions if needed
+DROP FUNCTION IF EXISTS getNearestNamedRoadPlaceId(integer, geometry, jsonb) CASCADE;
+DROP FUNCTION IF EXISTS find_road_with_postcode(integer, text) CASCADE;
+DROP FUNCTION IF EXISTS getNearestRoadPlaceId(integer, geometry) CASCADE;
+DROP FUNCTION IF EXISTS getAddressName(integer, geometry) CASCADE;
+DROP FUNCTION IF EXISTS getNearestParallelRoadFeature(integer, geometry) CASCADE;
```
lonvia, line 9: You don't need to explicitly drop the function, just always use 'CREATE OR REPLACE FUNCTION' below. Note that it will only work properly when the function signature is exactly the same. But that should be the case here.

lib-sql/functions/frozen-db-functions.sql
```diff
@@ -0,0 +1,131 @@
...
+
+  p1 := ST_LineInterpolatePoint(line, 0);
+  p2 := ST_LineInterpolatePoint(line, 0.5);
+  p3 := ST_LineInterpolatePoint(line, 1);
+
+  search_diameter := 0.0005;
+  WHILE search_diameter < 0.01 LOOP
+    FOR r IN
+      SELECT place_id FROM placex
+        WHERE ST_DWithin(line, geometry, search_diameter)
```
lonvia, line 36: You need to restrict to road feature here. This is done with: `rank_address between 26 and 27`.

lib-sql/functions/frozen-db-functions.sql
```diff
@@ -0,0 +1,131 @@
...
+  parent BIGINT;
+BEGIN
+  SELECT place_id INTO parent
+    FROM placex
+    WHERE postcode = postcode
+          AND rank_address between 26 and 27
+          AND class = 'highway'
+    ORDER BY ST_Distance(centroid, 
+                         (SELECT ST_Centroid(geometry) 
+                          FROM placex WHERE postcode = postcode LIMIT 1)) ASC LIMIT 1;
```
lonvia, line 89: Two issues with this query: there is no condition that goes to an index on placex, so it will scan over the entire 200mio entries. And postcodes are not in the placex table but in a separate table `location_postcode`.

So the better strategy here that solves both it to find the postcode in the location postcode table and then join on placex to find the nearest road for each result.

lib-sql/functions/frozen-db-functions.sql
```diff
@@ -0,0 +1,131 @@
...
+CREATE FUNCTION getNearestRoadPlaceId(in_partition INTEGER, point GEOMETRY)
+  RETURNS BIGINT AS $$
+DECLARE
+  parent BIGINT;
+BEGIN
+  SELECT place_id INTO parent
+    FROM placex
+    WHERE centroid && ST_Expand(point, 0.015)
+          AND rank_address between 26 and 27
+          AND class = 'highway'
```
lonvia, line 106: The additional class check is not necessary here. Ranks 26 and 27 are guaranteed to be roads.

src/nominatim_db/tools/freeze.py
```diff
@@ -44,6 +45,45 @@ def drop_update_tables(conn: Connection) -> None:
     conn.commit()
 
 
+def install_frozen_partition_functions(conn: Connection, config: Configuration) -> None:
+    """Frozen versions of partition-functions.sql"""
+
+    frozen_functions_file = config.lib_dir.sql / 'functions' / 'frozen-db-functions.sql'
```
lonvia, line 51: Simply use [`SQLPreprocessor.run_sql_file()`](https://github.com/osm-search/Nominatim/blob/58e56ec53d79e36632942b84822dbd27d875a427/src/nominatim_db/db/sql_preprocessor.py#L112) here.

src/nominatim_db/clicmd/freeze.py
```diff
@@ -29,10 +29,5 @@ def add_args(self, parser: argparse.ArgumentParser) -> None:
         pass  # No options
 
     def run(self, args: NominatimArgs) -> int:
-        from ..tools import freeze
```
lonvia, line 32: I know that function-level import are frowned upon but it is intentional here. If your Python IDE complains about it, please just silence the warning.

lib-sql/functions/frozen-db-functions.sql
```diff
@@ -0,0 +1,131 @@
...
+    FROM search_name
+    WHERE centroid && ST_Expand(point, 0.015)
+          AND rank_address between 26 and 27
+    ORDER BY ST_Distance(centroid, point) ASC LIMIT 1;
+
+  RETURN name;
+END
+$$
+LANGUAGE plpgsql STABLE PARALLEL SAFE;
+
```
lonvia, line 130: All the function in this file should also check that the partion ID in placex matches the input parameter.

lonvia (commented):
I've added a few preliminary review comments, I hope that is okay at this stage.

Here is an idea how you might have an easier time getting the SQL functions right: start on the master branch and replace the functions in `lib-sql/functions/partition-functions,sql` one by one with your implementation using placex. Every time you have rewritten a function, make sure that all tests still pass before going to the next. Once you have all replaced and all tests pass, move the function to your new file and reinstate the original implementation. That way you can be quite sure that the SQL code works correctly before working on the remaining Python code for replacing the functions.

lonvia:
I've now merged Ayush's PR as this one has been dormant for a bit. You are still welcome to follow up on this. Having the other add-data functions work in frozen mode is an interesting prospect.

PR #3901, opened 2025-12-08
Replace regex with `mwparserfromhell` based MW WikiCode Parsing for Special Phrases

### LINKED ISSUE #3829: Use full featured parser to parse special phrases wiki page
Our parser for special phrases wiki pager currently consists of a [simple regular expression](https://github.com/osm-search/Nominatim/blob/355cbcc7b866ac4ee6204467acb75492b3890c4a/src/nominatim_db/tools/special_phrases/sp_wiki_loader.py#L41). This is brittle and fails completely for tables that follow the one-cell-per-line approach. This should be replaced by a proper parser. [mwparserfromhell](https://github.com/earwig/mwparserfromhell) looks promising.

AyushDharDubey:
Closes #3829 

This PR replaces the fragile regular-expression-based parsing of the MediaWiki table used by the special phrases wiki loader with a `mwparserfromhell`-based parser. The change makes parsing robust and adds support for the cell-per-line table style.

* [x] Updated `test/testdata/special_phrases_test_content.txt` xml format with the latest [special export xml format](https://wiki.openstreetmap.org/wiki/Special:Export/Nominatim/Special_Phrases/EN) (into a proper row-per-line wiki table).
* [x] Updated test data to include a new example of the cell-per-line table format.

### DIFF
--- a/.github/actions/build-nominatim/action.yml
+++ b/.github/actions/build-nominatim/action.yml
@@ -22,7 +22,7 @@ runs:
 
         - name: Install prerequisites from apt
           run: |
-            sudo apt-get install -y -qq python3-icu python3-datrie python3-jinja2 python3-psutil python3-dotenv python3-yaml python3-sqlalchemy python3-psycopg python3-asyncpg
+            sudo apt-get install -y -qq python3-icu python3-datrie python3-jinja2 python3-psutil python3-dotenv python3-yaml python3-sqlalchemy python3-psycopg python3-asyncpg python3-mwparserfromhell
           shell: bash
           if: inputs.dependencies == 'apt'
 


lonvia:
This looks mostly good, thanks.

Can you please also add the new dependencies in the [Ubuntu installation instructions](https://github.com/osm-search/Nominatim/tree/master/vagrant) and in the [documentation](https://github.com/osm-search/Nominatim/blob/master/docs/admin/Installation.md#prerequisites). For Ubuntu, there should be a package `python3-mwparserfromhell` that can be used.

AyushDharDubey:
Thanks for the review!

Regarding the Ubuntu installation instructions, my understanding is that the existing [step here](https://github.com/osm-search/Nominatim/blob/681daeea293eec4bf6492cd85e4db609f09ff55f/vagrant/Install-on-Ubuntu-24.sh#L126) already pulls in the dependency from `packaging/nominatim-db/pyproject.toml`.

Please let me know if I’m missing something.

lonvia:
You are right, Ubuntu packages is not in the official instructions anymore. We do still test it, though. So the additional package needs to go [here in the CI](https://github.com/osm-search/Nominatim/blob/681daeea293eec4bf6492cd85e4db609f09ff55f/.github/actions/build-nominatim/action.yml#L25).

AyushDharDubey:
I’ve added it for now to keep things consistent with the other apt prerequisites. But it seems redundant to install these Python packages via apt as they’re already being installed as part of the [Nominatim build step](https://github.com/osm-search/Nominatim/blob/681daeea293eec4bf6492cd85e4db609f09ff55f/.github/actions/build-nominatim/action.yml#L42C1-L44C22).
I think we should omit the `Install prerequisites from apt` step from CI test too, as part of another PR?

lonvia:
The setup as is is intentionally done this way. The virtualenv allows system packages and the 'pip install' step will only install dependencies that are not available yet. So, if the packages get installed via apt first, then 'pip install' will pick up on the existing ones and just use those.

Thus, the Ubuntu-22 setup tests against the latest versions of dependencies from pypi, while the Ubuntu-24 runs with older versions of the dependencies. This gives a bit more insurance that Nominatim can run with different versions.

AyushDharDubey:
Oh, thanks for the insight.

PR #3906, opened 2025-12-11
Use `search_name` as fallback for TIGER imports when update tables are dropped

### LINKED ISSUE #2463: Use search_name table for TIGER data imports on 'dropped' databases

AyushDharDubey:
Closes #2463

This PR adds a fallback lookup that uses the persistent `search_name` table when the update information tables have been dropped. The existing lookup logic remains the primary path; only when an `undefined_table` exception is raised do we resolve the road via `search_name` instead.

The new `getNearestNamedRoadPlaceIdSlow` function leverages the existing `idx_search_name_name_vector` GIN index, which efficiently filters candidates by checking overlap between `name_vector` and the provided `token_info`, keeping the fallback reasonably fast (still slower than `getNearestNamedRoadPlaceId`).

### DIFF
--- a/lib-sql/tiger_import_start.sql
+++ b/lib-sql/tiger_import_start.sql
@@ -207,6 +207,7 @@ BEGIN
 
     RETURN 1;
   END IF;
+  RETURN 0;
 
 END;
 $$
--- a/lib-sql/tiger_import_start.sql
+++ b/lib-sql/tiger_import_start.sql
@@ -15,6 +15,99 @@ CREATE TABLE location_property_tiger_import (
   step SMALLINT,
   postcode TEXT);
 
+
+-- Lookup functions for tiger import when update 
+-- informations are dropped (see gh-issue #2463)
+CREATE OR REPLACE FUNCTION getNearestNamedRoadPlaceIdSlow(in_centroid GEOMETRY,
+                                                      in_token_info JSONB)
+  RETURNS BIGINT
+  AS $$
+DECLARE
+  out_place_id BIGINT;
+
+BEGIN
+  SELECT place_id INTO out_place_id 
+    FROM search_name
+    WHERE 
+        -- finds rows where name_vector shares elements with search tokens.
+        token_matches_street(in_token_info, name_vector)
+        -- limits search area
+        AND centroid && ST_Expand(in_centroid, 0.015) 
+        AND address_rank BETWEEN 26 AND 27
+    ORDER BY ST_Distance(centroid, in_centroid) ASC 
+    LIMIT 1;
+
+  RETURN out_place_id;
+END
+$$
+LANGUAGE plpgsql;
+
+
+CREATE OR REPLACE FUNCTION getNearestParallelRoadFeatureSlow(line GEOMETRY)
+  RETURNS BIGINT
+  AS $$
+DECLARE
+  r RECORD;
+  search_diameter FLOAT;
+  p1 GEOMETRY;
+  p2 GEOMETRY;
+  p3 GEOMETRY;
+
+BEGIN
+  IF ST_GeometryType(line) not in ('ST_LineString') THEN
+    RETURN NULL;
+  END IF;
+
+  p1 := ST_LineInterpolatePoint(line,0);
+  p2 := ST_LineInterpolatePoint(line,0.5);
+  p3 := ST_LineInterpolatePoint(line,1);
+
+    search_diameter := 0.0005;
+    WHILE search_diameter < 0.01 LOOP
+      FOR r IN
+        SELECT place_id FROM placex
+          WHERE ST_DWithin(line, geometry, search_diameter)
+          AND rank_address BETWEEN 26 AND 27
+          ORDER BY (ST_distance(geometry, p1)+
+                    ST_distance(geometry, p2)+
+                    ST_distance(geometry, p3)) ASC limit 1
+      LOOP
+        RETURN r.place_id;
+      END LOOP;
+      search_diameter := search_diameter * 2;
+    END LOOP;
+    RETURN NULL;
+END
+$$
+LANGUAGE plpgsql;
+
+
+CREATE OR REPLACE FUNCTION getNearestRoadPlaceIdSlow(point GEOMETRY)
+  RETURNS BIGINT
+  AS $$
+DECLARE
+  r RECORD;
+  search_diameter FLOAT;
+BEGIN
+    search_diameter := 0.00005;
+    WHILE search_diameter < 0.1 LOOP
+      FOR r IN
+        SELECT place_id FROM placex
+          WHERE ST_DWithin(geometry, point, search_diameter)
+          AND rank_address BETWEEN 26 AND 27
+          ORDER BY ST_Distance(geometry, point) ASC limit 1
+      LOOP
+        RETURN r.place_id;
+      END LOOP;
+      search_diameter := search_diameter * 2;
+    END LOOP;
+    RETURN NULL;
+END
+$$
+LANGUAGE plpgsql;
+
+
+-- Tiger import function
 CREATE OR REPLACE FUNCTION tiger_line_import(linegeo GEOMETRY, in_startnumber INTEGER,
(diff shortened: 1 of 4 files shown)


src/nominatim_db/clicmd/add_data.py
```diff
@@ -65,11 +63,6 @@ def add_args(self, parser: argparse.ArgumentParser) -> None:
     def run(self, args: NominatimArgs) -> int:
         from ..tools import add_osm_data
 
-        with connect(args.config.get_libpq_dsn()) as conn:
-            if is_frozen(conn):
-                print('Database is marked frozen. New data can\'t be added.')
-                return 1
```
lonvia, line 71: Your approach only works for Tiger data, so you can't completely get rid of this check. The function also imports other data which needs an unfrozen state.

lib-sql/tiger_import_start.sql
```diff
@@ -71,18 +71,29 @@ BEGIN
   place_centroid := ST_Centroid(linegeo);
   out_partition := get_partition('us');
 
-  out_parent_place_id := getNearestNamedRoadPlaceId(out_partition, place_centroid,
-                                                    token_info);
+  -- HYBRID LOOKUP LOGIC (see gh-issue #2463)
```
lonvia, line 167: You should work with template logic here instead of testing table presence: the SQLProcessor runs all SQL scripts through Jinja to allow simple preprocessing before executing the SQL. There are various variables at your disposal (grep for `{%` and `{{' in the 'lib-sql' directory, if you are interested, or see the variable setup  [here](https://github.com/osm-search/Nominatim/blob/58e56ec53d79e36632942b84822dbd27d875a427/src/nominatim_db/db/sql_preprocessor.py#L97)). The one of interest here is that you can do things depending on whether a table is present or not, see [this example](https://github.com/osm-search/Nominatim/blob/58e56ec53d79e36632942b84822dbd27d875a427/lib-sql/indices.sql#L90).

lib-sql/functions/tiger_lookup.sql
```diff
@@ -0,0 +1,52 @@
...
+-- This file is part of Nominatim. (https://nominatim.org)
+--
+-- Copyright (C) 2022 by the Nominatim developer community.
+-- For a full list of authors see the git log.
+
+-- Lookup functions for tiger import when update 
+-- informations are dropped (see gh-issue #2463)
+
+
+CREATE OR REPLACE FUNCTION lookup_road_in_search_name(
```
lonvia: No need for an extra file. Just move the function into tiger_import_start.sql, which is the only place were it is used.

lonvia (commented):
Thanks, that's the general direction I had in mind. I added a few inline comments,

There is a little gottcha here: the table `search_name` might not exist either, when Nominatim was set up in reverse-only mode. You could in that case fall back to doing name matching on placex entries or just tell the user that import is not possible.

lib-sql/tiger_import_start.sql
AyushDharDubey: That's much useful, thanks

AyushDharDubey:
Hello, thanks for the reviews. I’m still getting familiar with parts of the codebase. Going through the sql-preprocessor code was very helpful.

> the table `search_name` might not exist either, when Nominatim was set up in reverse-only mode. You could in that case fall back to doing name matching on placex entries or just tell the user that import is not possible.

based on my understanding, falling back to name matching on placex would be prohibitively slow for practically large databases, thus I’ve leaned towards explicitly reporting that the import is not possible in this configuration.

src/nominatim_db/tools/tiger_data.py
```diff
@@ -90,16 +89,18 @@ async def add_tiger_data(data_dir: str, config: Configuration, threads: int,
...
-            raise UsageError("Tiger cannot be imported when database frozen (Github issue #3048)")
-
     with TigerInput(data_dir) as tar:
         if not tar:
             return 1
 
         with connect(dsn) as conn:
             sql = SQLPreprocessor(conn, config)
+
+            if sql.env.globals['db']['reverse_only']:
```
lonvia: To be exact, the TIGER data can be imported in reverse_only as long as the database is not frozen. Only when frozen _and_ in reverse_only mode do we get in trouble.

Accessing the internals of the SQLPreprocessor is clever but might get us in trouble in the future. Better use a separate `table_exists('search_name')` here. It's a bit more expensive but this is a long running function anyway

lib-sql/tiger_import_start.sql
```diff
@@ -71,18 +117,34 @@ BEGIN
...
+    END IF;
+
+  -- When updatable information has been dropped:
+  -- Partition tables no longer exist, but search_name still persists.
+  {% elif 'search_name' in db.tables %}
+    RAISE WARNING 'Database frozen. Performing fallback tiger import lookup without partition tables';
+    -- Fallback: Look up in 'search_name' table 
+    -- though spatial lookups here can be slower.
+    out_parent_place_id := lookup_road_in_search_name(place_centroid, token_info);
+  {% else %}
```
lonvia: This else will never be reached (the Python code already checks for the condition), so I would suggest to just leave it out completely.

lib-sql/tiger_import_start.sql
```diff
@@ -15,6 +15,52 @@ CREATE TABLE location_property_tiger_import (
...
+  IF array_length(search_tokens, 1) IS NULL THEN
+    RETURN NULL;
+  END IF;
+
+
+  SELECT place_id INTO out_place_id 
+    FROM search_name
+    WHERE 
+        -- finds rows where name_vector shares elements with search tokens.
+        name_vector && search_tokens
```
lonvia: You should use the same comparison here as is used in getNearestNamedRoadPlaceId(): `token_matches_street(token_info, name_vector)`

lib-sql/tiger_import_start.sql
```diff
@@ -71,18 +117,31 @@ BEGIN
...
-  END IF;
+    IF out_parent_place_id IS NULL THEN
+      SELECT getNearestRoadPlaceId(out_partition, place_centroid)
+        INTO out_parent_place_id;
+    END IF;
+
+  -- When updatable information has been dropped:
+  -- Partition tables no longer exist, but search_name still persists.
+  {% elif 'search_name' in db.tables %}
+    RAISE WARNING 'Database frozen. Performing fallback tiger import lookup without partition tables';
```
lonvia: Please remove. This would create a lot of log entries for a full TIGER import.

lib-sql/tiger_import_start.sql
```diff
@@ -71,18 +117,31 @@ BEGIN
...
+    -- Fallback: Look up in 'search_name' table 
+    -- though spatial lookups here can be slower.
+    out_parent_place_id := lookup_road_in_search_name(place_centroid, token_info);
+  {% endif %}
 
 --insert street(line) into import table
 insert into location_property_tiger_import (linegeo, place_id, partition,
```
lonvia: While testing, I noticed that the insert should only be done, when a parent was found, i.e. when out_parent_place_id is not null. Without the parent Nominatim is not able to create an actual address from the line.

lonvia (commented):
I've tested this against a planet database. It's fairly fast but it leaves quite a few interpolations unmatched because the street name doesn't fit well enough. It would be good to have a replacement for getNearestParallelRoadFeture() as well which uses the placex table (again address_rank between 26 and 27 only) to find any parallel road.

AyushDharDubey:
Thanks for testing over planet db. I have added `placex` lookups for both parallel and nearest roads respectively, also squashed and force-pushed for cleaner commits.

lonvia:
This works now well on the full planet and finds parent streets for almost all entries. It's also reasonably fast, around 3 hours for the full Tiger dataset on an offline database.

Thanks for the work.

PR #3913, opened 2025-12-16
Reuse Configuration instance in Locales

### LINKED ISSUE #3909: API Performance issues and dotenv_values()
**Describe the bug**
The Nominatim API uses too much CPU. Around 12% of the total CPU time is being consumed by repeated calls to `dotenv.dotenv_values()`  in `config.py`, which seems unnecessary.

**To Reproduce**

(or rather, demonstrate)

1. Having previously set up a Nominatim DB with data for Monaco [with this PBF from Geofabrik](https://download.geofabrik.de/europe/monaco-latest.osm.pbf)
2. Modify `config.py` and replace the call to `dotenv_values()` by an empty dict `{}`, [here](https://github.com/osm-search/Nominatim/blob/58e56ec53d79e36632942b84822dbd27d875a427/src/nominatim_db/config.py#L67) and also comment out the `if` block [here](https://github.com/osm-search/Nominatim/blob/58e56ec53d79e36632942b84822dbd27d875a427/src/nominatim_db/config.py#L70). 
3. Pass the default values from [env.defaults](https://github.com/osm-search/Nominatim/blob/58e56ec53d79e36632942b84822dbd27d875a427/settings/env.defaults) as **real** environment variables instead.
4. Run the API with `gunicorn` (and uvicorn, falcon... just as in [the docs](https://nominatim.org/release-docs/latest/admin/Deployment-Python/) )
5. Execute any rudimentary synthetic benchmark against the /search endpoint, e.g. `ab -c1 -n1000 http://localhost/search?q=plaza`  
6. **Observe up to a 25% reduction of the time taken compared to the unmodified Python code.**

I have observed a similar improvement in our production environments. There, we do aprox. ~40% /reverse, ~%40 /details, ~20% /search. At a constant 50 requests per second the API requires 150% CPU (1.5 cores), while the same query rate only requires ~35% CPU (still high, but quite an improvement) with the `dotenv` calls removed from the code.

**Software Environment:**
- Nominatim version: 5.2
- Postgresql version: 17.6 (Debian 17.6-0+deb13u1)
- Postgis version: 3.5.2 dea6d0a
- OS: Linux 6.1.0-40-cloud-amd64

**Hardware Configuration:**
- RAM: 200GB+
- number of CPUs: 24+ (Xeon Icelake)
- type and size of disks: Enteprise SSDs 

(Tests also run on my dev computer, 64GB, 16/32 c/t Ryzen 9, very fast consumer SSD)

**Postgresql Configuration:**
Largely same as in the [docs](https://nominatim.org/release-docs/5.0/admin/Installation/#tuning-the-postgresql-database) for these tests. Postgres runs on separate hardware in our production environment.

**Nominatim Configuration:**
Only the database DSN is set through the system environment, **no .env file is added**. (NOTE: adding an empty .env does not improve the issue)

**Additional context**
We have been running many instances of Nominatim for the last few years. In late 2024 we tried to switch to the Python API, but decided to wait for a newer release since we could not reach acceptable response times when compared to the older PHP API. With 5.2, we are seeing considerable improvements on the actual query results (kudos! 🚀 ) so we are trying again to upgrade.

I understand that some performance loss was to be expected, as described in this thorough [blog post](https://nominatim.org/2023/04/03/reverse-performance.html). We are indeed seeing the same issues described in the post, perhaps even a bit worse in our use case: even slower response times, and considerably higher CPU usage compared to PHP.

The calls to `dotenv_values()` where identified using [py-spy](https://github.com/benfred/py-spy)

AyushDharDubey:
Closes #3909

This PR rewires the `Locales` constructor to accept an existing `Configuration` instance via a `config=` parameter, instead of instantiating a new (empty) configuration on each call.

The constructor remains forgiving by allowing `config=None` for call sites where a configuration object is not available in context [for example](https://github.com/osm-search/Nominatim/blob/master/src/nominatim_api/v1/format.py#L135).

By reusing a single `Configuration` instance where possible, this avoids repeated configuration loading and significantly reduces per-request overhead by over 25-30% per request.

### DIFF
--- a/src/nominatim_api/localization.py
+++ b/src/nominatim_api/localization.py
@@ -8,7 +8,6 @@
 Helper functions for localizing names of results.
 """
 from typing import Mapping, List, Optional
-from .config import Configuration
 from .results import AddressLines, BaseResultT
 
 import re
@@ -18,15 +17,15 @@ class Locales:
     """ Helper class for localization of names.
 
         It takes a list of language prefixes in their order of preferred
-        usage.
+        usage and comma separated name keys (Configuration.OUTPUT_NAMES).
     """
 
-    def __init__(self, langs: Optional[List[str]] = None):
-        self.config = Configuration(None)
+    def __init__(self, langs: Optional[List[str]] = None,
+                 names: str = 'name:XX,name') -> None:
         self.languages = langs or []
         self.name_tags: List[str] = []
 
-        parts = self.config.OUTPUT_NAMES.split(',')
+        parts = names.split(',') if names else []
 
         for part in parts:
             part = part.strip()
@@ -68,7 +67,7 @@ class Locales:
         return next(iter(names.values()))
 
     @staticmethod
-    def from_accept_languages(langstr: str) -> 'Locales':
+    def from_accept_languages(langstr: str, names: str = 'name:XX,name') -> 'Locales':
         """ Create a localization object from a language list in the
             format of HTTP accept-languages header.
 
@@ -96,7 +95,7 @@ class Locales:
             if len(parts) > 1 and all(c[0] != parts[0] for c in candidates):
                 languages.append(parts[0])
 
-        return Locales(languages)
+        return Locales(languages, names)
 
     def localize(self, lines: AddressLines) -> None:
         """ Sets the local name of address parts according to the chosen
--- a/src/nominatim_api/v1/server_glue.py
+++ b/src/nominatim_api/v1/server_glue.py
@@ -174,7 +174,8 @@ async def details_endpoint(api: NominatimAPIAsync, params: ASGIAdaptor) -> Any:
     if result is None:
         params.raise_error('No place with that OSM ID found.', status=404)
 
-    locales = Locales.from_accept_languages(get_accepted_languages(params))
+    locales = Locales.from_accept_languages(get_accepted_languages(params),
+                                            params.config().OUTPUT_NAMES)
     locales.localize_results([result])
 
     output = params.formatting().format_result(
@@ -215,8 +216,8 @@ async def reverse_endpoint(api: NominatimAPIAsync, params: ASGIAdaptor) -> Any:
         query = ''
 
     if result:
-        Locales.from_accept_languages(get_accepted_languages(params)).localize_results(
-            [result])
+        Locales.from_accept_languages(get_accepted_languages(params),
+                                      params.config().OUTPUT_NAMES).localize_results([result])
 
     fmt_options = {'query': query,
                    'extratags': params.get_bool('extratags', False),
@@ -255,7 +256,8 @@ async def lookup_endpoint(api: NominatimAPIAsync, params: ASGIAdaptor) -> Any:
     if debug:
         return build_response(params, loglib.get_and_disable(), num_results=len(results))
 
-    Locales.from_accept_languages(get_accepted_languages(params)).localize_results(results)
+    Locales.from_accept_languages(get_accepted_languages(params),
+                                  params.config().OUTPUT_NAMES).localize_results(results)
 
     fmt_options = {'extratags': params.get_bool('extratags', False),
                    'namedetails': params.get_bool('namedetails', False),
@@ -348,7 +350,8 @@ async def search_endpoint(api: NominatimAPIAsync, params: ASGIAdaptor) -> Any:
     except UsageError as err:
         params.raise_error(str(err))
 
-    Locales.from_accept_languages(get_accepted_languages(params)).localize_results(results)
+    Locales.from_accept_languages(get_accepted_languages(params),
+                                  params.config().OUTPUT_NAMES).localize_results(results)
 
     if details['dedupe'] and len(results) > 1:
         results = helpers.deduplicate_results(results, max_results)
--- a/src/nominatim_db/clicmd/api.py
+++ b/src/nominatim_db/clicmd/api.py
@@ -19,6 +19,7 @@ import nominatim_api as napi
 from nominatim_api.v1.helpers import zoom_to_rank, deduplicate_results
 from nominatim_api.server.content_types import CONTENT_JSON
 import nominatim_api.logging as loglib
+from ..config import Configuration
 from ..errors import UsageError
 from .args import NominatimArgs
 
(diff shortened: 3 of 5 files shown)


src/nominatim_api/localization.py
```diff
@@ -21,8 +21,12 @@ class Locales:
...
     """
 
-    def __init__(self, langs: Optional[List[str]] = None):
-        self.config = Configuration(None)
+    def __init__(
+            self,
+            langs: Optional[List[str]] = None,
+            config: Optional[Configuration] = None
+    ) -> None:
+        self.config = config or Configuration(None)
```
lonvia: Looking at this, it might be better to hand in only the contents of config.OUTPUT_NAMES instead of the full config object. And when it is not set, then simply fall back to the most minimalistic value "name". There shouldn't be a relevant part in the code, where it is not possible to set the value correctly.

src/nominatim_api/localization.py
```diff
@@ -68,7 +72,7 @@ def display_name(self, names: Optional[Mapping[str, str]]) -> str:
         return next(iter(names.values()))
 
     @staticmethod
-    def from_accept_languages(langstr: str) -> 'Locales':
+    def from_accept_languages(langstr: str, config: Optional[Configuration] = None) -> 'Locales':
```
lonvia: Here it would be fine to leave handing in the whole config as long as it is mandatory.

src/nominatim_api/v1/server_glue.py
```diff
@@ -215,8 +215,12 @@ async def reverse_endpoint(api: NominatimAPIAsync, params: ASGIAdaptor) -> Any:
         query = ''
 
     if result:
-        Locales.from_accept_languages(get_accepted_languages(params)).localize_results(
-            [result])
+        Locales.from_accept_languages(
+            get_accepted_languages(params),
+            params.config()
+        ).localize_results([result])
```
lonvia: Just to get this general observation out of the way: It's not exactly forbidden but I'm not very fond of this kind of line-breaking style. Whereever the space permits, I prefer something like:

```
        Locales.from_accept_languages(get_accepted_languages(params),
                                      params.config())
```
In cases where functions are chained like this, I usually went for:

```
Locales.from_accept_languages()\
       .localize_results()
```

Maybe it's possible to configure your editor to slightly change the autoformatting to get closer to this?

src/nominatim_db/clicmd/api.py
```diff
@@ -91,15 +92,15 @@ def _get_geometry_output(args: NominatimArgs) -> napi.GeometryFormat:
...
+def _get_locales(args: NominatimArgs, config: Optional[Configuration]) -> napi.Locales:
     """ Get the locales from the language parameter.
     """
     if args.lang:
-        return napi.Locales.from_accept_languages(args.lang)
-    if default:
-        return napi.Locales.from_accept_languages(default)
+        return napi.Locales.from_accept_languages(args.lang, config)
+    if config:
+        return napi.Locales.from_accept_languages(config.DEFAULT_LANGUAGE, config)
```
lonvia: Interesting. So it would make sense to make the first parameter of `from_accept_languages()` optional and set it to config.DEFAULT_LANGUAGE if not given.

lonvia (commented):
> The constructor remains forgiving by allowing config=None for call sites where a configuration object is not available in context [for example](https://github.com/osm-search/Nominatim/blob/master/src/nominatim_api/v1/format.py#L135).

Looking at the line you cite here:

https://github.com/osm-search/Nominatim/blob/5e965d521660304c2e89ff29e4e12a5edb648c1d/src/nominatim_api/v1/format.py#L135

I notice it was badly written. `Locale()` will always be instantiated here, even when it is available in the options dict. This should be: `locales = options.get('locales') or Locales()`.

And I don't think the code path where `Locales()` is instantiated is actually reachable afterwards.

src/nominatim_api/v1/server_glue.py
AyushDharDubey: sure, thanks for pointing out.

src/nominatim_api/localization.py
AyushDharDubey: Yes, that would remove the fallback call for Configuration too. There doesn't seem to be any drawbacks in doing so.

src/nominatim_api/localization.py
AyushDharDubey: We don't have any special use case for config in `from_accept_languages` too, so we can change this one also to `config.OUTPUT_NAMES` instead.

src/nominatim_db/clicmd/api.py
AyushDharDubey: Instead, if we are going with getting rid of config, we can leave it as it is and just rely on the calling site to pass in the `config.DEFAULT_LANGUAGE`

src/nominatim_api/localization.py
```diff
@@ -18,15 +18,17 @@ class Locales:
...
-        usage.
+        usage and comma separated name keys (Configuration.OUTPUT_NAMES).
     """
 
-    def __init__(self, langs: Optional[List[str]] = None):
-        self.config = Configuration(None)
+    def __init__(self, langs: Optional[List[str]] = None,
+                 names: Optional[str] = None) -> None:
+        if names is None:
+            names = Configuration(None).OUTPUT_NAMES
```
lonvia: Let's remove this fallback, too. The outcome would still be mildly surprising  as this only catches the system default. Also, it removes the dependency between configuration.py and localization.py, which is a good thing. Just set the preset value in line 25 to `names: str = 'name'`

src/nominatim_db/clicmd/api.py
```diff
@@ -91,13 +92,18 @@ def _get_geometry_output(args: NominatimArgs) -> napi.GeometryFormat:
         raise UsageError(f"Unknown polygon output format '{args.polygon_output}'.") from exp
 
 
-def _get_locales(args: NominatimArgs, default: Optional[str]) -> napi.Locales:
+def _get_locales(args: NominatimArgs, config: Optional[Configuration]) -> napi.Locales:
```
lonvia: config is always handed in, so don't make it Optional and simplify the code below accordingly.

src/nominatim_api/localization.py
AyushDharDubey: sure

src/nominatim_api/localization.py
```diff
@@ -18,15 +17,15 @@ class Locales:
...
 
         It takes a list of language prefixes in their order of preferred
-        usage.
+        usage and comma separated name keys (Configuration.OUTPUT_NAMES).
     """
 
-    def __init__(self, langs: Optional[List[str]] = None):
-        self.config = Configuration(None)
+    def __init__(self, langs: Optional[List[str]] = None,
+                 names: Optional[str] = 'name:XX,name') -> None:
```
lonvia: Tiny nitpick: doesn't need to be of type 'Optional[str]' anymore, can now be 'str'. (same below)

lonvia:
Thank you!

Issue #3912, opened 2025-12-16
Typo: Locales is inadvertently called as Locale


AyushDharDubey:
In [Getting Started](https://github.com/osm-search/Nominatim/blob/master/docs/library/Getting-Started.md#presenting-results-to-humans) docs: at some places there are calls to helper class `Locale`, this was intended to be [Locales](https://github.com/osm-search/Nominatim/blob/master/src/nominatim_api/localization.py#L17).

28Deepakpandey:
If no one else is working on it, I would like to take this up.

PR #3917, opened 2025-12-18
Fix typo: Locale → Locales in Getting Started docs

28Deepakpandey:
I corrected a documentation error in the Getting Started guide. The helper class name was mistakenly written as Locale instead of Locales in 3–4 places. This update ensures consistency with the actual class name and avoids confusion for new users following the docs.
These updates address the issue mentioned in https://github.com/osm-search/Nominatim/issues/3912 .

### DIFF
--- a/docs/customize/Result-Formatting.md
+++ b/docs/customize/Result-Formatting.md
@@ -229,7 +229,7 @@ _None._
 
 | Option          | Description |
 |-----------------|-------------|
-| locales         | [Locale](../library/Result-Handling.md#locale) object for the requested language(s) |
+| locales         | [Locales](../library/Result-Handling.md#locales) object for the requested language(s) |
 | group_hierarchy | Setting of [group_hierarchy](../api/Details.md#output-details) parameter |
 | icon_base_url   | (optional) URL pointing to icons as set in [NOMINATIM_MAPICON_URL](Settings.md#nominatim_mapicon_url) |
 
--- a/docs/library/Getting-Started.md
+++ b/docs/library/Getting-Started.md
@@ -248,19 +248,19 @@ of the result. To do that, you first need to decide in which language the
 results should be presented. As with the names in the result itself, the
 places in `address_rows` contain all possible name translation for each row.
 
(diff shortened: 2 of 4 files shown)


28Deepakpandey:
Hi @lonvia , I’ve corrected the remaining instances of Locale → Locales in the following files, as suggested:
docs/library/Getting-Started.md
docs/library/Result-Handling.md
docs/customize/Result-Formatting.md

docs/library/Getting-Started.md
```diff
@@ -248,19 +248,19 @@ of the result. To do that, you first need to decide in which language the
...
-locale = napi.Locale(['fr', 'en'])
+locale = napi.Locales(['fr', 'en'])
 ```
 
 creates a helper class that returns the name preferably in French. If that is
 not possible, it tries English and eventually falls back to the default `name`
 or `ref`.
 
-The `Locale` object can be applied to a name dictionary to return the best-matching
+The `locale` object can be applied to a name dictionary to return the best-matching
```
lonvia: Should start with an upper case letter.

docs/customize/Result-Formatting.md
```diff
@@ -229,7 +229,7 @@ _None._
 
 | Option          | Description |
 |-----------------|-------------|
-| locales         | [Locale](../library/Result-Handling.md#locale) object for the requested language(s) |
+| Locales         | [Locales](../library/Result-Handling.md#Locales) object for the requested language(s) |
```
lonvia: Lower-case was correct for the option name, see [here](https://github.com/osm-search/Nominatim/blob/5e965d521660304c2e89ff29e4e12a5edb648c1d/src/nominatim_api/v1/server_glue.py#L182). Same for URL, the fragments are lower-cased.

lonvia:
Something went wrong with this rebase. It now contains a mix of my and your changes. I'm not quite sure how this happened. It might be easier at this point to close this PR and start afresh. Sorry for the inconvenience. git can be difficult to get a handle on at the beginning.

Issue #3920, opened 2025-12-26
Replace use of eval() in BDD tests with JSON parsing


Aditya30ag:
## Is your feature request related to a problem? Please describe.

The test helper logic currently uses Python’s `eval()` to parse values coming from BDD feature files (e.g. in `place_inserter.py`, method `_add`). Because these inputs originate from test data strings, using `eval()` allows **arbitrary code execution**.

Example of the current implementation:

```python
self.columns[key] = eval('{' + value + '}')
```

also in checks.py file

These usages should be replaced with json parsing, see also comment below.

**Note:** _make sure to read and understand the [contribution guide](https://github.com/osm-search/Nominatim/blob/master/CONTRIBUTING.md) before starting to work on this issue. In particular, carefully read the part about using AI tools._

mtmail:
The line containing `eval` is called five times in the test suite, twice without any value:

```
test/bdd/features/db/import/entrances.feature
eval called with value
eval called with value 'wheelchair': 'yes'
eval called with value
test/bdd/features/db/import/rank_computation.feature
eval called with value "place": "city"
eval called with value "amenity": "playground"
```

What makes this critical? The values are hardcoded in `.feature` files. Is the assumption an attacker would try to submit a PR with malicious `.feature` files? It can't be executed if Nominatim is running as a service (CLI, API, webservice) and the test suite isn't loading external files an attacker would have control over.

When creating a Github issue, it asks what kind of issue it is. See https://github.com/osm-search/Nominatim/issues/new/choose. One option in the list is "Report a Security vulnerability" linking to https://github.com/osm-search/Nominatim/security/policy

Aditya30ag:
Thanks for the explanation, @mtmail 

I agree this isn’t a production or remotely exploitable issue since the values are hardcoded in test .feature files. My concern was mainly about avoiding eval() as a risky pattern and keeping test helpers safe by default, rather than an immediate security threat.

I’m fine to treat this as a code-quality / hardening issue and follow whatever approach the maintainers prefer.

lonvia:
If you feel like improving this: use a json parser instead of eval (keep the part where the outer brackets are added) and adapt the tests where necessary.

Aditya30ag:
Done Please check

Itz-Agasta:
Hi @lonvia its a pretty straight forward patch... i would like to work on it... can you assign it to me?

mtmail:
@Itz-Agasta Thanks. Feel free to work on it.

[...]

PR #3921, opened 2025-12-26
Replace unsafe eval() with json parser

Aditya30ag:
## Remove unsafe eval() usage from test helpers
"  --body "Replaces eval() with ast.literal_eval and a safe fallback parser. 

Adds unit tests to ensure malicious inputs are rejected and existing BDD parsing continues to work.

#3920 Closing issue

### DIFF
--- a/test/bdd/utils/checks.py
+++ b/test/bdd/utils/checks.py
@@ -58,7 +58,7 @@ COMPARISON_FUNCS = {
     None: lambda val, exp: str(val) == exp,
     'i': lambda val, exp: str(val).lower() == exp.lower(),
     'fm': lambda val, exp: re.fullmatch(exp, val) is not None,
-    'dict': lambda val, exp: val is None if exp == '-' else (val == eval('{' + exp + '}')),
+    'dict': lambda val, exp: val is None if exp == '-' else (val == json.loads('{' + exp + '}')),
     'in_box': within_box
 }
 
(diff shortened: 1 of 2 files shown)


mtmail:
Given the `eval()` is only executed with 3 different values in the whole test suite there might be easier work-arounds. Comma separated list, or treating the string values as JSON and using a JSON parser, for example. Your solution is good because it's generic but I'd favor something less than 40 lines of code. We'll review and discuss internally. For now don't spend more time on this PR.

```
test/bdd/features/db/import/entrances.feature
eval called with value
eval called with value 'wheelchair': 'yes'
eval called with value
test/bdd/features/db/import/rank_computation.feature
eval called with value "place": "city"
eval called with value "amenity": "playground"
```

Aditya30ag:
Thanks for the review and the feedback, @mtmail 
I agree that for the current usage (the few fixed values in the test suite), a simpler approach .
Happy to pause work on this PR as suggested. If you decide internally that a simpler implementation is preferred, I can rework this accordingly.

Thanks again for taking a look.

lonvia:
This looks about right. You still need to adapt the BDDs so that they contain valid json. For most part that means replacing single quotes with double quoting.

lonvia:
I'm closing this as it suspiciously looks like the poster is using AI without bothering to get any actual knowledge about the project. Shame really, as this PR could have been a useful contribution to the project.

Issue #3598, opened 2024-11-21
Appropriate asyncio.set_event_loop_policy to make scripts usable on Windows


rpahut:
**Is your feature request related to a problem? Please describe.**
Running `nominatim-cli.py import` on Windows eventually runs into an exception at `psycopg.AsyncConnection.connect` call:
`Psycopg cannot use the 'ProactorEventLoop' to run in async mode. Please use a compatible event loop, for instance by setting 'asyncio.set_event_loop_policy(WindowsSelectorEventLoopPolicy())'`

**Describe the solution you'd like**
The [official solution](https://www.psycopg.org/psycopg3/docs/advanced/async.html#async) to this exception is to set the event loop compatible with the OS some time before the async work:
```
asyncio.set_event_loop_policy(
    asyncio.WindowsSelectorEventLoopPolicy()
)
```
This removes the exception, and as far as I can tell there are no other fatal issues preventing the script from being used on Windows.

**Describe alternatives you've considered**
Simple ad-hoc modification to the nominatim-cli.py works, but `set_event_loop_policy` can be called almost anywhere before the `connect` and does not have to be a part of nominatim scripts.

lonvia:
We could add an appropriate line but I have no way to test if that actually works. Anybody interested to set up CI testing on a Windows machine?

kad-link:
Hi @lonvia !  I would like to work on this issue.
My plan is to fix the Windows async crash by setting `asyncio.WindowsSelectorEventLoopPolicy()` at process startup, guarded behind a Windows-only check, before any async psycopg connections are created.

I don't have a local Windows machine but I plan to validate the fix using GitHub Actions( `windows-latest` ). 

I will keep the PR small and focused on solving only this issue, with CI and install workflows (#3903 ) in separate PRs later. 
If this approach sounds acceptable, I will proceed and open a PR fixing #3598. 
Happy to adjust based on feedback.

kad-link:
> We could add an appropriate line but I have no way to test if that actually works. Anybody interested to set up CI testing on a Windows machine?

Since #3598 requires Windows verification, I will start by adding a Windows GitHub Actions workflow (`windows-latest`). 
Initially , I plan to do a smoke-level CI that: 

- [ ] checks out the project
- [ ] installs Python dependencies
- [ ] runs a basic CLI startup command

I think this should give us a reliable way to confirm Windows behavior going forward. 

[...]

PR #3923, opened 2025-12-30
CI: add dev-only Windows smoke workflow

kad-link:
## Description
This PR adds an experimental Windows CI workflow to provide a lightweight signal for Windows contributors. 

- Adds `windows-ci.yml` running on `windows-latest` for push and pull request
- Runs `flake8` on Windows (same lint target as Linux CI)
- Runs `python -m compileall` as a smoke check to catch syntax or import-time issues
- Marked as non-blocking via `continue-on-error: true`

### Note
This workflow is intentionally dev-only and does not attempt a full nominatim-db install on Windows (blocked by PyICU/ICU native build requirements on GitHub-hosted runners). Linux CI remains the authoritative test pipeline.

## Future Work
This is the first step toward full Windows support. I am planning a follow-up effort to resolve the PyICU / ICU native build issues so we can eventually run full Nominatim installs and tests on Windows CI.


lonvia:
That's a nice start but we'll need a full run of Nominatim's tests to make sure that the installation actually works. At a minimum the [tests workflow](https://github.com/osm-search/Nominatim/blob/20a333dd9b3649699b2e5269cda95c4f30b00aae/.github/workflows/ci-tests.yml#L36) should work just like on the Linux installations. Ideally, we get a [install workflow](https://github.com/osm-search/Nominatim/blob/20a333dd9b3649699b2e5269cda95c4f30b00aae/.github/workflows/ci-tests.yml#L143) to work as well.

kad-link:
> That's a nice start but we'll need a full run of Nominatim's tests to make sure that the installation actually works. At a minimum the [tests workflow](https://github.com/osm-search/Nominatim/blob/20a333dd9b3649699b2e5269cda95c4f30b00aae/.github/workflows/ci-tests.yml#L36) should work just like on the Linux installations. Ideally, we get a [install workflow](https://github.com/osm-search/Nominatim/blob/20a333dd9b3649699b2e5269cda95c4f30b00aae/.github/workflows/ci-tests.yml#L143) to work as well.

Thanks for the feedback. 
To spread this further in an organized way, I would like to approach this in stages (with one PR per stage) to keep things maintainable and reviewable. 
### Stage 1 : (Windows Test Parity) 
- Setup PostgreSQL on Windows
- Build the required Python environment
- Run the same Python unit tests and BDD tests that currently run on Linux (or as close as possible, skipping only where Windows-specific blockers exist)

### Stage 2 : 
Add full import support + osm2pgsql so that BDD import scenarios run successfully.

### Stage 3 : 
Replicate the full install workflow on Windows to match Linux install validation.

I’ll start with Stage 1 so we can first get reliable test coverage on Windows and iterate from there.
Please let me know if this plan aligns with expectations.

lonvia:
Restricting CI tests to the Python unit tests in a first stage PR sounds reasonable. Lets start with that.

kad-link:
> Restricting CI tests to the Python unit tests in a first stage PR sounds reasonable. Lets start with that.

Hi @lonvia , while going through the current [ci-tests.yml](https://github.com/osm-search/Nominatim/blob/master/.github/workflows/ci-tests.yml) , I noticed the `create-archive` job which prepares the source tarball and later consumed by subsequent jobs (including Python unit-tests). 
Since I am working on adding CI coverage for Windows, I am wondering if it should be implemented in the CI for Windows too. Since `create-archive` is only packaging the source tree , I think it would be reasonable to keep this job running on Linux, while the other jobs run on Windows. Linux runners seem well suited for shell-based packaging steps.

So, could you kindly clarify these 2 issues in summary ?
-  Should the Windows CI workflow include the `create-archive` job?
- If so, is it preferred to keep  the `create-archive` job Linux-only (while the rest of the code Windows-specific), or should there be a Windows-specific implementation for the `create-archive` job too?

lonvia:
The `create-archive` job just downloads the additional country data, making sure that this is only done once. There is no need to duplicate this job on Windows. Simply make your job depend on this one and use the created artifact in the same way the Linux jobs do.

kad-link:
> The `create-archive` job just downloads the additional country data, making sure that this is only done once. There is no need to duplicate this job on Windows. Simply make your job depend on this one and use the created artifact in the same way the Linux jobs do.

Just a small query here, by using the same artifact, do you mean to create the Windows CI coverage in the existing file itself  ([ci-tests.yml](https://github.com/osm-search/Nominatim/blob/master/.github/workflows/ci-tests.yml) ) ?or make another file for Windows CI ? 
If a separate file, then artifact can't be shared and it has to be written again right?

kad-link:
hi @lonvia , as discussed I added a Windows CI job to run a subset of Python unit tests. I used the same workflow as that of the existing tests for Linux job. I included only those tests that are Database independent and don't require PostgreSQL or PostGIS setup, since we decided to keep this as a “Python-only” unit test stage on Windows. 
Would you prefer that I open a fresh PR for this, or continue updating this existing one?

mtmail:
```
monkeypatch.setenv('NOMINATIM_FOOBAR', '/dont/care')
...
value = getattr(self, name)
cfgpath = Path(value)
if not cfgpath.is_absolute():
```
This behaves differently on Linux and Windows. On Linux `/dont/care` is an absolute path. On Windows it's not, Windows would require a drive letter like `C:\dont\care` for an absolute path. Try something like `(tmp_path / "does_not_exist").resolve()`.

kad-link:
Hi @mtmail , I have resolved the above problem but I'm seeing mypy failures on `src/nominatim_api/server/starlette/server.py`(`request.state.num_results` and `request.state.query_stats` attr-defined). Now what's confusing is the exact same mypy step (`mypy --strict --python-version 3.9 src`) passed in the previous CI run and I didn't modify this file. Could you confirm if this is expected or if we want to resolve this newly risen issue in the same PR?

mtmail:
@lonvia knows mypy better, there always seem to be tiny changes between versions. In the meantime I would skip the test or make that section optional and concentrate on the unit/bdd tests.

kad-link:
hi @mtmail , I have tried fixing the mypy failure around `request.state.*`. I used Python's built-in `setattr()`. Starlette's `request.state` is a dynamic container and now the attributes on `request.state` are dynamically added at runtime. Hence I did a small workaround so CI is not blocked, without changing runtime behavior.

mtmail:
Looks good.

I guess the next step (phase?) would be to get Postgres step and run more tests.

The install step for Linux (https://github.com/osm-search/Nominatim/blob/master/.github/actions/setup-postgresql/action.yml) has more logic because we want to be able run older versions of Postgres than the Github runner base image (https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md)

In the output of the Windows tests there's a link to documentation which software is installed by default
image

https://github.com/actions/runner-images/blob/win25/20260111.179/images/windows/Windows2025-Readme.md

```
ServiceName 	postgresql-x64-17
Version 	17.7
ServiceStatus 	Stopped
ServiceStartType 	Disabled
```

That should be good enough for the Windows tests. Make sure the service is started. And run the same `createuser` commands from the Linux `action.yml` file.

```
        - name: Setup database
          run: |
              sudo systemctl restart postgresql
              sudo -u postgres createuser -S www-data
              sudo -u postgres createuser -s runner
          shell: bash
```

It's not clear where the steps should be added. Either in a new `.github/actions/setup-postgresql/action-windows.yml` or in the existing `tests-windows` section of `ci-tests.yml`. I'd do the latter because it's only used once.

Good luck, thanks for your patience

kad-link:
Just a small query here @mtmail , shall we continue the work in the same PR or would you prefer new PRs for every stage ?

mtmail:
@mtmail Let's keep this PR open, then all Windows-test related discussion is in one place.

Can you create a new PR for the changes to src/nominatim_api/server/starlette/server.py ? In every release mypy keeps adding more warnings and this one is the same for all operating systems. You found a good solution and I'd be good if a PR with your name attached gets merged instead of me repeating the changes. Something like "fixes a new mypy type warning related to Starlette state object"

kad-link:
> @mtmail Let's keep this PR open, then all Windows-test related discussion is in one place.
> 
> Can you create a new PR for the changes to src/nominatim_api/server/starlette/server.py ? In every release mypy keeps adding more warnings and this one is the same for all operating systems. You found a good solution and I'd be good if a PR with your name attached gets merged instead of me repeating the changes. Something like "fixes a new mypy type warning related to Starlette state object"

sure, thank you

lonvia:
The main issue with Postgres on Windows used to be that it doesn't come with Postgis. If that's still the case, it might help to have a look at the [action script from osm2pgsql](https://github.com/osm2pgsql-dev/osm2pgsql/blob/master/.github/actions/win-postgres/action.yml) which downloads and installs postgis. Not sure it is still the best way of doing things, but it is one that works.

otbutz:
PostGIS is available via Stack Builder. However, a quick search didn't reveal how to invoke it via cli.

https://www.enterprisedb.com/docs/supported-open-source/postgresql/installing/using_stackbuilder/

kad-link:
> Looks good.
> 
> I guess the next step (phase?) would be to get Postgres step and run more tests.
> 
> The install step for Linux (https://github.com/osm-search/Nominatim/blob/master/.github/actions/setup-postgresql/action.yml) has more logic because we want to be able run older versions of Postgres than the Github runner base image (https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md)
> 
> In the output of the Windows tests there's a link to documentation which software is installed by default image
> 
> https://github.com/actions/runner-images/blob/win25/20260111.179/images/windows/Windows2025-Readme.md
> 
> ```
> ServiceName 	postgresql-x64-17
> Version 	17.7
> ServiceStatus 	Stopped
> ServiceStartType 	Disabled
> ```
> 
> That should be good enough for the Windows tests. Make sure the service is started. And run the same `createuser` commands from the Linux `action.yml` file.
> 
> ```
>         - name: Setup database
>           run: |
>               sudo systemctl restart postgresql
>               sudo -u postgres createuser -S www-data
>               sudo -u postgres createuser -s runner
>           shell: bash
> ```
> 
> It's not clear where the steps should be added. Either in a new `.github/actions/setup-postgresql/action-windows.yml` or in the existing `tests-windows` section of `ci-tests.yml`. I'd do the latter because it's only used once.
> 
> Good luck, thanks for your patience



hello @mtmail , there are three places i found for adding PostgreSQL/PostGIS Windows installation commands. They are :
- modifying the existing `.github/actions/setup-postgresql/action.yml` to automatically detect the OS running and run the correct script : 
  The files go like this : 
          - `action.yml`  ( if Linux run .sh  , if Windows run .ps1) 
          - `scripts/install-linux.sh`    
          - `scripts/install-windows.ps1`   

- create a completely new action file for Windows like `.github/actions/setup-postgresql/action-windows.yml`
- write the Windows installation commands directly inside the main `ci-tests.yml` as recommended by you


after considering the maintainability, readability and reliability of code, I found method 1 to be better. could you kindly confirm which way to proceed with ?

lonvia:
I'd say go for a completely separate action. There will be very little overlap with Windows. I think that has to go in `.github/actions/setup-postgresql-windows/action.yml` though?

kad-link:
Hello @lonvia , I have added `.github/actions/setup-postgresql-windows/action.yml` and called it in `ci-tests.yml`. Below , I am listing some of the major decisions I considered for the setup of PostgreSQL/PostGIS : 

- Created a description heading in `setup-postgresql` 
   `description: 'Installs PostgreSQL and PostGIS and configures it for CI tests` since the Action Metadata was firing an error.     
   I am attaching an image for your reference : 
image

- For the `.github/actions/setup-postgresql-windows/action.yml` action, I followed the [osm2pgsql/.github/actions/win-postgres/action.yml](https://github.com/osm2pgsql-dev/osm2pgsql/blob/master/.github/actions/win-postgres/action.yml) (since it is a known method) while aligning with the existing `setup-postgresql/action.yml`. 

- For the configuration of PostgreSQL ie. 
 ```
 - name: Adapt postgresql configuration
          run: |
              echo 'fsync = off' | sudo tee /etc/postgresql/${PGVER}/main/conf.d/local.conf
              echo 'synchronous_commit = off' | sudo tee -a /etc/postgresql/${PGVER}/main/conf.d/local.conf
              echo 'full_page_writes = off' | sudo tee -a /etc/postgresql/${PGVER}/main/conf.d/local.conf
              echo 'shared_buffers = 1GB' | sudo tee -a /etc/postgresql/${PGVER}/main/conf.d/local.conf
              echo 'port = 5432' | sudo tee -a /etc/postgresql/${PGVER}/main/conf.d/local.conf
          shell: bash
          env:
              PGVER: ${{ inputs.postgresql-version }}
```
I used SQL commands (`ALTER SYSTEM`) to ask PostgreSQL to update its own configuration since editing txt files on Windows is generally considered messy due to variable file paths.

These are the major things I took into consideration. I kindly request you to have a review. 

Also the `make lint` and `make tests` have passed : 
image
image

**Comment by kad-link** [comment]
Also, I am currently working on the `psycopg.errors.DeadlockDetected` problem , appearing in the BDD tests. 

The tests are actually failing with database deadlocks causing many errors.  But these same tests passed successfully (9/9) on my fork. 


image
image


I am currently checking their CPU core availability if they are triggering parallel execution differently. The solution will most probably be to force sequential execution  for these tests. I will bring forward a fix shortly.

**Comment by kad-link** [comment]
hi @lonvia , the Linux CI failed with a `psycopg.errors.DeadlockDetected` error in the BDD API tests. This sounds like a flaky race condition caused by `pytest-xdist` running the API tests in parallel, where the database lock was accessed by more than one worker. Since this is unrelated to my changes (and my Windows CI passed), could you please re-run the failed Linux job?

Also , I found a fix for this issue ,  using `pytest-xdist` grouping to force the specific tests that are failing to run sequentially. If this deadlock becomes a frequent nightmare, I would be happy to open a separate PR to fix it.

**Comment by kad-link** [comment]
Also, lemme tell you about the fix , 
I identified that this can be permanently resolved using `pytest-xdist` grouping. The following is the implementation : 
- Add a marker in `conftest.py` to group the API tests 
- Update the CI workflow to run with `--dist=loadgroup`


This forces the conflicting tests to run on the same worker sequentially while keeping the rest of the tests running parallel. 
A glimpse of the marker goes like this : 
`item.add_marker(pytest.mark.xdist_group(name = "api_exclusive"))`

**Comment by lonvia** [comment]
None of the tests should be run in parallel. We use hard-coded database names and the tests would get into each others way. Where have you see pytest-xdist used in the CI?

Note: there is an occasional fail on deadlock in the tests during installation of the functions. That's an actual issue with the code, not a problem of the tests. I just haven't been able to pin it down yet because it is difficult to reproduce.

**Comment by kad-link** [comment]
> None of the tests should be run in parallel. We use hard-coded database names and the tests would get into each others way. Where have you see pytest-xdist used in the CI?
> 
> Note: there is an occasional fail on deadlock in the tests during installation of the functions. That's an actual issue with the code, not a problem of the tests. I just haven't been able to pin it down yet because it is difficult to reproduce.


Ahh I see. My apologies for incorrectly assuming that `pytest-xdist` was running tests in parallel after I saw 2 different PIDs in the deadlock logs. Thanks for the clarification. 

Moving on to the CI(Windows) , could you please review the PostgreSQL/PostGIS setup and discuss the next steps? Hopefully, I think we can ignore the deadlock failure for now since Windows CI passed

**Comment by lonvia** [comment]
Looks all good so far. Just keep going to make the Python tests pass.

pyICU is a bit of a difficult one because they don't provide wheels with the shared library included. They expect libicu to be present externally. I see that somewhat has made full packages for Windows [here](https://github.com/cgohlke/pyicu-build/releases). Maybe try those. That might be easier than installing libicu on the Windows system.

File: .github/workflows/ci-tests.yml [hunk 1]
```diff
@@ -140,6 +140,37 @@ jobs:
...
+            - uses: actions/download-artifact@v4
+              with:
+                  name: full-source
+
+            - name: Unpack Nominatim
+              run: tar xf nominatim-src.tar.bz2
+
+            - uses: ./Nominatim/.github/actions/setup-postgresql-windows
+              with:
+                  postgresql-version: 16
```
lonvia: Lets aim for the latest version. That would be 18.

.github/workflows/ci-tests.yml
```diff
@@ -140,6 +140,37 @@ jobs:
...
+
+            - name: Set up Python
+              uses: actions/setup-python@v5
+              with:
+                  python-version: '3.11'
```
lonvia: And here the latest version would be 3.14.

.github/workflows/ci-tests.yml
kad-link: okay

.github/workflows/ci-tests.yml
kad-link: okay

kad-link:
It looks like we have bumped onto #3598. I verified my previous research and am positive about the solution: we need to set the event loop policy explicitly to `SelectorEventLoop` before the tests begin. This overrides the default Windows `ProactorEventLoop`(causing the crash) and ensures compatibility with `psycopg` and SQLAlchemy’s async database access.

kad-link:
Today's CI failures resolved successfully : 


- `UnicodeDecodeError` (Encoding):
   Issue: Windows runners default to `CP1252` encoding, causing `refresh.py` to crash when reading UTF8 files.
    Fix: Implemented `PYTHONUTF8=1` in the workflow environment to ensure uniform file handling.

- `OperationalError` (Spatialite issues):
    Issue: SQLite tests were failing because `mod_spatialite` and its dependencies (e.g., `libgeos.dll`) could not be found.
    Fix: Added a step to download `mod_spatialite-5.1.0-win-amd64` and correctly added its directory to `$env:GITHUB_PATH` so that Python can load it.

- `WinError 2` (Missing Executables):
    Issue: `osm2pgsql` related tests were crashing because the corresponding binaries are not yet present in the Windows CI. 
    Fix: Added `--ignore` flags to the `pytest` command to skip its tests

lonvia:
That looks great, thanks for having it seen through so far.

Do you want to see if you can get the final issue with osm2pgsql sorted out? We [provide ready to use binaries](https://osm2pgsql.org/doc/install/windows.html), so it should be as simple as downloading and unpacking https://osm2pgsql.org/download/windows/osm2pgsql-latest-x64.zip. If it's not that simple, we could also merge the state you have gotten to so far, but I think it's worth a try.

kad-link:
> That looks great, thanks for having it seen through so far.
> 
> Do you want to see if you can get the final issue with osm2pgsql sorted out? We [provide ready to use binaries](https://osm2pgsql.org/doc/install/windows.html), so it should be as simple as downloading and unpacking https://osm2pgsql.org/download/windows/osm2pgsql-latest-x64.zip. If it's not that simple, we could also merge the state you have gotten to so far, but I think it's worth a try.

sure. thank you for the resources. I would be happy to give it a try.

otbutz:
>     * `UnicodeDecodeError` (Encoding):
>       Issue: Windows runners default to `CP1252` encoding, causing `refresh.py` to crash when reading UTF8 files.
>       Fix: Implemented `PYTHONUTF8=1` in the workflow environment to ensure uniform file handling.

Isn't that just a band aid solution? If we expect things to be UTF8, we should specify that encoding when we're reading files.

kad-link:
> > ```
> > * `UnicodeDecodeError` (Encoding):
> >   Issue: Windows runners default to `CP1252` encoding, causing `refresh.py` to crash when reading UTF8 files.
> >   Fix: Implemented `PYTHONUTF8=1` in the workflow environment to ensure uniform file handling.
> > ```
> 
> Isn't that just a band aid solution? If we expect things to be UTF8, we should specify that encoding when we're reading files.



Yes, agreed that it's a temporary patch to avoid CI errors. I’ll manually add `encoding='utf-8'` in the relevant file reads.

kad-link:
> > ```
> > * `UnicodeDecodeError` (Encoding):
> >   Issue: Windows runners default to `CP1252` encoding, causing `refresh.py` to crash when reading UTF8 files.
> >   Fix: Implemented `PYTHONUTF8=1` in the workflow environment to ensure uniform file handling.
> > ```
> 
> Isn't that just a band aid solution? If we expect things to be UTF8, we should specify that encoding when we're reading files.

I looked into fixing the code considering it as proper long term solution, but there are dozens of file I/O calls scattered across the codebase that would require manual updating. I think modifying many files in this PR will make it messy and chaotic. So can we keep `PYTHONUTF8=1` for now to get the Windows CI going? I can maybe open a separate follow up PR dedicated to fixing the file encodings.

lonvia:
> I think modifying many files in this PR will make it messy and chaotic. So can we keep PYTHONUTF8=1 for now to get the Windows CI going? I can maybe open a separate follow up PR dedicated to fixing the file encodings.

Sounds reasonable. Maybe open an issue as a reminder that this needs looking into?

kad-link:
hi @lonvia , what I did since I last messaged : 

- **Added `osm2pgsql` installation step and removed the `--ignore` flags in Python unit tests** : 
But this gave rise to  2 major errors - `FileNotFoundError: [WinError 2] The system cannot find the file specified` and `OSError: [WinError 193] %1 is not a valid Win32 application`. 

- To resolve the `FileNotFoundError: [WinError 2]`, I tried resolving the `osm2pgsql` `PATH` by actually searching for the `osm2pgsql.exe` inside the archive, verifying its existence and adding it to `PATH`. Still both the errors were up there even after this step. 
Next, I avoided relying on the preconfigured `PostgreSQL` env vars on the Windows runner image and set them manually.

- I then added a step to build `Nominatim`, installing the `nominatim-api` and `nominatim-db` packages before running the tests, aligning with the Linux CI. 

- Here, a new error popped up : `cp: target 'C:\Program Files\PostgreSQL\18' is not a directory`
The reason was that the Windows runner did not provide a `PostgreSQL 18 `PostGIS bundle at the download location. So the `PostgreSQL 18` directory was never created , hence, this again led to the `FileNotFoundError: [WinError 2]` error. To fix this, the `PostgreSQL` version had to be downgraded to 17 as it was stable. 

- The PostgreSQL ran successfully and the CI tests moved forward. A new error appeared : `FileNotFoundError: Forced include not found: D:\a\Nominatim\Nominatim\Nominatim\packaging\nominatim-db\lib-lua`
To fix this, a new step was added on to the CI with the label `Fix Windows paths` that replaced Unix-style linked directories with actual copies by copying required folders into `packaging/nominatim-db` to avoid broken links and path related issues on Windows. 

- The errors now reduced to half. `FileNotFoundError: [WinError 2] The system cannot find the file specified` was completely removed and now only  `OSError: [WinError 193] %1 is not a valid Win32 application` remained in the CI. 
To fix this ,I initially dealt with the `osm2pgsql` for  locating `osm2pgsql.exe`. Instead of assuming a single directory, the workflow now searched recursively for `osm2pgsql.exe`, selected the first valid match, and added its directory to `PATH`. Also required debugging steps were added. 

- Finally there was one error that was not waving off : `args = '...\\Temp\\...\\osm2pgsql_mock --version'`
The test setup was creating a fake file named `osm2pgsql_mock` to trick the system into thinking `osm2pgsql` is installed during the test. On Linux, this was somehow running but on Windows, the OS sees `osm2pgsql_mock` has no `.exe` extension misinterprets as not a binary and throws `[WinError 193] %1 is not a valid Win32 application`.




For now, I have skipped the affected tests ( around 15 in number). I feel  this remaining issue is isolated and  can be handled separately in another PR. 
If everything looks good to you, would it be okay to merge this PR ?

kad-link:
> > I think modifying many files in this PR will make it messy and chaotic. So can we keep PYTHONUTF8=1 for now to get the Windows CI going? I can maybe open a separate follow up PR dedicated to fixing the file encodings.
> 
> Sounds reasonable. Maybe open an issue as a reminder that this needs looking into?

I will also open up this issue shortly

lonvia:
> Next, I avoided relying on the preconfigured PostgreSQL env vars on the Windows runner image and set them manually.

Any idea where exactly this was caused in the code? Sounds like something that can be improved in the code by respecting the right environment variables.

> I then added a step to build Nominatim, installing the nominatim-api and nominatim-db packages before running the tests, aligning with the Linux CI.

Good question if the Linux  CI build should be doing that. The test suite should be able to run without it.

> The PostgreSQL ran successfully and the CI tests moved forward. A new error appeared : FileNotFoundError: Forced include not found: D:\a\Nominatim\Nominatim\Nominatim\packaging\nominatim-db\lib-lua

This indicates a more fundamental issue with packaging. That's something to handle separately. If you can get away with running the tests without installing nominatim-{db,api}, lets do so.

> For now, I have skipped the affected tests ( around 15 in number).

Yes, that's okay.

lonvia:
One more thing. There is currently a deprecation warning: 

```
test\python\conftest.py:13
  D:\a\Nominatim\Nominatim\Nominatim\test\python\conftest.py:13: DeprecationWarning: 'asyncio.WindowsSelectorEventLoopPolicy' is deprecated and slated for removal in Python 3.16
    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```

Can you please find the appropriate replacement?

kad-link:
> Any idea where exactly this was caused in the code? Sounds like something that can be improved in the code by respecting the right environment variables.
> 
> >
There is no issue with the existing code. 
Actually what happened was initially @mtmail and I had a conversation in which PostgreSQL was supposed to use version 17.7 ie. the latest version the runners were providing. Below is the reference. 


> Looks good.
> 
> I guess the next step (phase?) would be to get Postgres step and run more tests.
> 
> The install step for Linux (https://github.com/osm-search/Nominatim/blob/master/.github/actions/setup-postgresql/action.yml) has more logic because we want to be able run older versions of Postgres than the Github runner base image (https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md)
> 
> In the output of the Windows tests there's a link to documentation which software is installed by default image
> 
> https://github.com/actions/runner-images/blob/win25/20260111.179/images/windows/Windows2025-Readme.md
> 
> ```
> ServiceName 	postgresql-x64-17
> Version 	17.7
> ServiceStatus 	Stopped
> ServiceStartType 	Disabled
> ```
> 
> That should be good enough for the Windows tests. Make sure the service is started. And run the same `createuser` commands from the Linux `action.yml` file.
> 
> ```
>         - name: Setup database
>           run: |
>               sudo systemctl restart postgresql
>               sudo -u postgres createuser -S www-data
>               sudo -u postgres createuser -s runner
>           shell: bash
> ```
> 
> It's not clear where the steps should be added. Either in a new `.github/actions/setup-postgresql/action-windows.yml` or in the existing `tests-windows` section of `ci-tests.yml`. I'd do the latter because it's only used once.
> 
> Good luck, thanks for your patience


Later, you told me to get the latest PostgreSQL version 18. The `PGBIN` and `PGROOT` global variables provided by the Github runners were hardcoded to version 17.7. So, to override the default preconfigured variables, I had to set them manually in the CI. But while testing CI for 18, I encountered CI errors saying PostGIS binaries could not be fetched and that was likely due to a URL or filename mismatch. Since there were other errors too at that time, I felt downgrading the version to 17 would be a good idea (I can try v18 again). But the setup of the env vars was not removed. 

But now instead of removing, I think of keeping the setup as it is since it takes the input of the version and fetches the resources dynamically rather than using the Github runners which fix to the latest version (which keeps updating).

kad-link:
> Good question if the Linux CI build should be doing that. The test suite should be able to run without it.
> 
> >



> This indicates a more fundamental issue with packaging. That's something to handle separately. If you can get away with running the tests without installing nominatim-{db,api}, lets do so.
> 
> >

Ok. Will remove `nominatim-api`, `nominatim-db` installation steps and run the tests

lonvia:
Makes sense now about the Postgresql versions. Thanks for the explanation and sorry for the confusion.

kad-link:
hi @lonvia , I have removed the packaging as discussed. The tests have passed successfully. My next checkpoints are to upgrade the PostgreSQL version to 18 and find a substitute for this deprecation warning : 

> ```
> test\python\conftest.py:13
>   D:\a\Nominatim\Nominatim\Nominatim\test\python\conftest.py:13: DeprecationWarning: 'asyncio.WindowsSelectorEventLoopPolicy' is deprecated and slated for removal in Python 3.16
>     asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
> ```

lonvia:
Doesn't seem to work with 18, lets stay with 17. It's fine for now.

kad-link:
@lonvia , just came up with a suggestion. I think it would be better to raise a separate PR for the `asyncio.set_event_loop_policy` fix. Since there is an existing issue for it ie. #3598 , a dedicated PR can neatly close it. What do you think ?

lonvia:
I see where the event loop issue is becoming complex. Already, let's leave it open. I've added a comment to that end.

Can you please rebase this on current master and reorganise the commits to get rid of all the fix-up commits?

kad-link:
Done @lonvia

lonvia:
Alright, lets see how this goes with daily use. Thanks a lot for patiently keeping on this.

Would you mind opening issues for the remaining todos? Preferably a separate one for each.

PR #3949, opened 2026-01-27
Feat: Add admin function for granting access to read-only user

### LINKED ISSUE #3931: Add admin function for granting access to read-only user
Nominatim has support to grant read-only access to its database tables to a special user which is then used to implement the search functions. Currently the GRANT commands are all over the place in the table creation SQL, for example [here](https://github.com/osm-search/Nominatim/blob/20a333dd9b3649699b2e5269cda95c4f30b00aae/lib-sql/tables.sql#L26). This makes it impossible to grant read access to a different user after the database has been imported. The goal her is to separate out the grant commands and provide an admin function to grant these rights later.

Steps to do:

* find all GRANT commands and move them into a new SQL file in `lib-sql`
* add a new function to `src/nominatim_db/tools/admin.py` that runs the SQL script (use [SQLPreprocessor](https://github.com/osm-search/Nominatim/blob/master/src/nominatim_db/db/sql_preprocessor.py) for that)
* add a subcommand `nominatim refresh --ro-access` to `src/nominatim_db/clicmd/refresh.py` which calls the function
* make sure import calls the new function

Note: the name of the user which gets read-only access is currently set with the [NOMINATIM_DATABASE_WEBSUSESR](https://nominatim.org/release-docs/latest/customize/Settings/#nominatim_database_webuser) config variable. That shouldn't change. So there is no need to hand in the user name explicitly anywhere.

_Please read the [guidelines on using AI](https://github.com/osm-search/Nominatim/blob/master/CONTRIBUTING.md#using-ai-assisted-code-generators) before starting to work on this issue._

Itz-Agasta:
## Summary
This PR implements the feature  to allow granting read-only database access to a different user after import, without requiring a full re-import. Closes #3931

### Changes Made
1. Created `lib-sql/grants.sql` - Extracted all GRANT statements into a single, centralized SQL file
2. Added `grant_ro_access()` function in `src/nominatim_db/tools/admin.py` - Runs the grants SQL
3. Added `--ro-access` CLI argument in `src/nominatim_db/clicmd/refresh.py` - New command: `nominatim refresh --ro-access` also flag in `src/nominatim_db/clicmd/args.py cuz mypy was shouting
4. Updated import process in `src/nominatim_db/tools/database_import.py` - Automatically calls grants.sql during table creation
5. Removed GRANT statements from lib-sql/tables.sql`

6. `test/python/tools/test_database_import.py` was causing some import  issues so add a mock for grant.sql file 

#### Results

https://github.com/user-attachments/assets/5de87315-26ce-4b4b-af4d-d67dba9c7d39

#### Test Results
Screenshot_27-Jan_17-34-00_9825

## AI usage
NONE

### DIFF
--- a/lib-sql/grants.sql
+++ b/lib-sql/grants.sql
@@ -15,7 +15,6 @@ GRANT SELECT ON nominatim_properties TO "{{config.DATABASE_WEBUSER}}";
 -- Location tables
 GRANT SELECT ON location_property_tiger TO "{{config.DATABASE_WEBUSER}}";
 GRANT SELECT ON location_property_osmline TO "{{config.DATABASE_WEBUSER}}";
-GRANT SELECT ON location_area TO "{{config.DATABASE_WEBUSER}}";
 GRANT SELECT ON location_postcodes TO "{{config.DATABASE_WEBUSER}}";
 
 -- Search tables
@@ -28,10 +27,6 @@ GRANT SELECT ON placex TO "{{config.DATABASE_WEBUSER}}";
 GRANT SELECT ON place_addressline TO "{{config.DATABASE_WEBUSER}}";
 GRANT SELECT ON placex_entrance TO "{{config.DATABASE_WEBUSER}}";
 
--- OSM data tables
-GRANT SELECT ON planet_osm_ways TO "{{config.DATABASE_WEBUSER}}";
-GRANT SELECT ON planet_osm_rels TO "{{config.DATABASE_WEBUSER}}";
-
 -- Error/delete tracking tables
 GRANT SELECT ON import_polygon_error TO "{{config.DATABASE_WEBUSER}}";
 GRANT SELECT ON import_polygon_delete TO "{{config.DATABASE_WEBUSER}}";
@@ -43,3 +38,10 @@ GRANT SELECT ON country_osm_grid TO "{{config.DATABASE_WEBUSER}}";
 {% if 'word' in db.tables %}
 GRANT SELECT ON word TO "{{config.DATABASE_WEBUSER}}";
 {% endif %}
+
+-- Special phrase tables
+{% for table in db.tables %}
+{% if table.startswith('place_classtype_') %}
+GRANT SELECT ON {{ table }} TO "{{config.DATABASE_WEBUSER}}";
+{% endif %}
+{% endfor %}
--- a/src/nominatim_db/tools/freeze.py
+++ b/src/nominatim_db/tools/freeze.py
@@ -18,7 +18,6 @@ UPDATE_TABLES = [
     'address_levels',
     'gb_postcode',
     'import_osmosis_log',
-    'import_polygon_%',
     'location_area%',
     'location_road%',
     'place',


Itz-Agasta:
One thing i saw  small inconsistency in the GRANT statements. In some places it was

```sql
GRANT SELECT ON import_status TO "{{config.DATABASE_WEBUSER}}";
```

while in other  the explicit form

```sql
GRANT SELECT ON TABLE country_name TO "{{config.DATABASE_WEBUSER}}";
```
Also `on`,  `ON` inconsistency..In pg the `TABLE` keyword is optional for tables so i usually omit it in my own SQL... I hope its fine by y guys...

I had to add a mock for grants.sql in the tests without it, the tests were failing with an import error originating from the database_import file (expected)...... will change the approach if y guys  suggest some different way

src/nominatim_db/tokenizer/icu_tokenizer.py
```diff
@@ -144,10 +144,6 @@ def update_statistics(self, config: Configuration, threads: int = 2) -> None:
             with conn.cursor() as cur:
                 cur.execute('SET max_parallel_workers_per_gather TO 0')
 
-            sqlp = SQLPreprocessor(conn, config)
-            sqlp.run_string(conn,
-                            'GRANT SELECT ON tmp_word TO "{{config.DATABASE_WEBUSER}}"')
-            conn.commit()
```
lonvia, line 150: This needs to stay. This is a refresh function that replaces the word table and the new table needs to have the same access rights.

src/nominatim_db/tools/migration.py
```diff
@@ -250,8 +249,6 @@ def create_place_postcode_table(conn: Connection, config: Configuration, **_: An
                     geometry Geometry(Geometry, 4326) NOT NULL
                 )
                 """)
-            sqlp.run_string(conn,
-                            'GRANT SELECT ON location_postcodes TO "{{config.DATABASE_WEBUSER}}"')
```
lonvia, line 254: Anything in this whole migration file shouldn't be touched either. The grants here are a one-time thing done when upgrading the Nominatim software versions.

test/python/tools/test_database_import.py
```diff
@@ -201,6 +201,8 @@ def test_create_tables(self, temp_db_conn, temp_db_cursor, reverse):
                        """CREATE FUNCTION test() RETURNS bool
                           AS $$ SELECT {{db.reverse_only}} $$ LANGUAGE SQL""")
 
+        self.write_sql('grants.sql', "-- Mock grants file for testing\n")
```
lonvia, line 204: You could do the same trick as for tables.sql: create a function in the mock file and the call it below, to test that your mock file was actually executed.

To be fair though, the usefulness of this test is rather limited. So I don't might if it just stays this way.

lib-sql/grants.sql
```diff
@@ -0,0 +1,50 @@
...
+GRANT SELECT ON country_osm_grid TO "{{config.DATABASE_WEBUSER}}";
+
+-- Tokenizer tables (word table)
+{% if 'word' in db.tables %}
+GRANT SELECT ON word TO "{{config.DATABASE_WEBUSER}}";
+{% endif %}
+
+-- Tiger import table (if exists)
+{% if 'location_property_tiger_import' in db.tables %}
+GRANT SELECT ON location_property_tiger_import TO "{{config.DATABASE_WEBUSER}}";
```
lonvia: This is the wrong table, it should be `location_property_tiger`.

The  table `location_property_tiger_import` is only a temporary table created during import. This way, an update of the table doesn't interfere with existing data: import data into new table, give that table the appropriate rights. Then in one transaction delete the old table and rename the newly import table, effectively moving the new data atomically into place.

lonvia (commented):
You missed at least `location_property_osmline` as far as I can see. The tests unfortunately don't cover the situation well. Here is how you can test your changes using the BDD API tests:

* run the BDD tests at least once with `pyest test/bdd/test_api.py --nominatim-purge` (you need to repeat that step every time you make changes to the source code that have an influence on the database layout)
* create a password for the www-data user in Postgresql: `psql -d nominatim -c "ALTER USER \"www-data\" PASSWORD '1234567890'"`
* now run the BDD tests against the www-data user: `PGPASSWORD=1234567890 PGUSER=www-data PGHOST=127.0.0.1 pytest test/bdd/test_api.py` (do not use purge here, you cannot create a new test database with this user)
* create another postgres user: `createuser granttest`
* create a password for that user (same as above)
* now run the granting script against the new user: `NOMINATIM_DATABASE_WEBUSER=granttest ./nomiantim-cli.py refresh --ro-access`
* finally confirm that the tests also pass with this new user: `PGPASSWORD=1234567890 PGUSER=granttest PGHOST=127.0.0.1 pytest test/bdd/test_api.py`

_Warning:_ PGPASSWORD is considered insecure. Please use this only on a test server and only when the PostgreSQL server is not listening on a port that is world-wide accessible. You also might want to make up your own random password instead of copying the obvious one above.

Itz-Agasta:
> You missed at least `location_property_osmline` as far as I can see. 

Ahhh  It's actually already in my grants.sql at line 17... I did include it when moving the grants. Perhaps you missed it in the review? 
About  other issues pushing fixes now

lib-sql/grants.sql
Itz-Agasta: I already have a `location_property_tiger` in line 16  in my garnts.sql ................ It was actually moved  form `lib-sql/tiger_import_finish.sql`... But i get it now the grant in `tiger_import_finish.sql` also  needs to stay there during the import process (like icu_tokenizer.py), because it happens before the rename ( Lamo i i added If statments to handle it tough)...

I shouldnt try to handle the tiger import table in grants.sql at all....

test/python/tools/test_database_import.py
Itz-Agasta: Mmm i also prefer keeping  it like this :(

Itz-Agasta:
> The tests unfortunately don’t cover this situation well. Here’s how you can test your changes using the BDD API tests.

hey @lonvia  While testing this like y said I ran into an issue with this approach.

The `pytest test/bdd/test_api.py` tests run against the `test_api_nominatim` database, which is created via `--nominatim-purge`.... However, `./nominatim-cli.py refresh --ro-access` grants access to the new web user on the **`nominatim`** database by default.

Because of this, running:

```bash
PGPASSWORD=1234567890 PGUSER=granttest PGHOST=127.0.0.1 pytest test/bdd/test_api.py
```

doesn’t work as expected.

I tried to work around this by setting:

```bash
NOMINATIM_DATABASE_DSN=pgsql:host=localhost;port=5432;user=agasta;dbname=test_api_nominatim
```

and re-running the steps. However, I then noticed that the BDD test database is **minimal** and does not contain all the tables that exist in a full Nominatim import.

image

but my grants.sql( assumes all tables exists)..... Now idk should we make the grants more robust ?........ My feeling is that this may be unnecessary, since in a real Nominatim db these tables will always be present..

Itz-Agasta:
On the good side after running with `--ro-access` things seems to  work correctly on a real db. For eg I can now do

```bash
psql -U granttest -d nominatim -c "SELECT COUNT(*) FROM placex;"
 count 
-------
  3991
(1 row)
```

which previously failed with an authorization error. Likewise, this now works as expected:

```bash
NOMINATIM_DATABASE_DSN="pgsql:host=localhost;port=5432;user=granttest;password=9876543210;dbname=nominatim" \
./nominatim-cli.py search --query "Monaco"
```

Earlier this was failing with:

```
sqlalchemy.exc.ProgrammingError: (psycopg.errors.InsufficientPrivilege)
permission denied for table nominatim_properties
```

So the fix itself seems correct  (i have also added a vedio in my pr showing these tesing in a docker pg)... Lemme know what you think.

lonvia:
Sorry about the `location_property_osmline`. I overlooked it indeed.

The API test database isn't crippled, it is just [frozen](https://nominatim.org/release-docs/latest/admin/Import/#dropping-data-required-for-dynamic-updates). The grant refresh definitely should be working on a frozen DB as well. It just so happens that your code nicely uncovers all the errors that the current code is making when granting.

From a quick check:
* location_area, planet_osm_ways and planet_osm_rels shouldn't be accessible via ro access at all.
* import_polygon_error and import_polygon_delete shouldn't be deleted by the freeze.  [This line](https://github.com/osm-search/Nominatim/blob/9a979b7429c17fc0f6751d11e31edd8d0345254b/src/nominatim_db/tools/freeze.py#L21) should be deleted.
* The `place_classtype_*` are missing for the grant. This is going to be slightly tricky because which tables exactly are created is different from installation to installation. You should have the list of existing tables in the jinja `db.tables` variable and can loop through that to find all tables with the prefix.

Itz-Agasta:
Hi @lonvia Thanks for the feedback... you can check now 

https://github.com/user-attachments/assets/13126532-19f2-40fc-8bc2-a3670f212307


> The place_classtype_* are missing for the grant. 

I have updated it and also done some tests on this. Its working too

image

Itz-Agasta:
One thing i was wondering 
> location_area, planet_osm_ways and planet_osm_rels shouldn't be accessible via ro access at all.

whats  the reasoning for removing the location_area grant ? I see that it was in the original tables.sql previously....aren't api is gonna query it ?

lonvia:
The location_area tables are only used when preparing the data during import/updates. It may be that a long. long time ago the  query part accessed these tables, but it certainly doesn't nowadays. It's just one of these left-over code parts nobody has looked at for years.

This is one of the other big advantages of this PR. With all grants in one place, it will be much easier to see and recheck that the permissions that are given out are sensible.

Looks like this is now passing nicely the test sequence I described above. Thanks for bearing with me while fixing all those old code smells.

PR #3952, opened 2026-01-28
Tests: resolve an issue in the place-linkage name expansion test case

jayaddison:
## Summary
This is a followup to pull request #3687 that added test coverage to illustrate how name expansion should work for linked places in a city/boundary example.

I think there is a bug in the test case -- the city and boundary don't yet seem to be linked.  To begin with, this draft pull request reactivates the test case with an extra assertion to confirm that (it is expected to fail).

If that seems reasonable then I think we may be able to add a suitable `admin` tag (administrative boundary level) to the boundary to resolve this.

## AI usage
None

### DIFF
--- a/test/bdd/features/db/import/linking.feature
+++ b/test/bdd/features/db/import/linking.feature
@@ -306,11 +306,17 @@ Feature: Linking of places
         Given the places
             | osm | class    | type           | name+name                | geometry    |
             | N9  | place    | city           | Popayán                  | 9           |
-            | R1  | boundary | administrative | Perímetro Urbano Popayán | (1,2,3,4,1) |
+        Given the places
+            | osm | class    | type           | name+name                | geometry    | admin |
+            | R1  | boundary | administrative | Perímetro Urbano Popayán | (1,2,3,4,1) | 8     |
         And the relations
             | id | members  |
             | 1  | N9:label |
         When importing
+        Then placex contains
+            | object      | linked_place_id |
+            | N9:place    | R1              |
+            | R1:boundary | -               |
         Then placex contains
             | object | name+_place_name | name+_place_name:es |
             | R1     | Popayán          | Popayán             |


lonvia:
You are right about the missing `admin` level. boundaries without it [get dropped from the address hierarchy](https://github.com/osm-search/Nominatim/blob/9a979b7429c17fc0f6751d11e31edd8d0345254b/settings/address-levels.json#L48) and subsequently not linked. So, please feel free to amend the test.

jayaddison:
Thanks @lonvia, will do.

test/bdd/features/db/import/linking.feature
```diff
@@ -297,7 +298,6 @@ Feature: Linking of places
          | R1         | LabelPlace |
 
 
-    @skip
```
jayaddison, line 300: NB: I'll also restore this `@skip` annotation in a few moments.

test/bdd/features/db/import/linking.feature
```diff
@@ -306,11 +306,17 @@ Feature: Linking of places
         Given the places
             | osm | class    | type           | name+name                | geometry    |
             | N9  | place    | city           | Popayán                  | 9           |
-            | R1  | boundary | administrative | Perímetro Urbano Popayán | (1,2,3,4,1) |
+        Given the places
```
jayaddison, line 309: Perhaps this should be phrased as `And the places`, to make the test case read more naturally (and similar with the added `Then` clause).

jayaddison: On the other hand, it makes the code diff simpler not to modify those other setup/assertion clause introduction lines.  I'll leave it as-is, for code review.

lonvia: Both is fine. Use whatever you like better. Even though I would prefer 'And' in the tests I wrote, what actually ended up there had a lot to do with how much the steps were copy and pasted or changed after the first version was written.

lonvia:
Do you mind squashing this to a single commit (and then force pushing onto the same branch)? Then we can merge it.

jayaddison:
Done, thanks - although I forgot to rewrite the commit message the first time, oops.

PR #3954, opened 2026-01-29
Refactor of linked-place name-expansion code

jayaddison:
## Summary
I think the place-name expansion logic for linked places in `placex_indexing_prepare` is a good candidate for some refactoring, and I'd like to do that before attempting a fix for #2714.

## AI usage
None

### DIFF
--- a/lib-sql/functions/placex_triggers.sql
+++ b/lib-sql/functions/placex_triggers.sql
@@ -87,7 +87,6 @@ BEGIN
     -- Add all names from the place nodes that deviate from the name
     -- in the relation with the prefix '_place_'. Deviation means that
     -- either the value is different or a given key is missing completely
-
     SELECT hstore(array_agg('_place_' || key), array_agg(value))
       INTO extra_names
       FROM each(location.name)


lib-sql/functions/placex_triggers.sql
```diff
@@ -83,27 +83,21 @@ BEGIN
...
-      -- either the value is different or a given key is missing completely
-      IF result.name is null THEN
-        SELECT hstore(array_agg('_place_' || key), array_agg(value))
-          INTO result.name
-          FROM each(location.name);
-      ELSE
-        SELECT hstore(array_agg('_place_' || key), array_agg(value)) INTO extra_names
-          FROM each(location.name - result.name);
-        {% if debug %}RAISE WARNING 'Extra names: %', extra_names;{% endif %}
+    {% if debug %}RAISE WARNING 'Names original: %, location: %', result.name, location.name;{% endif %}
```
jayaddison, line 86: NB: In fact, I think that the outer `IF location.place_id is not NULL ...` conditional block can be removed here too.  We should relocate it to wrap the debug output messaging, though, so that that doesn't become noisier.

lib-sql/functions/placex_triggers.sql
```diff
@@ -80,31 +80,31 @@ BEGIN
   -- place already here.
   SELECT * INTO location FROM find_linked_place(p);
 
+  result.linked_place_id := location.place_id;
+
+  {%- if debug -%}
```
jayaddison: These whitespace-trim directives (the little `-` markers in the Jinja template) might not be perfect; I haven't yet explored how the evaluated SQL function text is formatted (either in debug or non-debug mode).

lib-sql/functions/placex_triggers.sql
```diff
@@ -80,31 +80,31 @@ BEGIN
...
-        END IF;
-      END IF;
+  -- Add all names from the place nodes that deviate from the name
+  -- in the relation with the prefix '_place_'. Deviation means that
+  -- either the value is different or a given key is missing completely
+  SELECT hstore(array_agg('_place_' || key), array_agg(value))
+    INTO extra_names
+    FROM each(location.name)
+   WHERE result.name IS NULL
+      OR result.name -> key IS NULL
```
lonvia: This should be `NOT result.name ? key`.

Also, the convention with `->` and `->>` is to treat it like a `.` rather than an operator, i.e. no spaces around it.

lib-sql/functions/placex_triggers.sql
jayaddison: Ok, I have now checked, and in `debug` mode, the current formatting is not great.  I have a fix prepared to commit in a few moments, and I'll push that after the currently-running GitHub Actions continuous integration tests complete.

jayaddison:
> I'm in two minds about this. While your refactoring is certainly more elegant, it might have a noticeable impact on performance. This function is called 200 million times on a full planet import. In 99.9% of the cases the linked place will be empty. Until now these cases would shortcut while with your version two assignments, one select and one concatenation is done. At a minimum the the `location.place_id is not NULL` test needs to stay. The inner part is probably less critical as in most cases both the place and the linked location have a name and we end up with select and concatenation anyway.

In hindsight: the `Extra names` debug statement is also omitted, and I imagine that's useful under some circumstances.

I find relational-style SQL logic easier to reason about and modify than procedural `IF` branching -- but I would be OK with adjusting this or discarding the refactor.

jayaddison:
Either way: I'll try to find some time to performance-test this using a subset of the planet file, to learn more about how the PostgreSQL query planner handles this function.

lib-sql/functions/placex_triggers.sql
lonvia: I wouldn't worry too much what comes out after jinja is done as long as the syntax is right. It is not meant for humans to read.

lonvia:
Having some actual performance numbers on that function would be great. Just be aware that it behaves very differently for the cases where rank address is 0, 1-25, 26-27 and 28-30. So best to test those separately.

jayaddison:
NB && note-to-self: The commandline that I've prepared to reindex subsets of ranked places, while iterating on / experimenting with changes to `placex_indexing_prepare`, is:

```sh
$ psql nominatim -c 'update placex set indexed_status = 2 where linked_place_id is not null'; nominatim refresh --functions; nominatim index --minrank 0 --maxrank 0
```

jayaddison:
I did learn a few things here, but I think it's an unnecessary -- and unnecessarily complicated -- refactor.  Similar to #3956, I'm going to close this and to refocus on a simpler fix.

PR #3955, opened 2026-01-30
docs: fix multiple typos in documentation and source code

AmmarYasser455:
Fixes miscellaneous typos found via codespell in the documentation, source comments, and variable names (where safe).

### DIFF
--- a/lib-sql/tiger_import_start.sql
+++ b/lib-sql/tiger_import_start.sql
@@ -17,7 +17,7 @@ CREATE TABLE location_property_tiger_import (
 
 
 -- Lookup functions for tiger import when update 
--- informations are dropped (see gh-issue #2463)
+-- tables are dropped (see gh-issue #2463)
 CREATE OR REPLACE FUNCTION getNearestNamedRoadPlaceIdSlow(in_centroid GEOMETRY,
                                                       in_token_info JSONB)
   RETURNS BIGINT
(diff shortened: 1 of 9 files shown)


src/nominatim_api/search/query.py
```diff
@@ -274,8 +274,11 @@ class QueryStruct:
         need to be direct neighbours. Thus the query is represented as a
         directed acyclic graph.
 
+        When created, a query contains a single node: the start of the
```
mtmail: where does this sentence come from?

matkoniecz: yes, other changes are typo fixes that make sense but this change is not a typo fix despite PR description

AmmarYasser455:
Thanks for catching that! I've removed the added sentence that wasn't a typo fix. It was mistakenly included (and duplicated) in the `QueryStruct` docstring. The updated commit now contains only the legitimate typo corrections.

lib-sql/tiger_import_start.sql
```diff
@@ -17,7 +17,7 @@ CREATE TABLE location_property_tiger_import (
 
 
 -- Lookup functions for tiger import when update 
--- informations are dropped (see gh-issue #2463)
+-- information are dropped (see gh-issue #2463)
```
lonvia: This needs a grammar fix. Although the better fix would probably be to go from 'information' to 'tables'.

AmmarYasser455: Fixed! Changed 'information' to 'tables' as suggested. Thanks for the feedback!

lonvia:
Thank you.

PR #3956, opened 2026-02-01
Indexer: enable zero-ranked addresses to be indexed independently

jayaddison:
## Summary
During attempts to [benchmark reindexing of subsets of the `placex` table](https://github.com/osm-search/Nominatim/pull/3954#issuecomment-3817892192), I've encountered two problems:

1. Items with rank zero did not appear to be detected as pending.
1. The indexer CLI process kept on running in a loop, without reindexing any items.

The first issue seems to be when a requested `minrank` of zero is coerced up to the minimum value `1` instead, and the second seems to be due to `has_pending` checking for _any_ pending places-to-index, regardless of the requested rank range.

## AI usage
None

### DIFF
--- a/src/nominatim_db/indexer/indexer.py
+++ b/src/nominatim_db/indexer/indexer.py
@@ -146,7 +146,7 @@ class Indexer:
                 total_tuples = {row.rank_address: row.count for row in cur}
 
         with self.tokenizer.name_analyzer() as analyzer:
-            for rank in range(max(0, minrank), maxrank + 1):
+            for rank in range(max(1, minrank), maxrank + 1):
                 if rank >= 30:
                     batch = 20
                 elif rank >= 26:


lonvia:
Address rank 0 is an odd one. It contains all the features that don't make up or are part of an address and are not a point of interest either: rivers, mountain peaks, lakes and the like. They used to be mixed into the other features but that turned out cause errors because they would need features for their address computation that weren't ready yet. So at some point I moved computation of rank 0 into an extra step _after_ computing rank 30. And being very, very lazy, I just mingled that together with rank 30 computation:

https://github.com/osm-search/Nominatim/blob/1cf5464d3ac593a3cd54517af2262805a66cd659/src/nominatim_db/indexer/indexer.py#L150-L152

(As you can see, address interpolations are mingled with that rank as well. It's a bit less evil because you can't select on the command line to index interpolations only.)

That is, of course, very confusing for somebody who doesn't know the code's history. So a fix is welcome. But the part you need to fix is that little piece of code above.

lonvia:
Addendum: why I was being lazy here. The import code at https://github.com/osm-search/Nominatim/blob/1cf5464d3ac593a3cd54517af2262805a66cd659/src/nominatim_db/indexer/indexer.py#L68

calls the indexing in multiple steps. If 0 is to index rank 0, then another step is needed around there.

jayaddison:
Ok, gradually that's making sense; rank=zero places have some processing dependency on other features/places?  One more question, either way:

- The comment for `index_full` mentions that boundaries are indexed before anything else - I don't think that's true, though - during each iteration of the `while` loop, other elements will be indexed too (and only up to 100 boundaries).

jayaddison:
Ok, and rank=0 are _not_ part of the address, understood (I misread that on the first-pass).

jayaddison:
> Ok, gradually that's making sense; rank=zero places have some processing dependency on other features/places? One more question, either way:
> 
>     * The comment for `index_full` mentions that boundaries are indexed before anything else - I don't think that's true, though - during each iteration of the `while` loop, other elements will be indexed too (and only up to 100 boundaries).

My mistake here; if more than `100` boundaries are indexed, then the database statistics will be updated (via `ANALYZE`) - but there's no limit on the batch size when indexing those.

...plus, features in address-rank levels `1...4` are continents/oceans/countries -- so even though those might be indexed first, they probably don't tend to be enclosed by many (any?) outer boundaries (?), so from a practical point of view, perhaps it's OK to index them first.

src/nominatim_db/indexer/indexer.py
```diff
@@ -31,14 +31,19 @@ def __init__(self, dsn: str, tokenizer: AbstractTokenizer, num_threads: int):
...
         """
         with connect(self.dsn) as conn:
             with conn.cursor() as cur:
-                cur.execute("SELECT 'a' FROM placex WHERE indexed_status > 0 LIMIT 1")
+                cur.execute(""" SELECT 'a'
+                                 FROM placex
+                                WHERE rank_address BETWEEN %s AND %s
+                                  AND indexed_status > 0
+                                LIMIT 1""",
+                            (minrank, maxrank))
```
jayaddison, line 46: More caution/wariness than I initially considered should be applied to this query change, if accepted.  I hadn't considered the performance implications -- and although I do notice that the `idx_placex_rank_address_sector` index is probably the one that the database engine chooses to service this query, _and_ that it has `rank_address` as the first column (making it sargeable), it's worth being wary of dragons here.

jayaddison:
For git-history cleanliness, and adhering to the `CONTRIBUTING.md` guidance, I think it would be best for me to file a separate issue for the bug/quirk that the CLI indexer may not exit when items outside of the selected reindexing rank-range exist -- and then to extract the relevant code from here into a separate pull request.

jayaddison:
I think I've overcomplicated my approach my focusing on the refactor/preparation before determining how/where a fix could be added, especially given the introduction-level scoping of the task.  I'm going to close these pull requests and refocus.  There might be one or two useful details to extract for separate pull requests later.

PR #3957, opened 2026-02-02
Indexing: add default-language placename from linked places

### LINKED ISSUE #2714: Linked place names need to include the default language

jayaddison:
## Summary
Enables, for example, linked city boundaries to appear in place search results when querying for the local default-language name of the city.

Resolves #2714.

## AI usage
None

Edit: fixup: typo/grammar correction
Edit: resolve the two concerns mentioned in this description
Edit: cleanup pull request description.
Edit: redraft description to explain the purpose instead of the implementation.

### DIFF
--- a/lib-sql/functions/placex_triggers.sql
+++ b/lib-sql/functions/placex_triggers.sql
@@ -87,11 +87,9 @@ BEGIN
     IF location.name is not NULL THEN
       {% if debug %}RAISE WARNING 'Names original: %, location: %', result.name, location.name;{% endif %}
 
+      -- Add the linked-place (e.g. city) name as a searchable placename in the default language (if any)
       default_language := get_country_language_code(location.country_code);
-      IF default_language is NULL OR location.name ? ('name:' || default_language) THEN
-        -- No default language, or a name in the local language is already configured; do nothing
-      ELSE
-        -- Merge in a default-language name from the linked place
+      IF default_language is not NULL AND NOT location.name ? ('name:' || default_language) THEN
         location.name := location.name || hstore('name:' || default_language, location.name->'name');
       END IF;
 


test/bdd/features/db/import/linking.feature
```diff
@@ -297,15 +297,14 @@ Feature: Linking of places
...
     Scenario: Linked places expand default language names
         Given the grid
             | 1 |   | 2 |
             |   | 9 |   |
             | 4 |   | 3 |
         Given the places
-            | osm | class    | type           | name+name                | geometry    |
-            | N9  | place    | city           | Popayán                  | 9           |
+            | osm | class    | type           | name+name                | geometry    | country |
+            | N9  | place    | city           | Popayán                  | 9           | co      |
```
lonvia: You need to put the grid into Columbia to make sure the country assignment comes out as excepted. See for example https://github.com/osm-search/Nominatim/blob/cc9c8963f31d6175ca098969bedfffaee06780c3/test/bdd/features/db/query/postcodes.feature#L17

Once you have done that, you can just use `country_code` (of either the linker or the linkee, they are forcibly the same).

lib-sql/functions/utils.sql
```diff
@@ -124,7 +124,7 @@ CREATE OR REPLACE FUNCTION add_default_place_name(country_code VARCHAR(2),
 DECLARE
   default_language VARCHAR(10);
 BEGIN
-  IF name is not null AND array_upper(akeys(name),1) > 1 THEN
+  IF name is not null AND array_upper(akeys(name),1) > 0 THEN
```
lonvia: The assumption here was that if there is only exactly one name in the hstore, then it will always be chosen no matter what language is selected. Adding a special field for the default language just adds noise.

This assumption doesn't really work anymore in the case of a linked place. This could only be skipped if the combined name set of linker and linkee have exactly one name. This complicates things somewhat. I wouldn't want to drop the assumption for the default case where there is no linkee and the object has exactly one name tag as that is the situation for probably 90% of the places in the database.

That might mean that you cannot reuse this function after all and have to do something clever in the loop that merges the two name sets instead.

jayaddison: Ok, thank you.  That makes sense, I think -- and I had a sense that that might be the reason.

To attempt to rephrase that, to check my understanding: most places simply have `name`, and in those cases, there's no value in adding `name:<default-lang>` in addition -- so the precondition that `add_default_place_name` checks for is that there are at least two names (either `name` and `name:xx`, or `name:xx` and `name:yy`).

I have to be honest: I'm still trying to understand what the logic will be to do this inside `placex_indexing_prepare` -- but I think I'll be able to figure it out by experimenting with a few approaches.  Basically we want the locally-understood name of the place to be annotated in the `_place_name:<local-lang>` field -- but only if it differs from the existing place name (and fortunately, the existing name-merge code handles most of that duplication-check already).

test/bdd/features/db/import/linking.feature
jayaddison: Brilliant - much simpler; thanks!

lib-sql/functions/placex_triggers.sql
```diff
@@ -85,6 +86,13 @@ BEGIN
 
     IF location.name is not NULL THEN
       {% if debug %}RAISE WARNING 'Names original: %, location: %', result.name, location.name;{% endif %}
+
+      -- Merge in a default-language name from the linked place, if none is set
+      default_language_key := 'name:' || get_country_language_code(location.country_code);
+      IF NOT location.name ? default_language_key THEN
+        location.name := location.name || hstore(default_language_key, location.name->'name');
+      END IF;
```
jayaddison: I _think_ we can omit a conditional branch to check for the case when `name:<default-language>` _does_ exist -- because the location itself is within the country that has the default language, and so I think/hope it's a reasonable assumption that the `name` field, without any language-code specifier, is the native/local placename.

But I am finding this surprisingly difficult to reason about!

jayaddison: (afterthought: are there places where `name` is _not_ specified, and only `name:<lang-one>`, `name:<lang-two>`, and so on exist?  e.g. absence of a default-lang `name` in preference for explicit naming for each locally-used language?)

lib-sql/functions/placex_triggers.sql
```diff
@@ -93,7 +101,8 @@ BEGIN
           INTO result.name
           FROM each(location.name);
       ELSE
-        SELECT hstore(array_agg('_place_' || key), array_agg(value)) INTO extra_names
+        SELECT hstore(array_agg('_place_' || key), array_agg(value))
+          INTO extra_names
```
jayaddison: This change is extraneous, and I would be happy to omit it.  It was how I got started understanding this part of the code; I lean towards reverting it.

lib-sql/functions/placex_triggers.sql
```diff
@@ -85,6 +86,13 @@ BEGIN
 
     IF location.name is not NULL THEN
       {% if debug %}RAISE WARNING 'Names original: %, location: %', result.name, location.name;{% endif %}
+
+      -- Merge in a default-language name from the linked place, if none is set
+      default_language_key := 'name:' || get_country_language_code(location.country_code);
+      IF NOT location.name ? default_language_key THEN
```
jayaddison: Perhaps we could remove this `IF` condition?

jayaddison: I think it is safer and less confusing if we ensure that any existing `name:<lang-code>` value is retained.

lib-sql/functions/placex_triggers.sql
jayaddison: Same answer as in the [other thread](https://github.com/osm-search/Nominatim/pull/3957/files#r2769693095) (why did I open two threads about it?) - no, I think it's simpler and safer to leave any existing `name:<default-lang>` entry unmodified, so let's keep the conditional branch.

lib-sql/functions/placex_triggers.sql
jayaddison: Yep; I'll rebase and omit this modification.

lib-sql/functions/placex_triggers.sql
```diff
@@ -85,6 +86,13 @@ BEGIN
 
     IF location.name is not NULL THEN
       {% if debug %}RAISE WARNING 'Names original: %, location: %', result.name, location.name;{% endif %}
+
+      -- Merge in a default-language name from the linked place, if none is set
+      default_language_key := 'name:' || get_country_language_code(location.country_code);
```
lonvia: Sorry but there is one further complication here: there are countries without a default languages (think Canada or Switzerland). `get_country_language_code()` returns null for them. You shouldn't do anything in that case.

jayaddison: Ah yep, thank you - I'll adjust the logic to manage that.  That might result in me changing the variable name to `default_language`.. I'll explore a few options before pushing a commit.

jayaddison: I'm in two minds about whether the repeated string-concatenation (`'name:' || ...`) are worth consolidating into the variable assignment step -- it does evaluate to null if the right-hand value is `null`.  I'll leave it unchanged for now, though.

Edit: markdown syntax fixup.

jayaddison:
Something I'm thinking about is: whether it is possible to create a kind of diff to inspect the effects of this change.  I'd like to do that to get a sense of what practical effects it has, before merge.

lib-sql/functions/placex_triggers.sql
```diff
@@ -85,6 +86,15 @@ BEGIN
 
     IF location.name is not NULL THEN
       {% if debug %}RAISE WARNING 'Names original: %, location: %', result.name, location.name;{% endif %}
+
+      default_language := get_country_language_code(location.country_code);
+      IF default_language is NULL OR location.name ? ('name:' || default_language) THEN
+        -- No default language, or a name in the local language is already configured; do nothing
```
lonvia: Invert your IF condition and get rid of the empty THEN path.

jayaddison:
Thanks @lonvia!

lib-sql/functions/placex_triggers.sql
```diff
@@ -85,6 +86,13 @@ BEGIN
 
     IF location.name is not NULL THEN
       {% if debug %}RAISE WARNING 'Names original: %, location: %', result.name, location.name;{% endif %}
+
+      -- Add the linked-place (e.g. city) name as a searchable placename in the default language (if any)
+      default_language := get_country_language_code(location.country_code);
+      IF default_language is not NULL AND NOT location.name ? ('name:' || default_language) THEN
```
lonvia, line 92: So, this actually missed one condition: `location.name` may exist but `location.name->'name'` may be empty. As a result a null name is being added which later breaks the Python code.

@jayaddison Could you add a BDD test for that case and then fix it?

jayaddison: @lonvia can do, yep.

NB: yesterday I posted what I think was an offensive/dismissive comment while closing and locking a discussion thread as (what I regarded as) spam in an unrelated FOSS project.  I mention it so that you're aware in case you'd prefer not to merge my contributions.  I can provide a link to the context if helpful.

Maybe it seems unusual of me to mention it, but I think I'd prefer to explain that now than for you to potentially learn of a problem independently later (in fact, the reason I logged back on a few minutes ago was because I'd decided to mention it).

My default in the interim will be to proceed because this is known-incomplete and since I started the work I can complete it.  I won't be offended if you'd prefer that I not contribute that (or if you'd like to rollback any of the contributions I'd made so far).

I'm finding the current AI situation kinda heated and confusing - I don't use AI, intentionally turn off any AI features, and I remain fairly robust in believing that that's an acceptable stand to take (while trying not to be too critical of those who do), but it seems to have gotten to me a bit over this weekend.

jayaddison: I'm having difficulty arranging a test case where one of the name hstore fields is NULL; it's probably fairly straightforward, but I'll have to revisit it again later.

lonvia: You don't need the NULL in the test. Simply give the linked place a name with a different local than the default one and you should be able to recreate the behaviour, e.g. `"name:xx": "Foo"`

jayaddison: That makes sense; thank you

PR #3960, opened 2026-02-02
Tests: parameter-agnostic 'Indexer.has_pending' monkeypatching

jayaddison:
## Summary
This is a pedantic/nitpicky refactor of the monkeypatching of the Indexer `has_pending` method.  Assigning the instance-method of a `list`'s `pop` method is clever, but is fragile because it requires that the runtime parameters are compatible with `list.pop`'s method signature.

By adjusting the substitute method to call an intermediate `lambda` function that accepts (and ignores) arbitrary parameters, we can allow the underyling `has_pending` method signature to change.

Discovered during an exploratory branch in #3956 that modified the signature of the `Indexer.has_pending` method.

## AI usage
None

Edit: add context to the description to explain where this idea/suggestion originated from.
Edit: update the implementation strategy.

### DIFF
--- a/test/python/cli/test_cli.py
+++ b/test/python/cli/test_cli.py
@@ -91,8 +91,9 @@ class TestCliWithDb:
         postcode_mock = async_mock_func_factory(nominatim_db.indexer.indexer.Indexer,
                                                 'index_postcodes')
 
+        has_pending_retvals = [True, False]
         monkeypatch.setattr(nominatim_db.indexer.indexer.Indexer, 'has_pending',
-                            [False, True].pop)
+                            lambda *args, **kwargs: has_pending_retvals.pop(0))
 
         assert self.call_nominatim('index', *params) == 0
 


test/python/cli/test_cli.py
```diff
@@ -91,8 +91,12 @@ def test_index_command(self, monkeypatch, async_mock_func_factory, table_factory
         postcode_mock = async_mock_func_factory(nominatim_db.indexer.indexer.Indexer,
                                                 'index_postcodes')
 
+        def _has_pending_once(*args, **kwargs):
+            monkeypatch.setattr(nominatim_db.indexer.indexer.Indexer, 'has_pending',
+                                lambda *args, **kwargs: False)
+            return True
         monkeypatch.setattr(nominatim_db.indexer.indexer.Indexer, 'has_pending',
-                            [False, True].pop)
+                            _has_pending_once)
```
lonvia: That piece of code is not easy to understand. A simpler version:

```
expected_results = [True, False]
monkeypatch.setattr(nominatim_db.indexer.indexer.Indexer, 'has_pending',	
                    lambda *args, **kwargs: expected_results.pop())
```

Note, that this also ensures that the function is not called again after having returned False because pop() then throws at the empty list. That was intentional.

jayaddison: That is even more clever than I realised - and your suggestion is better here.  Thanks!

jayaddison: NB: `pop` takes items from the end of the list; so the ordering of the list has to be reversed (this confused me, to begin with)

jayaddison:
I've updated the pull request description with the newly-selected implementation approach, and I'll squash and rebase the commits here.  Again, it is a pedantic change, so merging is not strictly necessary - I'd like it in order to support a subsequent fix/change, that I'll file a bugreport for soon.

jayaddison:
Thanks!

PR #3971, opened 2026-02-05
Indexer: allow 'has_pending' to consider address-rank subsets

### LINKED ISSUE #3969: Indexer: infinite loop in CLI indexer when selected ranks do not cover all pending items
**Describe the bug**
The CLI indexer calls the `Indexer.has_pending` method as a loop continuation condition; the idea is that once there are no pending items to index, the process can stop.

However: `has_pending` checks for _any_ pending items to index, regardless of their address rank.  If the indexer CLI is run for an address rank range that excludes those items, the loop continues indefinitely.

**To Reproduce**
1. Import a fresh dataset into `Nominatim` - it can be a small dataset.
2. Configure some places for reindexing
  * For example, I encountered this bug after using `psql nominatim -c 'UPDATE placex SET indexed_status = 2 WHERE linked_place_id IS NOT NULL'`, to flag places that have linked places for reindexing.
3. Invoke the command-line indexer to run for a min/max address rank range that does _not_ include all of the items from step (2).
  * Items in rank 1 should not have linked places, so we could run: `nominatim index --minrank 1 --maxrank 1`

**Expected behaviour**
The indexer should exit after the selected address ranks have been reindexed.

**Actual behaviour**
The indexer continues to run after the selected address ranks have been reindexed.

Sample output:

```
$ nominatim index --minrank 1 --maxrank 1
...
2026-02-05 15:48:57: Starting indexing boundaries using 2 threads
2026-02-05 15:48:57: Starting indexing rank (1 to 1) using 2 threads
2026-02-05 15:48:57: Starting rank 1 (using batch size 1)
2026-02-05 15:48:57: Done 0/0 in 0 @ 0.000 per second - FINISHED rank 1

2026-02-05 15:48:57: Starting indexing postcodes using 2 threads
2026-02-05 15:48:57: Starting postcodes (location_postcode) (using batch size 20)
2026-02-05 15:48:57: Done 0/0 in 0 @ 0.000 per second - FINISHED postcodes (location_postcode)

2026-02-05 15:48:57: Starting indexing boundaries using 2 threads
2026-02-05 15:48:57: Starting indexing rank (1 to 1) using 2 threads
2026-02-05 15:48:57: Starting rank 1 (using batch size 1)
2026-02-05 15:48:57: Done 0/0 in 0 @ 0.000 per second - FINISHED rank 1
...
```

**Software Environment (please complete the following information):**
- Nominatim version: [v5.2.0](https://github.com/osm-search/Nominatim/commit/9a13b62fb9f6905e707d2c3ec9d6959e7249bb76)
- Postgresql version: 18.1
- Postgis version: 3.6.1
- OS: Debian Linux (forky/testing)

**Hardware Configuration (please complete the following information):**
N/A

**Postgresql Configuration:**
N/A

**Nominatim Configuration:**
N/A

**Additional context**
Noticed during an attempt to run some local performance benchmarking for (abandoned) refactor pull request #3954.

A fix may be blocked-by / require #3960.

jayaddison:
## Summary
The `Indexer.has_pending` method hasn't been aware of min/max rank options supplied during CLI indexing -- and this means that in some cases the indexer will run continuously because it still thinks that there are places requiring reindexing, despite those being outside the min/max request range.

This pull request enables the method to check within subsets of the places by rank address number.

Note: A supporting PostgreSQL database index is available for this, named [`idx_placex_rank_address_sector`](https://github.com/osm-search/Nominatim/blob/cd2f6e458b7f4ce634bd16747788f1603ed265bc/lib-sql/tables.sql#L187-L189) -- it is defined over the columns `rank_address, geometry_sector` and is a [partial index](https://www.postgresql.org/docs/18/indexes-partial.html) for places that have a non-zero `indexed_status`.

Resolves #3969.

## AI usage
None

Edit: simplify pull request description.

### DIFF
--- a/src/nominatim_db/indexer/indexer.py
+++ b/src/nominatim_db/indexer/indexer.py
@@ -31,14 +31,19 @@ class Indexer:
         self.tokenizer = tokenizer
         self.num_threads = num_threads
 
-    def has_pending(self) -> bool:
+    def has_pending(self, minrank: int = 0, maxrank: int = 30) -> bool:
         """ Check if any data still needs indexing.
             This function must only be used after the import has finished.
             Otherwise it will be very expensive.
         """
         with connect(self.dsn) as conn:
             with conn.cursor() as cur:
-                cur.execute("SELECT 'a' FROM placex WHERE indexed_status > 0 LIMIT 1")
+                cur.execute(""" SELECT 'a'
+                                  FROM placex
+                                 WHERE rank_address BETWEEN %s AND %s
+                                   AND indexed_status > 0
+                                 LIMIT 1""",
+                            (minrank, maxrank))
                 return cur.rowcount > 0
 
     async def index_full(self, analyse: bool = True) -> None:
--- a/test/python/cli/test_cmd_index.py
+++ b/test/python/cli/test_cmd_index.py
@@ -1,0 +1,42 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# This file is part of Nominatim. (https://nominatim.org)
+#
+# Copyright (C) 2025 by the Nominatim developer community.
+# For a full list of authors see the git log.
+"""
+Tests for index command of the command-line interface wrapper.
+"""
+import pytest
+
+import nominatim_db.indexer.indexer
+
+
+class TestCliIndexWithDb:
+
+    @pytest.fixture(autouse=True)
+    def setup_cli_call(self, cli_call, cli_tokenizer_mock):
+        self.call_nominatim = cli_call
+        self.tokenizer_mock = cli_tokenizer_mock
+
+    def test_index_empty_subset(self, monkeypatch, async_mock_func_factory, placex_row):
+        placex_row(rank_address=1, indexed_status=1)
+        placex_row(rank_address=20, indexed_status=1)
+
+        mocks = [
+            async_mock_func_factory(nominatim_db.indexer.indexer.Indexer, 'index_boundaries'),
+            async_mock_func_factory(nominatim_db.indexer.indexer.Indexer, 'index_by_rank'),
+            async_mock_func_factory(nominatim_db.indexer.indexer.Indexer, 'index_postcodes'),
+        ]
+
+        def _reject_repeat_call(*args, **kwargs):
+            assert False, "Did not expect multiple Indexer.has_pending invocations"
+
+        has_pending_calls = [nominatim_db.indexer.indexer.Indexer.has_pending, _reject_repeat_call]
+        monkeypatch.setattr(nominatim_db.indexer.indexer.Indexer, 'has_pending',
+                            lambda *args, **kwargs: has_pending_calls.pop(0)(*args, **kwargs))
+
+        assert self.call_nominatim('index', '--minrank', '5', '--maxrank', '10') == 0
+
+        for mock in mocks:
+            assert mock.called == 1, "Mock '{}' not called".format(mock.func_name)
--- a/src/nominatim_db/clicmd/index.py
+++ b/src/nominatim_db/clicmd/index.py
@@ -64,4 +64,4 @@ class UpdateIndex:
             if not args.boundaries_only:
                 await indexer.index_by_rank(args.minrank, args.maxrank)
                 await indexer.index_postcodes()
-            has_pending = indexer.has_pending()
+            has_pending = indexer.has_pending(args.minrank, args.maxrank)


src/nominatim_db/indexer/indexer.py
```diff
@@ -31,15 +31,24 @@ def __init__(self, dsn: str, tokenizer: AbstractTokenizer, num_threads: int):
...
             Otherwise it will be very expensive.
         """
         with connect(self.dsn) as conn:
             with conn.cursor() as cur:
-                cur.execute("SELECT 'a' FROM placex WHERE indexed_status > 0 LIMIT 1")
-                return cur.rowcount > 0
+                cur.execute(""" SELECT 'a'
+                                  FROM placex
+                                 WHERE rank_address BETWEEN %s AND %s
+                                   AND indexed_status > 0 LIMIT 1""",
```
jayaddison:
```suggestion
                                   AND indexed_status > 0
                                 LIMIT 1""",
```

jayaddison:
Ok; thank you - I'll put this PR back into a draft status and will later reopen #3956; that'll probably be this evening or tomorrow, because I'll want to re-check some of the details.

jayaddison:
> Ok; thank you - I'll put this PR back into a draft status and will later reopen #3956; that'll probably be this evening or tomorrow, because I'll want to re-check some of the details.

Oh, good: GitHub prevents restoring a branch from an earlier pull request and force-pushing that to a different commit and re-opening it.  That's almost certainly sensible and for reasons -- it also means that in this case, I'll open a fresh pull request instead of #3956 to isolate/extract the rank-zero special casing.

jayaddison:
Awaiting / depends-on #3979

jayaddison:
> I'm not convinced it is a good idea to enshrine the rank 0 hack even further in the code. Your previous approach was better: get rid of the rank 0 hack first. It should be fairly easy to fix: make sure rank 0 runs after all other ranks have run and make sure that anybody calling the function doesn't assume that rank 30 includes rank 0.

@lonvia something that has been on the back of my mind: is there any risk that removing the special-casing behaviour from the code means that operational complexity increases?  e.g. any data-consumers/users/crontabs etc who rely on `maxrank=30` implicitly reindexing rank-zero now have to adjust to index both explicitly?

src/nominatim_db/indexer/indexer.py
jayaddison: Argh.  I thought I might forget this during rebasing.  And yet..

jayaddison: (change applied locally, and branch rebased again)

lonvia:
Using the `--minrank` and `--maxrank` parameters for indexing is a highly advanced feature where you need to know what you are doing. It's not meant for daily operations. So I'm not worried.

I've reworked the tests for indexing a bit. Can you please rebase this on latest master? And then it would be good if you can add a test that reproduces the issue you have seen with indexing running forever, which should then pass with this code.

jayaddison:
Is it wasteful that we call the various indexing methods at-least-once?  I am wondering if we could modify the code to bypass that when there are no places to index.

https://github.com/osm-search/Nominatim/blob/2ddb19c0b0f1775779700d5bfb6dce3dea747c64/src/nominatim_db/clicmd/index.py#L60

However.. if so, there would need to be an additional/separate check for pending postcode indexing.

test/python/cli/test_cmd_index.py
```diff
@@ -0,0 +1,48 @@
...
+        mocks = [
+            async_mock_func_factory(nominatim_db.indexer.indexer.Indexer, 'index_boundaries'),
+            async_mock_func_factory(nominatim_db.indexer.indexer.Indexer, 'index_by_rank'),
+            async_mock_func_factory(nominatim_db.indexer.indexer.Indexer, 'index_postcodes'),
+        ]
+
+        def _has_pending_proxy_once(*args, **kwargs):
+            global has_pending_called
+            assert not has_pending_called, "Did not expect multiple Indexer.has_pending invocations"
+            has_pending_called = True
```
jayaddison: - [x] TODO: factor-out the global variable by repatching the method after the first invocation.

test/python/cli/test_cmd_index.py
```diff
@@ -0,0 +1,42 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# This file is part of Nominatim. (https://nominatim.org)
+#
+# Copyright (C) 2025 by the Nominatim developer community.
+# For a full list of authors see the git log.
+"""
+Tests for import command of the command-line interface wrapper.
```
jayaddison:
```suggestion
Tests for index command of the command-line interface wrapper.
```

jayaddison:
> Is it wasteful that we call the various indexing methods at-least-once? I am wondering if we could modify the code to bypass that when there are no places to index.
> 
> https://github.com/osm-search/Nominatim/blob/2ddb19c0b0f1775779700d5bfb6dce3dea747c64/src/nominatim_db/clicmd/index.py#L60
> 
> However.. if so, there would need to be an additional/separate check for pending postcode indexing.

@lonvia what do you think about a potential small refactor to add `has_pending_postcodes` -- or alternatively for `has_pending` to check the postcode data?

We might also want to add an index on `indexed_status` for the postcode table (I notice there's been some work on other indexes for that table recently - that seems unrelated though I think).

These would allow the CLI indexer to exit immediately when there's no work for it to do -- currently it attempts each item of work at least once, which is OK, but maybe wasteful.

lonvia:
This looks good. Sorry for the long wait.

> Is it wasteful that we call the various indexing methods at-least-once?

There is a short-cut in the code for a single rank round, so it is pretty fast to go through empty ranks.

> We might also want to add an index on indexed_status for the postcode table (I notice there's been some work on other indexes for that table recently - that seems unrelated though I think).

In theory, I agree. In practice, it doesn't look like it is much of an issue to go through the 3M lines sequentially. Even our slowest server takes only a second to do a sequential scan. Interesting.

What _should_ be changed though is the order of indexing in the rank index here

https://github.com/osm-search/Nominatim/blob/c81fb58b6345c1d34aaca3ffe05b5045c48e31a3/src/nominatim_db/clicmd/index.py#L65

If rank 0 is requested, it should be done last.

jayaddison:
That's OK - thanks again @lonvia.  I'll take another look at the indexing-order issue for rank 0 soon - that may also be a few days though.

jayaddison:
@lonvia I'm taking another read into the code for this at the moment.  When you mention the zero-rank special case:

> If rank 0 is requested, it should be done last.

Does this mean that if we request ranks `0..30` using the CLI, then:

- 1..30 should be indexed first
- _then_ postcodes should be indexed
- _then_ rank 0 should be indexed

(the reason I ask is that I didn't think that rank-zero places depend on postcodes -- and I think that [`Indexer.index_by_rank` already internally has the logic to run rank=0 last](https://github.com/osm-search/Nominatim/blob/f11b21b0ba6272ca01dd7b7754c7bb5a2d280d59/src/nominatim_db/indexer/indexer.py#L76-L77), so I am wondering if the scenario is already accommodated)

lonvia:
Ah, you are right, I was looking at the wrong piece of code. Everything good then.

jayaddison:
Ok, thank you :+1:

PR #3976, opened 2026-02-08
phase 1: intersection search

OpenSourceEverything:


lonvia:
I'm sorry but query parsing in Nominatim isn't as easy as throwing a couple of regexes on the text. You have to go through the the whole parsing pipeline. Query parsing in Nominatim doesn't just generate one single search, it creates a whole list of possible interpretations of the query. An IntersectionSearch would just be one of many options.

Also: you've ignore the PR template and in particular the question of how much AI was used in generating this PR.

PR #3979, opened 2026-02-09
Indexer: relocate zero-ranked-address indexing

jayaddison:
## Summary
Extract the special-case handling of rank zero so that it doesn't occur automagically when rank 30 is requested/indexed.

Context / ref: https://github.com/osm-search/Nominatim/pull/3971#pullrequestreview-3772308501

NB: I considered creating a separate method called something like `index_environs` -- but with a less-urban connotation noun than `environs` -- to make this more distinctive than yet another `index_rank` call.  However: the construction of a redundant/duplicate name analyzer within that method felt inelegant, too.

## AI usage
None

### DIFF
--- a/src/nominatim_db/indexer/indexer.py
+++ b/src/nominatim_db/indexer/indexer.py
@@ -56,7 +56,7 @@ class Indexer:
                         cur.execute('ANALYZE')
 
             while True:
-                if await self.index_by_rank(0, 4) > 0:
+                if await self.index_by_rank(1, 4) > 0:
                     _analyze()
 
                 if await self.index_boundaries() > 100:
@@ -67,6 +67,9 @@ class Indexer:
 
                 if await self.index_by_rank(26, 30) > 1000:
                     _analyze()
+
+                # Special case: rank zero depends on the previously-indexed [1..30] ranks
+                await self.index_by_rank(0, 0)
 
                 if await self.index_postcodes() > 100:
                     _analyze()
@@ -147,8 +150,11 @@ class Indexer:
                 total += await self._index(runners.RankRunner(rank, analyzer),
                                            batch=batch, total_tuples=total_tuples.get(rank, 0))
 
-            if maxrank == 30:
+            # Special case: rank zero depends on ranks [1..30]
+            if minrank == 0:
                 total += await self._index(runners.RankRunner(0, analyzer))
+
+            if maxrank == 30:
                 total += await self._index(runners.InterpolationRunner(analyzer), batch=20)
 
         return total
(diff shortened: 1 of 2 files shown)


src/nominatim_db/indexer/indexer.py
```diff
@@ -68,6 +68,10 @@ def _analyze() -> None:
                 if await self.index_by_rank(26, 30) > 1000:
                     _analyze()
 
+                # Special case: rank zero depends on the previously-indexed [1..30] ranks
+                if await self.index_by_rank(0, 0) > 0:
+                    _analyze()
```
lonvia: Another analyze isn't really needed here. You can just call the function ignoring the return value.

test/python/indexer/test_indexing.py
```diff
@@ -225,6 +225,7 @@ async def test_index_partial_with_30(test_db, threads, test_tokenizer):
 
     idx = indexer.Indexer('dbname=test_nominatim_python_unittest', test_tokenizer, threads)
     await idx.index_by_rank(28, 30)
+    await idx.index_by_rank(0, 0)
```
lonvia: Lets adapt this test the other way around and check that indeed rank 0 isn't indexed anymore. Then add another test, that checks that partial indexing with minrank = 0 does index rank 0.

jayaddison: Looking again at the other tests, there is coverage of that scenario - for example, [`test_index_all_by_rank` creates places from rank `0..30` inclusive](https://github.com/osm-search/Nominatim/blob/615804b1b3bb49b2585ea183d468822abaac44e1/test/python/indexer/test_indexing.py#L162-L163), and then [indexes from `minrank=0`](https://github.com/osm-search/Nominatim/blob/615804b1b3bb49b2585ea183d468822abaac44e1/test/python/indexer/test_indexing.py#L170).

That might not be obvious from the immediate diff of this change, and from a strict reading I could understand wanting not to reduce the coverage provided by `test_index_partial_with_30`.  Let me know what you think - it's possible I'm just being lazy/avoiding writing an additional test case.

test/python/indexer/test_indexing.py
```diff
@@ -226,13 +226,17 @@ async def test_index_partial_with_30(test_db, threads, test_tokenizer):
...
+    assert test_db.placex_unindexed() == 28
     assert test_db.osmline_unindexed() == 0
 
     assert test_db.scalar("""
                     SELECT count(*) FROM placex
                       WHERE indexed_status = 0 AND rank_address between 1 and 27""") == 0
 
+    assert test_db.scalar("""
+                    SELECT count(*) FROM placex
+                      WHERE indexed_status > 0 AND rank_address = 0""") > 0
```
jayaddison: This may have been a thinko - I probably should've used the same `indexed_status = 0` condition as the check for no successfully-indexed items between `1..27`.

jayaddison: (in fact: it's simpler to combine/merge it into the `1..27` range-not-indexed check; I'll do that shortly)

Issue #3993, opened 2026-02-20
PostcodeCollector's executemany is too slow for GB postcode file


mtmail:
**Describe the bug**

The new logic creating geometries around postcode centroids is much slower than in the past. I stopped my import on very capable hardware after 9 hours stuck in `Calculate postcodes`.

When `src/nominatim_db/tools/postcodes.py` reads a `gb_postcodes.csv` file it runs 1.7 million SQL INSERT statements using `cursor.executemany`. Each statement casts the centroid to geography type, then adds a ST_Buffer.

It might be faster to first insert the rows via COPY (`cursor.copy_from()` ?) and then add the extend in a single execute query. Pre-calculating one extend per country isn't precise enough but maybe there's a computationally faster approach caching by latitude or such. Or importing into a temporary table before adding extends.

**To Reproduce**

On current master I imported Ireland + Northern Ireland with `-v` (verbose). The setup is stuck at
```
Using external postcode file 'gb_postcodes.csv.gz'.
Processing country 'gb' (1758110 added, 0 deleted).
```
and I stopped the script after 9 hours.

Postgresql running queries shows
```sql
INSERT INTO location_postcodes                                                               
(country_code, rank_search, postcode, centroid, geometry)           
VALUES ('gb', 23, $1,                                                 
        ST_SetSRID(ST_MakePoint($2, $3), 4326),                       
        expand_by_meters(ST_SetSRID(ST_MakePoint($2, $3), 4326), 700));
```
each with minimal query time, e.g. `00:00:00.0007`, and that number updates. I can' see progress, the queries run in a transaction in a session I can't access, so the `location_postcodes` is still empty for me.

`EXPLAIN` isn't helpful here
```
                           QUERY PLAN
----------------------------------------------------------------
 Insert on location_postcodes  (cost=0.00..0.01 rows=0 width=0)
   ->  Result  (cost=0.00..0.01 rows=1 width=144)
```

```
select country_code, count(*) from placex group by country_code;
 country_code |  count
--------------+---------
 es           |       1
 im           |       2
 ie           | 1058239
 fr           |       4
              |      10
 gb           |  246895
```

**Software Environment (please complete the following information):**
- Nominatim version: master (commit `fbe0be9`)
- Postgresql version: 18.2
- Postgis version: 3.6
- OS: Ubuntu 24

**Hardware Configuration (please complete the following information):**
- RAM: 128
- number of CPUs: 20
- type and size of disks: 1TB Nvme

[...]

PR #3994, opened 2026-02-20
postcode import: make query planner use the index on location_postcodes

mtmail:
Fix for https://github.com/osm-search/Nominatim/issues/3993

## Summary

`_PostcodeCollector.commit()` (`src/nominatim_db/tools/postcodes.py`) has a loop of INSERT sql statements using Psyops' `executemany` which runs all inside a transaction. Before the loop it's possible that Postgresql query planner
sees 0 postcodes in the `location_postcodes` table and has no statistics.

Each INSERT runs a trigger (`lib-sql/functions/postcode_triggers.sql`) which checks if a postcode already exists
in the database. It runs basically `SELECT [...] FROM location_postcodes WHERE country_code = [...] AND postcode = postcode`. The index (`idx_location_postcodes_postcode`) would be ideal but Postgresql chooses a sequential
scan.

When importing one UK county I saw the INSERTs became slower over time. The "Calculate postcodes" step never
finished.

| Rows inserted | Time per INSERT | Cumulative wall time |
|---------------|----------------:|---------------------:|
| 1             | 0.5ms           | 0s                   |
| 10,000        | ~0.8ms          | 4s                   |
| 20,000        | ~1.4ms          | 16s                  |
| 40,000        | ~2.9ms          | 59s                  |
| 60,000        | ~4.3ms          | 131s                 |
| 80,000        | ~5.8ms          | 232s                 |
| 1,700,000     | ~120ms (proj.)  | ~45 hours (proj.)    |

With this PR the same import on a fresh database takes about 2 minutes
```
# https://en.wikipedia.org/wiki/Rutland
nominatim import -v --osm-file rutland-latest.osm.pbf
...
2026-02-20 21:46:51: Calculate postcodes
2026-02-20 21:46:51: Using external postcode file 'nominatim-project/gb_postcodes.csv'.
2026-02-20 21:46:58: Processing country 'gb' (1756829 added, 0 deleted).
2026-02-20 21:49:07: Indexing places
...
```

It's possible importing one UK county or Ireland was an edge case and the same issue wouldn't happen with
a full planet installation, or when importing another country.

## AI usage
AI helped me summarize the postgresql logfile and found the pattern that the individual INSERT statement
duration increases over time. It couldn't explain it though. It helped me write a test script importing the same
file with similar methods (`executemany`) but the script was always fast and no amount of tweaking made
it reproduce the issue. It advised me against the change in this PR and suggested instead to disable and
recreate the trigger during the `executemany` insert, or to drop and recreate the index. The `commit()`
runs many times, looping through countries, that wouldn't have been pragmatic.

### DIFF
--- a/src/nominatim_db/tools/postcodes.py
+++ b/src/nominatim_db/tools/postcodes.py
@@ -102,6 +102,13 @@ class _PostcodeCollector:
 
         with conn.cursor() as cur:
             if to_add:
+                # postcodes_insert trigger checks for duplicates via
+                # SELECT on (postcode, country_code). On an initial import the
+                # table stats can be 0 rows, so the query planner ignores the
+                # index. All inserts in executemany are in a transaction and
+                # postgres doesn't update the table state. The 'ANALYSE' after
+                # is too late.
+                cur.execute("SET LOCAL enable_seqscan = off")
                 cur.executemany(pysql.SQL(
                     """INSERT INTO location_postcodes
                          (country_code, rank_search, postcode, centroid, geometry)
@@ -112,6 +119,7 @@ class _PostcodeCollector:
                                 pysql.Literal(_extent_to_rank(self.extent)),
                                 pysql.Literal(self.extent)),
                     to_add)
+                cur.execute("SET LOCAL enable_seqscan = on")
             if to_delete:
                 cur.execute("""DELETE FROM location_postcodes
                                WHERE country_code = %s and postcode = any(%s)


lonvia:
Thanks. That's one way to fix it.

I did in fact some experiments for a special first-import mode yesterday. Let me do some experiments if that also fixes the problem. It has a few other optimizations as well.

lonvia:
Thanks but I went for the other solution.

PR #4012, opened 2026-03-04
Move interpolations into a separate table

lonvia:
This moves interpolation ways out of the place table into its own place_interpolation table. This simplifies the SQL functions around updates and also makes import/update slightly faster because the code no longer needs the large planet_osm_ways table to find nodes for an interpolation.

Also updates the schema description in the developer section of the docs.


src/nominatim_db/tools/migration.py
```diff
@@ -350,3 +350,73 @@ def create_place_postcode_table(conn: Connection, config: Configuration, **_: An
...
+
+    if mutable and not has_place_table:
+        # create tables
+        conn.execute("""
+            CREATE TABLE place_interpolation (
+              osm_id BIGINT NOT NULL,
+              type TEXT NOT NULL,
+              address HSTORE,
+              nodes BIGINT[] NOT NULL,
+              geometry Geometry(LineString, 4326)
```
mtmail: Maybe uppercase GEOMETRY here

test/python/conftest.py
```diff
@@ -246,6 +246,35 @@ def _insert(osm_type='N', osm_id=None, postcode=None, country=None,
...
+@pytest.fixture
+def place_interpolation_table(temp_db_with_extensions, table_factory):
+    """ Create an empty version of the place_interpolation table.
+    """
+    table_factory('place_interpolation',
+                  """osm_id bigint NOT NULL,
+                     type TEXT,
+                     address HSTORE,
+                     nodes BIGINT[],
+                     geometry Geometry(Geometry, 4326)""")
```
mtmail: maybe uppercase the column type here. `geometry GEOMETRY(Geometry, 4326)` isn't much easier to understand but it'd be a hint

src/nominatim_db/tools/migration.py
lonvia: Fixed

test/python/conftest.py
lonvia: Fixed that and all the other type definitions in that file.

PR #2689, opened 2022-05-02
Accept any OSM type in street member of associatedStreet

lonvia:
The French community has started putting pedestrian areas into associatedStreet relations which are mapped as multipolygons and consequently as relations. Nominatim has so far assumed that the object is a way as per Wiki documentation. That leads to a couple of interesting bugs with the assignment of the parent street. This change makes sure that the type is properly checked. Given that it doesn't make much of a difference, it allows that the 'street' member may now be of any type. The subsequent lookup in placex guarantees that the referenced OSM object is indeed a street.

Fixes #2669.

PR #4015, opened 2026-03-05
Re-parent house numbers when associatedStreet relations change

### LINKED ISSUE #3166: Changes to `associatedStreet` relations don't trigger update of contained house numbers
While trying to fix some errors reported by the Nominatim-Data-Analyser, I noticed two closely related issues in the context of `associatedStreet`  relations:
1. There are a few addresses in Ellikon, e.g. [Hallgartenstrasse 10][1], where the Nominatim-Data-Analyser currently reports an error because the street name ("Hallgartenstrasse") doesn't seem to match the parent name ("Zugwisenweg"). The tagging seems correct for me, "Hallgartenstrasse" exists, as does "Zugwisenweg", but the latter is pretty far away. The building is part of an `associatedStreet` relation which also looks fine. When I first tried to fix the error, I noticed that [the `associatedStreet` relation][4] actually contained at least a part of "Zugwisenweg", so [I removed it][5], however, that did not seem to have an effect. The number of reported errors in the area seems to have gone down though in the meantime, but that's probably caused by other changes to the buildings that have been done in the meantime which probably caused them to be updated in the Nominatim database. When searching for the address that is tagged on the building, [it isn't found][2]. When searching using the parent from the error, [it is found however][3].
2. The Nominatim-Data-Analyser also reports an error for [a building in Gundetswil][6], saying that the parent ("Hauptstrasse") doesn't match the street ("Püntstrasse").  When I tried to fix the error, I noticed that the building actually was part of an [`associatedStreet` relation for "Hauptstrasse"][10], [so I removed it][11], but the error persists. Searching for the building using "Püntstrasse" [doesn't seem to find it][7], [neither does searching for it using "Hauptstrasse"][8] (I'm pretty sure the latter worked though some time ago, and even found multiple hits). [Reverse geocoding looks fine for me][9].

Since the popularity of the [`associatedStreet` relation seems to be dropping][12] and as it even seems [deprecated in some countries][13], attempting to fix this is likely not worth the effort, so the intention of this issue is mostly to just document some effects of the current behavior for others and my future self.

[1]: https://www.openstreetmap.org/way/281234381
[2]: https://nominatim.osm.org/ui/search.html?q=Hallgartenstrasse+10%2C+Ellikon
[3]: https://nominatim.osm.org/ui/search.html?q=Zugwisenweg+10%2C+Ellikon
[4]: https://www.openstreetmap.org/relation/3753531
[5]: https://www.openstreetmap.org/changeset/136120055
[6]: https://www.openstreetmap.org/way/160500398
[7]: https://nominatim.osm.org/ui/search.html?q=P%C3%BCntstrasse+4%2C+Gundetswil
[8]: https://nominatim.osm.org/ui/search.html?q=Hauptstrasse+4%2C+Gundetswil
[9]: https://nominatim.osm.org/ui/reverse.html?lat=47.54033&lon=8.81868&zoom=18
[10]: https://www.openstreetmap.org/relation/4185549
[11]: https://www.openstreetmap.org/changeset/136161150
[12]: https://wiki.openstreetmap.org/wiki/Relation:associatedStreet
[13]: https://wiki.openstreetmap.org/wiki/DE:Relation:associatedStreet

Itz-Agasta:
## Summary
When a mapper edits an `associatedStreet` relation on OSM , osm2pgsql updates the raw `planet_osm_rels` middle table, but Nominatim never watches that table for associatedStreet changes. The houses keep their stale `parent_place_id` indefinitely; the only accidental fix happens if someone separately edits the building itself.... Following the plan outlined in the issue discussion:

1. **Add  `place_associated_street` flex table**:  populated by osm2pgsql during import. Each row stores one relation member (`relation_id`, `member_type`, `member_id`, `member_role`). This replaces the need to parse the raw `planet_osm_rels` JSONB `members` column.

2. **Refactored `find_associated_street()`** : the SQL function that resolves  which street to assign to a house now uses a clean self-join on `place_associated_street` instead of parsing `planet_osm_rels`. 

3. **Update trigger** : `invalidate_associated_street_members()` fires on INSERT/DELETE of `place_associated_street` rows... each affected non-street member gets `indexed_status = 2`, so the next `nominatim index` recomputes its parent.

Backward compatibility is maintained  via:

```
{% if 'place_associated_street' in db.tables %}     -- new path
{% elif db.middle_db_format == '1' %}               -- legacy array format
{% else %}                                          -- JSONB middle table
{% endif %}
```

**NOTE:** I have added the Update trigger only on the new table... dbs using planet_osm_rels will still experience this bug.

Closes #3166

## Results

https://github.com/user-attachments/assets/473714fa-fae7-4a5b-b1e7-0fd8d1ba9ff3

## AI usage
The sequential cmds used in the results video to reproduce the bug and demonstrate the fix were suggested by AI after analyzing the Monaco tables in pg.

### DIFF
--- a/test/bdd/features/db/import/parenting.feature
+++ b/test/bdd/features/db/import/parenting.feature
@@ -295,12 +295,12 @@ Feature: Parenting of objects
          | W1  | highway | residential | foo  | 10,11    |
          | W2  | highway | residential | bar  | 20,21    |
         And the relations
-         | id | members         | tags+type |
+         | id | members            | tags+type        |
          | 1  | W1:street,N1:house | associatedStreet |
         When importing
         Then placex contains
          | object | parent_place_id |
-         | N1     | W1 |
+         | N1     | W1              |
 
     Scenario: Building without tags, closest street from center point
         Given the grid


lib-lua/themes/nominatim/init.lua
```diff
@@ -89,6 +89,17 @@ local table_definitions = {
...
             { column = 'nodes', method = 'gin' }
         }
+     },
+    place_associated_street = {
+        ids = { type = 'relation', id_column = 'relation_id' },
+        columns = {
+            { column = 'member_type', type = 'text', not_null = true },
+            { column = 'member_id', type = 'int8', not_null = true },
+            { column = 'member_role', type = 'text', not_null = true }
+        },
```
lonvia, line 99: This table is very cheap to create from an existing `planet_osm_rels` table. I tried it on a planet database using

```
CREATE TABLE place_addociated_street AS
  (SELECT id, x.type as member_Type, x.ref as member_id, x.role as member_role
   FROM planet_osm_rels, LATERAL jsonb_to_recordset(members) as x(ref bigint, role text, type text)   
  WHERE tags->>'type' = 'associatedStreet');
```

and it took about 10s.

So a migration is the better way to go here. It also mean we can get rid of all those if cases around the db_middle_format further down in placex_triggers.sql. Don't worry about migrating from a database with `db.middle_db_format == '1' `. The new format has been around for 2 years and was introduced before the last major version change. Just print a big fat warning and leave the table empty.

lib-lua/themes/nominatim/init.lua
```diff
@@ -674,6 +685,16 @@ function module.process_way(object)
 end
 
 function module.process_relation(object)
+    if object.tags.type == 'associatedStreet' then
+        for _, member in ipairs(object.members) do
+            insert_row.place_associated_street{
+                member_type = member.type:upper(),
+                member_id = member.ref,
+                member_role = member.role or ''
```
lonvia: There are only two [officially supported roles](https://wiki.openstreetmap.org/wiki/Relation:associatedStreet) but about [0.5% of members use some other role](https://taginfo.openstreetmap.org/relations/associatedStreet#roles). Probably best to simply drop the members with unknown roles.

There is another interesting observation on that wiki page: apparently there are quite a few associatedStreet relations without a single street. Those would be useless to Nominatim as well. We could drop those directly here as well and not have them take up any space.

lib-sql/tables.sql
```diff
@@ -21,14 +21,17 @@ CREATE SEQUENCE seq_place start 1;
...
 CREATE INDEX planet_osm_rels_parts_associated_idx ON planet_osm_rels USING gin(parts)
   {{db.tablespace.address_index}}
   WHERE tags @> ARRAY['associatedStreet'];
-{% else %}
+  {% else %}
 CREATE INDEX planet_osm_rels_relation_members_idx ON planet_osm_rels USING gin(planet_osm_member_ids(members, 'R'::character(1)))
   WITH (fastupdate=off)
   {{db.tablespace.address_index}};
+  {% endif %}
 {% endif %}
```
lonvia: This can just be deleted completely. The code is only executed when a new database is created. A new database will always have the new associated street table.

lib-sql/functions/placex_triggers.sql
```diff
@@ -122,15 +122,49 @@ CREATE OR REPLACE FUNCTION find_associated_street(poi_osm_type CHAR(1),
...
+      FROM place_associated_street h
+      JOIN place_associated_street s
+           ON h.relation_id = s.relation_id AND s.member_role = 'street'
+      JOIN placex p
+           ON p.osm_type = s.member_type::char(1)
+              AND p.osm_id = s.member_id
+              AND p.name IS NOT NULL
+              AND p.rank_search BETWEEN 26 AND 27
+     WHERE h.member_type = poi_osm_type
+       AND h.member_id = poi_osm_id
```
lonvia, line 144: This is missing a `h.member_role = 'house'`.

lib-sql/functions/placex_triggers.sql
```diff
@@ -147,9 +181,6 @@ BEGIN
                  and name is not null
                  and rank_search between 26 and 27
         LOOP
-          -- Find the closest 'street' member.
-          -- Avoid distance computation for the frequent case where there is
-          -- only one street member.
```
lonvia, line 152: Is that the IDE removing comments in random parts of the code? I've seen that with other PRs as well and wonder what it is all about.

lib-sql/functions/placex_triggers.sql
```diff
@@ -1416,3 +1444,46 @@ BEGIN
...
+BEGIN
+  IF TG_OP = 'DELETE' THEN
+    m_type := OLD.member_type;
+    m_id   := OLD.member_id;
+    m_role := OLD.member_role;
+  ELSE
+    m_type := NEW.member_type;
+    m_id   := NEW.member_id;
+    m_role := NEW.member_role;
+  END IF;
```
lonvia, line 1412: This would be a bit less verbose if you define a record variable like this in declare:

    object RECORD;

and then just assign NEW/OLD to `object`. Also saves the recheck below because you can just do `RETURN object`.

lib-sql/functions/placex_triggers.sql
Itz-Agasta: No no... this was intentional....We were essentially doing the same thing across three if-else blocks with just different conditions..... So I removed the comments under the other two and kept the explanation in the first if block.....

Image

Itz-Agasta: > I've seen that with other PRs as well and wonder what it is all about.

Idt the IDE is the issue.... IDEs don’t usually remove comments if an auto-formatter is enabled. Ig  likely that sometimes AI agents hallucinate or accidentally delete comments maybe

lib-sql/functions/placex_triggers.sql
Itz-Agasta: ok wow .... Thats samrt

lib-lua/themes/nominatim/init.lua
Itz-Agasta: I  started with looking for only `house` & `street` but  2 tests  in import parenting  use members with no explicit role (e.g., W1:street,N1 where N1 has no :house suffix)... so had to consider it too 

Image

- 'street' ->  keep
- 'house' -> keep  
- (empty/no role) -> treat as 'house' 

shouldt it be better if we ignore empty ones too... i mean  those are useless to Nominatim too nha just like other roles ?

Itz-Agasta:
Tell me what y recommend for the `empty-string` case....then i will squash & rebase the pr

lonvia:
Lets fix the test for the empty-string case and disallow it.

lib-sql/functions/placex_triggers.sql
```diff
@@ -122,15 +122,50 @@ CREATE OR REPLACE FUNCTION find_associated_street(poi_osm_type CHAR(1),
...
+      new_distance := ST_Distance(parent.geometry, bbox);
+      IF new_distance < distance THEN
+        distance := new_distance;
+        result := parent.place_id;
+        waygeom := parent.geometry;
+      END IF;
+    END IF;
+  END LOOP;
+
+{% elif db.middle_db_format == '1' %}
```
lonvia: Now that you have a migration, the old ways of looking up associated streets can be both removed.

lib-sql/functions/placex_triggers.sql
```diff
@@ -1418,3 +1447,35 @@ BEGIN
 END;
 $$
 LANGUAGE plpgsql;
+
+
+-- Invalidates house members of associatedStreet relations
+-- whenever the place_associated_street table is modified.
+-- osm2pgsql flex handles updates as DELETE-all + re-INSERT, so each
+-- row-level trigger call covers exactly one member.
+{% if 'place_associated_street' in db.tables %}
```
lonvia: Here too is no need for conditionals.

lib-sql/table-triggers.sql
```diff
@@ -36,3 +36,11 @@ CREATE TRIGGER place_interpolation_before_insert BEFORE INSERT ON place_interpol
     FOR EACH ROW EXECUTE PROCEDURE place_interpolation_insert();
 CREATE TRIGGER place_interpolation_before_delete BEFORE DELETE ON place_interpolation
     FOR EACH ROW EXECUTE PROCEDURE place_interpolation_delete();
+
+-- Propagate changes to associatedStreet relations to house members
+-- so that the indexer re-computes their parent_place_id.
+{% if 'place_associated_street' in db.tables %}
```
lonvia: And here neither.

src/nominatim_db/tools/migration.py
```diff
@@ -437,3 +437,56 @@ def backfill_importance(conn: Connection, **_: Any) -> None:
...
+              member_id   BIGINT NOT NULL,
+              member_role TEXT NOT NULL
+            )
+        """)
+
+        # Populate from planet_osm_rels if the JSONB members format is in use.
+        cur.execute("""
+            SELECT data_type FROM information_schema.columns
+            WHERE table_name = 'planet_osm_rels' AND column_name = 'members'
+        """)
```
lonvia: The easier way to do that is to look into the osm2pgsql properties table, see https://github.com/osm-search/Nominatim/blob/2097a3676106a4cd0c8f1d5210bc344bacec0f42/src/nominatim_db/db/sql_preprocessor.py#L43

test/bdd/features/db/import/parenting.feature
```diff
@@ -245,8 +245,8 @@ Feature: Parenting of objects
          | W1  | highway | residential | foo  | 10,11 |
          | W2  | highway | service     | bar  | 10,12 |
         And the relations
-         | id | members            | tags+type |
-         | 1  | W1:street,N1,N2,N3 | associatedStreet |
+         | id | members                          | tags+type |
+         | 1  | W1:street,N1:house,N2:house,N3:house | associatedStreet |
```
lonvia, line 249: Nitpicking here but can you make the columns align nicely in the table.

lib-sql/functions/placex_triggers.sql
```diff
@@ -122,15 +122,50 @@ CREATE OR REPLACE FUNCTION find_associated_street(poi_osm_type CHAR(1),
...
+{% endif %}
+BEGIN
+{% if 'place_associated_street' in db.tables %}
+  FOR parent IN
+    SELECT p.place_id, p.geometry
+      FROM place_associated_street h
+      JOIN place_associated_street s
+           ON h.relation_id = s.relation_id AND s.member_role = 'street'
+      JOIN placex p
+           ON p.osm_type = s.member_type::char(1)
```
lonvia: This will cause trouble because `placex` only has [partial indexes over the different osm types](https://github.com/osm-search/Nominatim/blob/d43e95f177040ed98c4d6aeddb9bc24d149b39c7/lib-sql/tables/placex.sql#L32). Postgresql cannot use such indexes with a variable type.

Nominatim should be able to handle streets of way type and [relation type](https://github.com/osm-search/Nominatim/pull/2689). It might be enough to add a `p.osm_type IN ('W', 'R')` condition to get the query planner to use the indexes but I'm not sure. You'd need to use EXPLAIN to see what happens.

lonvia: It _is_ causing trouble on a planet DB.

As a way forward I suggest have two SELECT statements, one for ways and one for relations and UNION the two together. Looks to be running smoothly afterwards.

If we only support 'W' and 'R' here (which is sensible) then we should also ignore the node type for streets in the lua script.

Itz-Agasta: wow fr..... Didn't realize i  was doing a full table scan :(

### Now
after


### Before
before

lonvia:
Planet import came through without issues and was even 45min faster. And there is no degradation with address search in France where most of the associatedStreet relations are.

Thank you.

Issue #3831, opened 2025-09-09
Full word matches don't work properly with Chinese characters


lonvia:
Searching for [小学校](https://nominatim.openstreetmap.org/ui/search.html?q=%E5%B0%8F%E5%AD%A6%E6%A0%A1) should yield a couple of full word matches, e.g. [this](https://nominatim.openstreetmap.org/ui/details.html?osmtype=N&osmid=12495752868&class=highway). Instead it seems to prefer random partial matches.

Oddly enough, there is a word entry for "小学校" which transliterates to "xiao". This needs investigation as well.

lonvia:
This is going to need some more fundamental changes to solve. Han script doesn't work well with the current search based on transliterated partials because too many han characters match the same transliteration.

lonvia:
I think the best way forward here is to exclude Chinese characters from transliteration for the internal index.

To make sure that Chinese names still can be found using Latin script, we'd need to add a [sanitizer](https://nominatim.org/release-docs/latest/customize/Tokenizers/#sanitizers) that adds a transliterated name variant for places that only have a name in Chinese characters.

Additional bonus of doing a manual transliteration beforehand is that we can actually take into account different transliterations for Japanese and Chinese usage. (Cantonese is a bit more tricky because the sanitizer may not have the necessary location information.)

wangshamian:
Yeah, exclude Chinese characters from transliteration and add special analyzer in token-analysis, my idea is that Chinese query characters be analyzed to Chinese phrases (using word segmentation algorithms e.g. jieba), then the Chinese phrases saved in Word Table.

PR #4017, opened 2026-03-06
Preserving orginal HAN characters for better results.

var12958:
## Summary
Chinese search queries split words into characters instead of searching as a single word. This causes wrong results. This PR preserves the original HAN characters before transliteration, so an exact match is found. (issue: #3831)

In "split_query()" in icu_tokenizer.py, the transliterator converts the Chinese words into space separated lexical representation. The "split_transliteration()" then splits that string into multiple tokens, so the real Han words are never preserved as a searchable unit.

The tokenizer was splitting Chinese words into pinyin tokens, so the original words were never searched. This PR change keeps the original Han word as an additional token before transliteration occurs. So now the system can match the full Chinese word directly, improving the rank and search results become more accurate.

## AI usage
I mainly used AI for help with CJK scripts and their Unicode concepts, since they are not the languages I usually work with. All the codes were implemented, tested and verified manually.

### DIFF
--- a/src/nominatim_api/search/icu_tokenizer.py
+++ b/src/nominatim_api/search/icu_tokenizer.py
@@ -11,10 +11,10 @@ from typing import Tuple, Dict, List, Optional, Iterator, Any, cast
 import dataclasses
 import difflib
 import re
+import unicodedata
 from itertools import zip_longest
 
 from icu import Transliterator
-
 import sqlalchemy as sa
 
 from ..errors import UsageError
@@ -27,6 +27,28 @@ from ..query_preprocessing.config import QueryConfig
 from ..query_preprocessing.base import QueryProcessingFunc
 from .query_analyzer_factory import AbstractQueryAnalyzer
 from .postcode_parser import PostcodeParser
+
+def _is_han_word(text: str) -> bool:
+    """ Return True if every character is a CJK Unified Ideograph.
+        Uses unicodedata.name() — covers all Han extensions automatically.
+        Fast-exits for empty text or text starting outside CJK codepoint
+        ranges, avoiding unicodedata lookup cost for Latin queries.
+    """
+    if not text or ord(text[0]) < 0x3400:
+        return False
+    return all(
+        unicodedata.name(ch, '').startswith('CJK UNIFIED IDEOGRAPH')
+        for ch in text
+    )
+
+def _is_japanese_script(text: str) -> bool:
+    """ Return True if text contains hiragana or katakana characters.
+        Reliable guard to distinguish Japanese from Chinese text.
+    """
+    return any(
+        unicodedata.name(ch, '').startswith(('HIRAGANA', 'KATAKANA'))
+        for ch in text
+    )
 
 
 DB_TO_TOKEN_TYPE = {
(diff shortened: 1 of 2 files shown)


lonvia:
This PR contains commits from the previous one. You need to create a new branch for a new PR.

My main question at the moment would be: how have you tested that the code works? Do you have examples where it improves results?

var12958:
I am sorry for the confusion. I will close this PR and open a new PR shortly. 

Regarding the testing, I verified what I have done by running both core functions is_han_word and is_japanese_script directly with many test cases. I have also attached images.

Screenshot 2026-03-09 064101

Screenshot 2026-03-09 064159

I will close this PR and open a new one shortly.

lonvia:
> Regarding the testing, I verified what I have done by running both core functions is_han_word and is_japanese_script directly with many test cases. I have also attached images.

This only answers the question if the functions do what you expect. My question is more: have you tested that your code solves the original issue from #3831.

var12958:
I am sorry because I misunderstood that testing only the logic of _is_han_word() and _is_japanese_script() was sufficient for the fix. Now I understand that real verification is a must to be at search level.

I have already set up local nominatim instance, I'll test various queries and update the results shortly.

Thanks again for pointing this out.

var12958:
Testing is now complete, and the results confirm that the fix works as expected.

Previously, the tokenizer split 小学校 into 3 separate tokens, which prevented matching the original Han word. After the fix, the original Han word is preserved as a single token while still providing transliteration tokens. This ensures that the original word could also participate in exact matching.

Test cases:

#1
Screenshot 2026-03-10 063306

Query: 小学校  
Language: Chinese (Han characters) 
Result: `<小学校`xiao`xue`xiao>` / `<小学校`小`學`校>` 
Shows the original Han word is preserved as a single token

#2
Screenshot 2026-03-10 063338

Query: 九龍
Language: Chinese (Traditional Han characters)
Result: <九龍`9long> / <九龍`9龍>
The tokenizer preserves the original Han token even when ICU normalization converts 九 to the numeral 9 in the transliteration.

#3
Screenshot 2026-03-10 063935

Query: ありがとう
Language: Japanese (Hiragana)
Result: <arigatou> / <ありがとう>
Shows that Hiragana queries are not affected, and also no Han token is added.

#4
Screenshot 2026-03-10 063956

Query: 香港 Hong Kong
Language: Mixed Chinese (Han) + English (Latin)
Result: <xiang gang hong kong> / <香 港 hong kong>
Shows the correct handling of mixed Chinese and English input.

#5
Screenshot 2026-03-10 064613

Query: 東京
Language: Japanese (Kanji / Han characters)
Result: <東京`dong`jing> / <東京`東`京>
Shows that fix also works for Japanese Kanji while preserving the original Han characters.

#6
Screenshot 2026-03-10 064011

Query: Kowloon
Language: English (Latin)
Result: <kowloon> / <kowloon>
Shows English input remains unchanged.

Note: Our current Hong Kong OSM dataset doesn't include a place named "小学校," so I am unable to show the ranking. However, the token stream confirms the fix is correctly working at tokenizer level (the original Han word is emitted as a single exact match token).  

Is this level of testing sufficient, or is a more thorough verification preferred? I am happy to extend the testing further if needed.

PR #4018, opened 2026-03-07
feat: accept OSM IDs in exclude_place_ids for server independent exclusions

### LINKED ISSUE #4006: exclude_place_ids should accept OSM IDs
[exclude_place_ids](https://nominatim.org/release-docs/develop/api/Search/#result-restriction) is the last parameter in the API that relies on the server-dependent `place_id`. This is an issue for osm.org because you might end up at different servers when sending a query to nominatim.openstreetmap.org. The place_id that comes back with one request might not be the same as in a subsequent request.

The exclude_place_ids parameter should be extended to accept server-independent identifiers. As a minimum `<osm type><osm_id` parameter (e.g `N3446644`) just like the details and lookup endpoint do. Interpolation and Tiger data should get excluded when the parent street is in the exclusion list.

For postcodes and countries we need to find a different solution later. That would be out of scope for now.

To implement this:
* accept OSM IDs for exclude_place_ids parameter (as well as place_ids)
* make sure the exclude_place_ids parameter is constructed using OSM IDs as far as possible (leave place_ids for postcodes)
* go through search code where excluded IDs are used and make sure they can handle other things than place IDs
* adapt documentation
* extend tests (including BDD tests) to cover the new format

As I expect a PR for this change to be slightly longish, I won't mark this as a 'new contributor opportunity' but in general it should be doable with limited experience with the code. The foundations for using OSM IDs in the exclusion are there. If you struggle with extending the search SQL, make a PR that has the rest in place and ask for advice.

kad-link:
## Summary

- Allows `exclude_place_ids` to accept OSMIDs (e.g. `N1234`, `W12345`, `R123456`) along with usual placeIDs
- Resolves OSMIDs to place IDs via placex lookup before search execution so later SQL queries remain unchanged
  Since the database doesn't understand N12345 (OSMIDs) and rather than changing the SQL queries in the 6 search files (`address_search.py`, `country_search.py`, `near_search.py`, `place_search.py`, `poi_search.py`, `postcode_search.py`) and join in the placex table, resolving them to local placeIDs of the servers in `geocoder.py` was a more considerate solution.
- Excludes interpolation/tiger results when parent street is in the exclusion list
  The `_make_interpolation_subquery` subquery builds the Sql for both interpolation and tiger data. So the row itself and the  rows whose parent street is in the list are excluded.  
- Use OSMIDs instead of placeIDs in `moreurl` in `server_glue.py` for server independence.
  The More results button now sends a `moreurl` that carries OSMIDs with it

Now I will quickly explain the file changes I have made: 
- `types.py` : I changed `excluded` field to `List[PlaceRef]`and added `format_excluded()` parser for mixed input (for both placeIDs and OSMIDs) and added `excluded_place_ids` property
- `geocoder.py` : added `_resolve_excluded_osm_ids()` to convert OSM IDs to place IDs before searches run
- Search files: modified bind params to use `excluded_place_ids`and added parent street exclusion for interpolations. Here I also fixed missing exclusion filter in `poi_search.py`
- `server_glue.py` and `helpers.py`:  made it use OSM IDs in moreurl 

I will come up with the BDD tests plan and docs updation shortly. 
I ensured `make tests` have passed: 
image
image
image

I also manually checked the implementations (of course in a single server only)....

I first searched for 5 avenues in Monaco:
image

Here,  exclusion by placeIDs itself
image

By OSMIDs ...
image

And a mix...
image

I also checked the moreurl whether it was carrying the OSMIDs ...
image
image

The only thing I couldn't test is the interpolation and tiger data since I am working on the tiny Monaco dataset. But I will make sure they get covered during the BDD tests. 

This issue really helped me get with the roots of Nominatim and really happy to tackle this one. 

Resolves #4006 [excluding BDD tests and docs updation]

## AI usage

NIL

### DIFF
--- a/src/nominatim_api/search/geocoder.py
+++ b/src/nominatim_api/search/geocoder.py
@@ -45,7 +45,7 @@ class ForwardGeocoder:
 
     async def _resolve_excluded_osm_ids(self) -> None:
         """ Resolve any OsmID entries in the excluded list to PlaceID entries
-            by looking them up in the placex table.
+            by looking them up in the placex and location_postcode table.
         """
         excluded = self.params.excluded
         if not excluded or all(isinstance(e, PlaceID) for e in excluded):
@@ -63,6 +63,13 @@ class ForwardGeocoder:
             sql = sa.select(t.c.place_id).where(sa.or_(*conditions))
             place_ids.extend(PlaceID(row.place_id)
                              for row in await self.conn.execute(sql))
+
+            relation_ids = [oid.osm_id for oid in osm_ids if oid.osm_type == 'R']
+            if relation_ids:
+                p = self.conn.t.postcode
+                sql = sa.select(p.c.place_id).where(p.c.osm_id.in_(relation_ids))
+                place_ids.extend(PlaceID(row.place_id)
+                                 for row in await self.conn.execute(sql))
 
         self.params.excluded = place_ids
 
--- a/test/bdd/features/api/search/params.feature
+++ b/test/bdd/features/api/search/params.feature
@@ -258,6 +258,18 @@ Feature: Search queries
           | category | type |
           | place    | village |
 
+    Scenario: Exclude a specific place using an OSM ID
+        When geocoding "Malbun"
+        Then result 0 contains
+          | category | type |
+          | place | village |
+        When geocoding "Malbun"
+          | exclude_place_ids |
+          | N347290636 |
+        Then result 0 contains
+          | category | 
+          | landuse | 
+
     Scenario Outline: Search with polygon threshold (json)
         When sending v1/search with format json
           | q           | polygon_geojson | polygon_threshold |


src/nominatim_api/types.py
```diff
@@ -583,12 +598,22 @@ class SearchDetails(LookupDetails):
     """
 
     def __post_init__(self) -> None:
+        if self.excluded and not all(isinstance(e, (PlaceID, OsmID)) for e in self.excluded):
+            self.excluded = format_excluded(self.excluded)
```
lonvia: This shouldn't be necessary. The `from_kwargs()` function takes care of calling the transform function. And when instantiating directly, then the type info needs to be respected.

src/nominatim_api/search/geocoder.py
```diff
@@ -41,6 +43,29 @@ def limit(self) -> int:
...
 
+    async def _resolve_excluded_osm_ids(self) -> None:
+        """ Resolve any OsmID entries in the excluded list to PlaceID entries
+            by looking them up in the placex table.
+        """
+        excluded = self.params.excluded
+        if not excluded or all(isinstance(e, PlaceID) for e in excluded):
+            return
+
+        place_ids = [e.place_id for e in excluded if isinstance(e, PlaceID)]
```
lonvia: No need to extract the id integer here. If place_ids remains a list[PlaceID], you can directly assign it in line 67 and save one list allocation.

src/nominatim_api/search/geocoder.py
```diff
@@ -41,6 +43,29 @@ def limit(self) -> int:
...
+
+        if osm_ids:
+            t = self.conn.t.placex
+            conditions = [
+                sa.and_(t.c.osm_type == oid.osm_type, t.c.osm_id == oid.osm_id)
+                for oid in osm_ids
+            ]
+            sql = sa.select(t.c.place_id).where(sa.or_(*conditions))
+            for row in await self.conn.execute(sql):
+                place_ids.append(row.place_id)
```
lonvia: This can be done a little bit more efficiently with extend() and list comprehensions:

```
place_ids.extend(PlaceID(r.place_id) for r in await self.conn.execute(sql))
```

src/nominatim_api/v1/helpers.py
```diff
@@ -85,7 +85,10 @@ def extend_query_parts(queryparts: Dict[str, Any], details: Dict[str, Any],
     if parsed.countries:
         queryparts['countrycodes'] = ','.join(parsed.countries)
     queryparts['exclude_place_ids'] = \
-        ','.join(chain(excluded, map(str, (e for e in parsed.excluded if e > 0))))
+        ','.join(chain(excluded,
+                       (str(e.place_id) if isinstance(e, PlaceID)
+                        else f"{e.osm_type}{e.osm_id}"
+                        for e in parsed.excluded)))
```
lonvia: This can be made prettier by implementing the `__str__` function for PlaceID and OsmID to give the proper representation.

Also, I'm trying to remember why there is a `e > 0` condition: there are some cases where the place id is 0, corresponding to "unknown".

src/nominatim_api/types.py
kad-link: Oh, I somehow overlooked it , thanks for pointing it. Its anyways redundant. I will remove these lines

src/nominatim_api/v1/helpers.py
kad-link: > This can be made prettier by implementing the `__str__` function for PlaceID and OsmID to give the proper representation.

yea, will make the implementation logic cleaner by defining in their respective classes rather than scattered at call sites.

kad-link: > Also, I'm trying to remember why there is a `e > 0` condition: there are some cases where the place id is 0, corresponding to "unknown".

yea e>0 was to prevent placeid=0 from appearing it into the URL. ok so now since they are objects (placeid or osmid), maybe I can add the same checking inside format_excluded().

src/nominatim_api/search/geocoder.py
kad-link: sure. no extra list allocation.

src/nominatim_api/search/geocoder.py
kad-link: sure 👍

kad-link:
just resolved the merge conflict. will get onto work with the docs and testing phase now 👍

lonvia:
Can you please rebase on the master branch instead of merging. This gives a cleaner commit history. You can then force-push the changes on this branch.

kad-link:
hi @lonvia, I have added unit and BDD tests that test `format_excluded()` aggressively. 

- Unit Tests: Firstly, added a class `TestFormatExcluded` so that the core logic remains at one place. A method to test empty values was the first choice, then tested valid IDs, then onto invalid IDs and lastly invalid list element type. 
So here, the 2 error gates [here](https://github.com/kad-link/Nominatim/blob/53b5dd1aeb76d1c38b7da1d44e1aca8255a6b2f5/src/nominatim_api/types.py#L445) and [here](https://github.com/kad-link/Nominatim/blob/53b5dd1aeb76d1c38b7da1d44e1aca8255a6b2f5/src/nominatim_api/types.py#L443) that I opened in `format_excluded()` are strictly being covered. 

- BDD Tests: Edited the already existing positive test [(this)](https://github.com/kad-link/Nominatim/blob/53b5dd1aeb76d1c38b7da1d44e1aca8255a6b2f5/test/bdd/features/api/search/simple.feature#L81) slightly to adapt to both kinds of IDs. Then for the negative test apart from [this](https://github.com/osm-search/Nominatim/blob/b147c1c6eea147c3324ad9bd3a85453af9314f3e/test/bdd/features/api/search/simple.feature#L91) , I added one more to strictly adhere to `format_excluded()`.

I tried covering maximum number of edge cases but with your suggestions I can add more too. 

One more thing I thought of asking you lately is right now the logic is forgiving. Should the API strictly take in inputs that follow the exact syntax? For example, a PlaceID like 00123 would be parsed to 123, n101 would be parsed to N101.  So should such situations be entertained or should we make the parsing stricter ?

src/nominatim_api/v1/server_glue.py
```diff
@@ -374,7 +374,11 @@ async def search_endpoint(api: NominatimAPIAsync, params: ASGIAdaptor) -> Any:
                                    params.get('featureType', ''),
                                    params.get_bool('namedetails', False),
                                    params.get_bool('extratags', False),
-                                   (str(r.place_id) for r in results if r.place_id))
+                                   (f"{r.osm_object[0]}{r.osm_object[1]}"
+                                    if r.osm_object
+                                    else str(r.place_id)
+                                    for r in results
+                                    if r.osm_object or r.place_id))
```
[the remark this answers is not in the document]
lonvia: Copilot is right about the postcode table here. Although it mostly contains artificial postcodes, which are out of scope for now, there are also some original [OSM postcode areas](https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dpostal_code). They are always relations and can be identified by the `osm_id` column. If it isn't NULL, then it references a OSM relation.

Rather than excluding postcodes here though, I would suggest you add a lookup of OSM relations in location_postcodes in the `_resolve_excluded_osm_ids` function.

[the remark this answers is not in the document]
kad-link: Understood, I will leave server_glue.py untouched. Will update the `_resolve_excluded_osm_ids` function to handle postcodes too.

kad-link:
hi @lonvia , I updated `_resolve_excluded_osm_ids` function to do a lookup against the `location_postcode` database table for any OSM IDs where the `osm_type == 'R'`. Also added the BDD test scenario [link](https://github.com/osm-search/Nominatim/blob/2097a3676106a4cd0c8f1d5210bc344bacec0f42/test/bdd/features/api/search/params.feature#L249) as you suggested using Malbun. 



> Nominatim has traditionally been lenient about the format of its parameter. So it is good to be tolerant with the format.

I see... I left the parsing as it is so it stays consistent with this intention, thanks for clarifying :)

lonvia:
Looks good now. Thank you.

PR #4035, opened 2026-03-16
Fix: Enable qualifier searches for unnamed POIs near small places

### LINKED ISSUE #3750: Special phrases for finding unnamed objects in small places
We currently have to kinds of special phrases: "qualifiers" are added to the name and restrict the type of object ("King's Head hotel"), "near items" are for category searches in a certain place ("hotels in London"). This distinction works well for larger places but doesn't hold for smaller ones.

In smaller places there is often only one kind of POI (pub, church, cemetery, etc). These POIs often don't really have a name but are simply referred to by their category, only adding the name of the place for reference ('Kingston pub', 'cemetery Oxbridge', etc.). That's not well covered by the current special places search. We'd need to allow "qualifier" words with a near search for that.

Some ideas to restrict search to not end up with an impossibly long list of odd results again:

* only allow for singular qualifier terms
* restrict the search area (it is small places after all) and add a penalty when there is more than one POI of the category found
* add a penalty for named places in the result

agam1092005:
Fixes #3750 

Enable qualifier searches for unnamed POIs near small places.

Previously, queries like "Kingston pub" would fail to return unnamed amenities (pubs) located near a small place (Kingston village). The root cause was that qualifier+name search logic in SearchBuilder.build() only produced a 
PlaceSearch with the qualifier as a filter, which doesn't work for unnamed POIs near small places.

Changes:
New QualifierNearSearch class (qualifier_search.py): A dedicated search type for qualifier+address queries. Unlike 
NearSearch, it restricts inner results to actual places (rank_address < 30), only proceeds when there is a single matching place, and applies accuracy penalties for named POIs (+0.4) and multiple results (+0.2).

_build_qualifier_address_search() method in db_search_builder.py: When a query has qualifiers and a name but no near items, builds a QualifierNearSearch alongside the existing PlaceSearch. Temporarily modifies sdata in place (clearing qualifiers, isolating rankings) and restores it before returning. Only wraps PlaceSearch results, skipping AddressSearch.

No changes to NearSearch: All custom logic is contained in the new search type, leaving existing near searches (e.g., "pub in Kingston") unaffected.

Test Results:
make lint: passed
make mypy: passed (152 source files, no issues)
make bdd: 776 passed, 4 skipped
New BDD scenarios: qualifier search for unnamed pubs near Kingston, [amenity=restaurant] Vaduz API endpoint test
Tested on actual Nominatim installation using the --nominatim-purge flag

## AI usage

AI tool (GitHub Copilot) was used for planning approaches and suggesting possible testing scenarios.

All implementation, code modifications, and final validation were manually reviewed and tested on the actual Nominatim installation using the --nominatim-purge flag.

### DIFF
--- a/src/nominatim_api/search/db_search_builder.py
+++ b/src/nominatim_api/search/db_search_builder.py
@@ -103,6 +103,15 @@ class SearchBuilder:
                 builder = self.build_special_search(sdata, assignment.address,
                                                     bool(near_items))
         else:
+            if (sdata.qualifiers and not near_items and
+                    not self.details.categories):
+                near_builder = self._build_qualifier_address_search(
+                    sdata,
+                    assignment.name, assignment.address,
+                    assignment.penalty)
+                for search in near_builder:
+                    yield search
+
             builder = self.build_name_search(sdata, assignment.name, assignment.address,
                                              bool(near_items))
 
@@ -118,6 +127,39 @@ class SearchBuilder:
             for search in builder:
                 search.penalty += assignment.penalty
                 yield search
+
+    def _build_qualifier_address_search(self, sdata: dbf.SearchData,
+                                        name: qmod.TokenRange,
+                                        address: List[qmod.TokenRange],
+                                        base_penalty: float
+                                        ) -> Iterator[dbs.AbstractSearch]:
+        """ Build a QualifierNearSearch for queries like 'Kingston pub' where
+            a qualifier category is combined with an address but no explicit
+            name. Searches for the address as a place, then finds POIs of the
+            given category nearby.
+
+            Temporarily modifies sdata in place and restores it before
+            returning.
+        """
+        categories = sdata.qualifiers
+        penalty = min(categories.penalties)
+        categories.penalties = [p - penalty for p in categories.penalties]
+
+        sdata.qualifiers = dbf.WeightedCategories([], [])
+        saved_rankings = sdata.rankings
+        sdata.rankings = list(saved_rankings)
+
+        for search in self.build_name_search(sdata, name, address,
+                                             is_category=False):
+            if isinstance(search, dbs.PlaceSearch):
+                search_penalty = search.penalty
+                search.penalty = 0.0
+                yield dbs.QualifierNearSearch(
+                    penalty + base_penalty + search_penalty,
+                    categories, search)
+
+        sdata.qualifiers = categories
+        sdata.rankings = saved_rankings
 
     def build_poi_search(self, sdata: dbf.SearchData) -> Iterator[dbs.AbstractSearch]:
         """ Build abstract search query for a simple category search.
--- a/src/nominatim_api/search/db_searches/__init__.py
+++ b/src/nominatim_api/search/db_searches/__init__.py
@@ -15,3 +15,4 @@ from .country_search import CountrySearch as CountrySearch
 from .postcode_search import PostcodeSearch as PostcodeSearch
 from .place_search import PlaceSearch as PlaceSearch
 from .address_search import AddressSearch as AddressSearch
+from .qualifier_search import QualifierNearSearch as QualifierNearSearch
--- a/src/nominatim_api/search/db_searches/qualifier_search.py
+++ b/src/nominatim_api/search/db_searches/qualifier_search.py
@@ -1,0 +1,157 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+# This file is part of Nominatim. (https://nominatim.org)
+#
+# Copyright (C) 2026 by the Nominatim developer community.
+# For a full list of authors see the git log.
+"""
+Implementation of search for qualifier+address queries like 'Kingston pub'.
+
+Finds a place by name, then searches for nearby POIs of the given category.
+The inner place search is restricted to actual places (not POIs) and the
+near lookup only proceeds when there is exactly one matching place.
+"""
+from typing import List, Tuple
+
+import sqlalchemy as sa
+
+from . import base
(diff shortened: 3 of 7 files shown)


agam1092005:
@lonvia 
Kindly take a look

src/nominatim_api/search/db_searches/near_search.py
```diff
@@ -128,8 +129,17 @@ async def lookup_category(self, results: nres.SearchResults,
...
             result.accuracy = self.penalty + penalty
+            if result.names:
+                result.accuracy += 0.4
             result.bbox = Bbox.from_wkb(row.bbox)
-            results.append(result)
+            new_results.append(result)
+
+        if len(new_results) > 1:
+            for result in new_results:
+                result.accuracy += 0.2
```
lonvia: You can't add a general penalty hear because near search is also used for actual near searches ('pub in kingston') where multiple results are perfectly okay.

src/nominatim_api/search/db_search_builder.py
```diff
@@ -103,6 +103,21 @@ def build(self, assignment: TokenAssignment) -> Iterator[dbs.AbstractSearch]:
                 builder = self.build_special_search(sdata, assignment.address,
                                                     bool(near_items))
         else:
+            if (sdata.qualifiers and not near_items and
+                    len(sdata.qualifiers.values) == 1 and
```
lonvia: This condition is actually prevents 'kingston pub' from working when using the wiki special phrases set. The word pub is mapped to two qualifiers: `amenity=bar` and `amenity=pub`. That is not unusual and shouldn't really be a problem here.

src/nominatim_api/search/db_search_builder.py
```diff
@@ -103,6 +103,21 @@ def build(self, assignment: TokenAssignment) -> Iterator[dbs.AbstractSearch]:
...
+                    not self.details.categories):
+                import copy
```
lonvia: imports should always go to the top of the file.

src/nominatim_api/search/db_search_builder.py
```diff
@@ -103,6 +103,21 @@ def build(self, assignment: TokenAssignment) -> Iterator[dbs.AbstractSearch]:
...
+                sdata_near = copy.copy(sdata)
+                sdata_near.qualifiers = dbf.WeightedCategories([], [])
```
lonvia: You overwrite the qualifiers here only to hand them in again in the function below.

It's probably better to simply hand in the original sdata, make temporary modifications in place and then restore the original values before returning. Not ideal, I know, but at least it keeps the related code together.

src/nominatim_api/search/db_search_builder.py
```diff
@@ -103,6 +103,21 @@ def build(self, assignment: TokenAssignment) -> Iterator[dbs.AbstractSearch]:
...
+                sdata_near.rankings = list(sdata.rankings)
+
+                near_builder = self._build_qualifier_address_search(
+                    sdata_near, sdata.qualifiers,
+                    [assignment.name] + assignment.address,
```
lonvia: Hand those two parameters in separately instead of concatenating. They are just separated to their original state in the function.

src/nominatim_api/search/db_search_builder.py
```diff
@@ -119,6 +134,26 @@ def build(self, assignment: TokenAssignment) -> Iterator[dbs.AbstractSearch]:
...
+                                        base_penalty: float
+                                        ) -> Iterator[dbs.AbstractSearch]:
+        """ Build a NearSearch for queries like 'Kingston pub' where a qualifier
+            category is combined with an address but no explicit name.
+            Searches for the address as a place, then finds POIs of the given
+            category nearby.
+        """
+        penalty = min(categories.penalties)
+        categories.penalties = [p - penalty for p in categories.penalties]
+        for search in self.build_name_search(sdata, address[0], address[1:],
```
lonvia: You can't really use the name builder here because it creates address and place searches. We'd be only interested in village names, so place search es only.

And this brings up another problem: the place search would need to be restricted to [ranks](https://nominatim.org/release-docs/latest/customize/Ranking/#address-rank) that describe small villages/suburbs etc. PlaceSearch currently cannot do that. It only gets such restrictions from the details parameter that is injected internally.

agam1092005:
Thanks for the detailed review! I'm addressing all the feedback:

- Custom search type — Created a dedicated QualifierNearSearch class instead of modifying NearSearch, so penalties and restrictions are self-contained.
- 
- Removed len == 1 guard — "pub" mapping to both amenity=bar and amenity=pub now works.
- 
- Reverted near_search.py — No more shared penalties that would break regular near searches like "pub in Kingston".
- 
- Import cleanup — Removed import copy entirely (no longer needed).
- 
- In-place sdata modification — Passing original sdata into _build_qualifier_address_search, temporarily clearing qualifiers/rankings and restoring after.
- 
- Separate params — name and address passed individually, not concatenated.
- 
- PlaceSearch only — Filtering out AddressSearch results since we only want village/place names.

Regarding the rank restriction for small places: 
QualifierNearSearch currently handles this at lookup time by filtering base results to rank_address < 30 and bbox.area < 5.0. 

Happy to explore a different approach if you'd prefer the restriction to happen earlier in the query.
Pushing the updated commit shortly after testing.

mtmail:
Can you enable Github Actions for https://github.com/agam1092005/Nominatim/actions ? There's a section in https://github.com/osm-search/Nominatim/blob/master/CONTRIBUTING.md#workflow-for-pull-requests explaining how to do that. It will run the full test suite automatically for you.

PR #4048, opened 2026-03-23
Contributing: clarify Github Actions/CI

mtmail:
Adds a helpful link how to enable Github Actions on a repository. I assume some readers might not be familiar with "CI". Github changed their URLs to be extra long.

- [X] I have adhered to the coding style - The file is excluded from the style, too bad about the long URLs
- [X] I didn't use LLM/AI


lonvia:
I fear the overall problem is more one of RTM but the additional link is definitely helpful. Thanks.

Issue #4022, opened 2026-03-08
SQLAlchemy schema needs to be updated with new constraints


lonvia:
#3939 introduced quite a few new NOT NULL constraints and then forgot to update the SQL schema for the API in `src/nominatim_api/sql/sqlalchemy_schema.py`. This file should be reviewed and updated to have the same definitions as `lib-sql`. Please be careful when doing this, some of the tables have a couple of columns are intentionally missing in this file. Thus the update should not add any new columns.

_Note: do not try to let AI solve this for you. It will utterly fail as four PRs have already proven.._

Arkadeep455:
Hi @lonvia, hope you are doing well. I have started working on this issue.

I am iterating over the issue just to make sure that I have understood the problem statement clearly and please correct me if I am wrong. So, the pull request #3939 updated the SQL database definitions to make certain columns mandatory. NOT NULL constraints were added so the database basically rejects any empty field. Now, the python code thinks those fields are optional but the database needs those fields. So, there is a mismatch so what I am suppose to do is to fix the code by placing 'nullable=False' to specific definitions actually. I cannot think there is any other way of doing this as of now so I will follow this approach.

Rampage270906:
Hey if this hasn't been solved yet,  I would like to work on this issue.

Arkadeep455:
Hi @Rampage270906 , I am working on this, solved it, going to make the pull request today.

kad-link:
if no one's working on it, I would like to go through this

Rampage270906:
Hey @kad-link, I am working on this and will hopefully submit pr by tomorrow evening

PR #4049, opened 2026-03-24
Add uv support for development environment

Itz-Agasta:
## Summary
- Add a root `pyproject.toml` with uv workspace configuration and dev dependency groups, enabling contributors to use uv as an alternative to the existing virtualenv + pip workflow.
- Add documentation for using uv in the dev environment setup.

### How it works
Contributors using uv can set up their env with a single command:
```sh
uv sync
# This installs runtime dependencies (from the package pyproject.toml files)
# and all dev tools (pytest, mypy, flake8, mkdocs, etc.) into a local .venv.

# Individual groups can be installed separately:
uv sync --group test --no-dev    # Only test dependencies
uv sync --group docs --no-dev    # Only docs dependencies
uv sync --group serve --no-dev   # Only server dependencies
```

**NOTE:** uv.lock is in .gitignore as per project policy.... each developer resolves dependencies locally. For more info refer to the thread https://github.com/osm-search/Nominatim/discussions/4044

## AI usage
None

### DIFF
--- a/docs/develop/Development-Environment.md
+++ b/docs/develop/Development-Environment.md
@@ -68,14 +68,14 @@ Create a virtual environment and install all dependencies:
 ```sh
 virtualenv ~/nominatim-dev-venv
 . ~/nominatim-dev-venv/bin/activate
-pip install --group runtime --group dev .
+pip install --group runtime --group dev
 ```
 
 To install dependencies individually:
 
 ```sh
 pip install \
-    psutil 'psycopg[binary]' PyICU SQLAlchemy \
+    psutil 'psycopg[binary]' PyICU 'SQLAlchemy[asyncio]' \
     python-dotenv jinja2 pyYAML \
     mkdocs 'mkdocstrings[python]' mkdocs-gen-files mkdocs-material \
     pytest pytest-asyncio pytest-bdd flake8 \
--- a/docs/develop/Development-Environment.md
+++ b/docs/develop/Development-Environment.md
@@ -63,20 +63,24 @@ sudo apt install build-essential libsqlite3-mod-spatialite osm2pgsql \
 
 #### Using pip
 
-Create a virtual environment and install all dependencies from the project's
-`pyproject.toml`:
+Create a virtual environment and install all dependencies:
+
 ```sh
 virtualenv ~/nominatim-dev-venv
 . ~/nominatim-dev-venv/bin/activate
-pip install --group dev .
+pip install --group runtime --group dev .
 ```
 
-This installs the two Nominatim packages (`nominatim-api`, `nominatim-db`)
-along with all development dependencies (tests, linting, type stubs, docs, serve).
+To install dependencies individually:
 
-To install runtime dependencies only:
 ```sh
-pip install .
+pip install \
+    packaging/nominatim-db packaging/nominatim-api \
+    pytest pytest-asyncio pytest-bdd flake8 mypy \
+    mkdocs mkdocstrings mkdocs-gen-files mkdocs-material \
+    osmium aiosqlite falcon starlette uvicorn \
+    types-jinja2 types-markupsafe types-psutil types-psycopg2 \
+    types-pygments types-pyyaml types-requests types-ujson types-urllib3 typing-extensions
 ```
 
 #### Using uv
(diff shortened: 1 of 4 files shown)


Itz-Agasta:
## Demo
https://github.com/user-attachments/assets/4c2f0cf4-ff93-4078-8ca6-4d45a2bab886

This is a testing rollout; the CI hasn’t been touched yet.

docs/develop/Development-Environment.md
```diff
@@ -82,6 +82,27 @@ Now enter the virtual environment whenever you want to develop:
 . ~/nominatim-dev-venv/bin/activate
 ```
 
+### Alternative: Using uv
```
lonvia: I think this still needs the basic packages from the system as installed by `apt install` above. To make that clear, I'd suggest to make this a fourth level heading (####) and add another fourth level subheading in line 63. Something like "Using virtualenv" maybe.

docs/develop/Development-Environment.md
lonvia, line 72: The dependency groups work nicely with pip as well:

```
~/nominatim-dev-venv/bin/pip install --group dev
```

Sadly this isn't a complete replacement for the command above because the required dependencies that are defined in the packages are missing. If there isn't a way to pull them from the subpackages automatically, I'm leaning towards adding another dependency group with those requirements. It'd still be a duplication but in contrast to the line above in the documentation it could be tested in the CI to ensure it stays up-to-date.

pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+docs = [
+    "mkdocs",
+    "mkdocstrings[python]",
+    "mkdocs-gen-files",
+    "mkdocs-material",
+]
+
+serve = [
+    "falcon",
+    "gunicorn",
```
lonvia: I'd remove gunicorn here. It's for deployment only.

lonvia (commented):
Finally got around trying this out. It's a very nice solution to #4022, even for the pip users, and uv opens up testing against other python versions.

Having the dependencies in the root pyproject.toml also makes much more sense because development only happens on the full package. The only optional dependencies that are not always development-related are around serving. But even there it makes sense to have different extra groups (see comment on gunicorn). Something to think about separately.

I've added small comments. Let me know what you think about them.

docs/develop/Development-Environment.md
Itz-Agasta: > I'm leaning towards adding another dependency group with those requirements. 

I will argue with that, i dont think it will a good idea to manage dependencies form two files lt might raise ambiguity later.

I would say  lets make  the root `pyproject.toml` a proper package (nominatim) that declares `nominatim-api` and `nominatim-db`  as its dependencies. With this in place,
```python
 pip install --group dev .
```
It will does everything in one-shot, Installs `nominatim-api` and `nominatim-db` (and their deps) via `[project].dependencies` and installs all dev tooling via `--group dev`

Itz-Agasta:
I had to pin mypy to <=1.19.1.... As v 1.20 giving false positives locally

image

> Not related to uv. i have tested with pip too

Mypy 1.20 changed how it handles attribute access on class objects vs instances also some breaking changes.

Refs:
- https://github.com/python/mypy/issues/21137
- https://github.com/python/mypy/issues/20726

Idk what y recommend for this ?

lonvia:
Those mypy issues are fixed on master. Simply rebase.

Itz-Agasta:
> Those mypy issues are fixed on master. Simply rebase.

ohh my bad :( ... done

docs/develop/Development-Environment.md
lonvia: We unfortunately can't have a package "nominatim". The name is [already used](https://pypi.org/project/nominatim) and I won't to avoid confusion at any cost. I've [contacted the owner](https://github.com/damianbraun/nominatim/issues/10) but with no success so far.

The other thing to note is that for my use-case I don't want nominatim-{api,db} themselves installed in the virtual environment. I just want the dependencies, i.e. a virtualenv with which I can then run the `nominatim-cli.py` script in place.

Oh well, this is just a nice-to-have. Lets defer it to later.

otbutz: It looks like the owner abandoned his Github account. His personal email address can still be retrieved via `git log` if you want to try that.

lonvia:
Can we go back to the version without a project in the base pyproject.toml?

Itz-Agasta:
Pls have a look now... I’ve added the runtime group as y mentioned... Also i noticed that you’ve added the async version of SQLAlchemy in the API deps, so updated the dev docs accordingly.

> and sorry for the communication delay.. my end sem exams are on-going :(

lonvia:
Thanks and sorry for the long wait.

Issue #3210, opened 2023-09-25
City doesn't come in the result


tiagomn:
Hi Guys. There was a fix on issue  #3196  but the Capital of Brazil which is "Brasilia, DF, Brazil" is not showing neither and I think it's the same issue. What is interesting, if we search "Bra", the result comes. Would you mind to take a look at this issue, please?

mtmail:
Would you mind using the issue template?

lonvia:
Works fine here: https://nominatim.openstreetmap.org/ui/search.html?q=Brasilia%2C+DF%2C+Brazil

tiagomn:
If you type only Brasilia, it doesn't work., as People type only the name of the city and not the complete one

tiagomn:
Hi guys. Did you take a chance to check this issue? I suspect it happens to other cities too as we have been facing this kind of issue about 2 months.

Thanks in Advance.

lonvia:
This needs language-dependent penalties to be fixed. 'Brasilia' is a valid name for the country of Brasil in quite a few languages.

divyanshuagarwal-23:
Hello, I want to pick this up

lonvia:
Another example: [Turgi](https://nominatim.openstreetmap.org/ui/details.html?osmtype=N&osmid=102305481&class=place) always finds Turkey. See https://community.openstreetmap.org/t/the-case-of-5300-turgi-for-search-engine/133886 .

Itz-Agasta:
> This needs language-dependent penalties to be fixed.

Hi @lonvia i was thinking something diff.....What if we dampen importance for countries (search_rank == 4) when they're matched via fuzzy/partial tokens? 

what i undertnd is when importance is calculated and later combined with penalty in the outer query, countries (search_rank=4) with high importance (~0.97) overpower the edit-distance penalties (~0.3) from match_penalty() in icu_tokenizer.py.... like for "Brasilia"--> 
Brazil: importance=0.97, penalty≈0.3 = final score ~ - 0.67 and 
Brasília: importance=0.65, penalty≈0.0 =  final score ~ - 0.65

The country wins despite being a partial match because -0.67 <  -0.65.... so if we dampen country importance by ~30% i think it may work also it will Only affects countries on fuzzy matches

lonvia:
[...]

PR #4088, opened 2026-05-03
Use locale-aware names for search result reranking

### LINKED ISSUE #3871: Athens: unexpected search result order
## What did you search for?

[Athens](https://nominatim.openstreetmap.org/ui/search.html?q=Athens)

## What result did you get?

Athens, Clarke County, Georgia, United States Administrative

Athens, Municipality of Athens, Regional Unit of Central Athens, Attica, 105 57, Greece City

## Further details

The capital of Greece is only the 7th result, after the following:

1. Athens, Clarke County, Georgia, United States Administrative
2. Athens County, Ohio, United States Administrative
3. Athens, Athens Township, Athens County, Ohio, 45701, United States Administrative
4. Athens, Henderson County, Texas, 75751, United States Administrative
5. Athens, Limestone County, Alabama, 35613, United States Administrative
6. West Athens, Los Angeles County, California, United States Administrative

Given that Athens, Greece is a capital, the first place named "Athens" and that it probably has a larger population than all the above results combined, I would expect it to be first.

Itz-Agasta:
## Summary

Searching for "Athens" with `accept-language: en` returned **Athens, Georgia, US** before **Athens, Greece** because the reranking logic used `display_name` for word matching, which returned the local Greek name "Αθήνα". This caused a distance penalty for Athens, Greece since "Athens" had no match in the word set.

 Introduce `_get_result_rerank_text()` which builds the reranking text using locale-aware name resolution via `locales.display_name()`. This ensures that translated name variants (e.g., `name:en`) are included in the word matching pool when the caller specifies a language preference.

closes  #3871 #4062 

## AI usage
None

### DIFF
--- a/src/nominatim_api/search/geocoder.py
+++ b/src/nominatim_api/search/geocoder.py
@@ -175,6 +175,28 @@ class ForwardGeocoder:
 
         return final
 
+    def _get_result_rerank_text(self, result: SearchResult) -> str:
+        if not self.params.locales:
+            return result.display_name or ''
+
+        label_parts: List[str] = []
+        if result.address_rows:
+            for line in result.address_rows:
+                if line.isaddress and line.names:
+                    address_name = self.params.locales.display_name(line.names)
+                    if address_name and (
+                        not label_parts or label_parts[-1] != address_name
+                    ):
+                        label_parts.append(address_name)
+
+        if label_parts:
+            return ', '.join(label_parts)
+
+        if result.names:
+            return self.params.locales.display_name(result.names)
+
+        return result.display_name or ''
+
     def rerank_by_query(self, query: QueryStruct, results: SearchResults) -> None:
         """ Adjust the accuracy of the localized result according to how well
             they match the original query.
@@ -192,8 +214,12 @@ class ForwardGeocoder:
                or (result.importance is not None and result.importance < 0):
                 continue
             distance = 0.0
-            norm = self.query_analyzer.normalize_text(' '.join((result.display_name,
-                                                                result.country_code or '')))
+            # Use locale-aware text for word matching so that translated names
+            # (e.g., name:en) are included in the match pool.
+            rerank_text = self._get_result_rerank_text(result)
+            norm = self.query_analyzer.normalize_text(
+                ' '.join((rerank_text, result.country_code or ''))
+            )
             words = set((w for w in re.split('[-,: ]+', norm) if w))
             if not words:
                 continue
--- a/src/nominatim_api/search/geocoder.py
+++ b/src/nominatim_api/search/geocoder.py
@@ -7,7 +7,7 @@
 """
 Public interface to the search code.
 """
-from typing import List, Any, Optional, Iterator, Tuple, Dict, Set
+from typing import List, Any, Optional, Iterator, Tuple, Dict
 import itertools
 import re
 import difflib
@@ -175,11 +175,11 @@ class ForwardGeocoder:
 
         return final
 
-    def _get_result_rerank_text(self, result: SearchResult) -> Set[str]:
+    def _get_result_rerank_text(self, result: SearchResult) -> set[str]:
         if not self.params.locales:
             return {result.display_name} if result.display_name else set()
 
-        label_parts: Set[str] = set()
+        label_parts: set[str] = set()
         if result.address_rows:
             for line in result.address_rows:
                 if line.isaddress and line.names:
--- a/src/nominatim_api/search/geocoder.py
+++ b/src/nominatim_api/search/geocoder.py
@@ -7,7 +7,7 @@
 """
 Public interface to the search code.
 """
-from typing import List, Any, Optional, Iterator, Tuple, Dict
+from typing import List, Any, Optional, Iterator, Tuple, Dict, Set
 import itertools
 import re
 import difflib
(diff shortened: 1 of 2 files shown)


Itz-Agasta:
## Results
toto-result

src/nominatim_api/search/geocoder.py
lonvia: This should be able to use the computed 'norm' now from above, shouldn't it?

src/nominatim_api/search/geocoder.py
```diff
@@ -175,6 +175,28 @@ def sort_and_cut_results(self, results: SearchResults) -> SearchResults:
 
         return final
 
+    def _get_result_rerank_text(self, result: SearchResult) -> str:
+        if not self.params.locales:
+            return result.display_name or ''
+
+        label_parts: List[str] = []
```
lonvia: Given that we only match against the rerank text in the end, order doesn't really matter. That means you can use a set here and simplify the code in lines 187-190.

You might even get more simple code if you then just return the set. Words will be joined in line 221 anyways.

test/bdd/features/db/query/search_simple.feature
```diff
@@ -100,3 +100,17 @@ Feature: Searching of simple objects
...
          | N10    |
+
+    # github #3871
+    Scenario: Localized result name is used for reranking
+        Given the places
+         | osm | class | type | name+name | name+name:en | geometry   |
+         | N1  | place | city | Αθήνα     | Athens       | country:gr |
+         | N2  | place | town | Athens    |              | country:us |
+        When importing
+        And geocoding "Athens"
```
lonvia, line 111: Could you repeat line 111-166 with an `accept-language=de` to prove that this will indeed still return N2?

lonvia (commented):
Looks good in general, just some minor suggestions for code improvements.

lonvia:
Please rebase this on master. #4090 contained a few new fixes for mypy.

src/nominatim_api/search/geocoder.py
Itz-Agasta: Mmm we can ig.... but doesn’t the `norm/words` set include both the localized names + the country codes too...Won’t that cause problems later?

When I designed the country penalty logic, the idea was basically “does the query match the localized country name?”. But if we use `words`, then I’d effectively be checking “does the query match anything in the rerank token pool, including country codes?” .... wouldn’t that introduce some weird edge cases later?

Itz-Agasta: ok ok correct me if im wrong... lets say a suer search for `brasilia BR` (query )...so our tokenizer matches it to the country brazil and  looks at `norm/words` (which includes {"brasil", "br"}). Since the query contains "BR", the words are not disjoint....... so it wont give the county penalty ( #3210 bug)

Itz-Agasta: @lonvia current version striping country code manually... will change it if needed...i feel it can be done without striping also idk ...suggest me what y think about this :(

lonvia: Hmm, by that argumentation the presence of an country code in addition to other terms should even add a penalty. That would mean that our country check can be much stricter. Instead of checking presence of any words from the name, the query as a whole should be checked against the whole of the name.

src/nominatim_api/search/geocoder.py
```diff
@@ -7,7 +7,7 @@
 """
 Public interface to the search code.
 """
-from typing import List, Any, Optional, Iterator, Tuple, Dict
+from typing import List, Any, Optional, Iterator, Tuple, Dict, Set
```
lonvia: Minor general styling comment here: since #3785 we can use [type hinting generics in standard collections](https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections). I've no intention to reformat the entire project but new code should use the standard types for type hinting instead of the imports from `typing`.

src/nominatim_api/search/geocoder.py
Itz-Agasta: sorry.. I'm not sure I fully understood your suggestion. Do you want me to change the country penalty to something like  checking if the full query text is a substring of the full localized name... or something like that ?

lonvia: Not substring, checking that they are equal.

Your observation about `brasilia, BR` is right but it equally applies to `Brasilia, Brasil`. So basically that means the country penalty should apply for any query that is not exactly a country name.

Itz-Agasta: ah I see, you want the normalized query string to be checked for equality against the normalized country name. So if the query is "Brasilia" and the localized country name is "Brasil", they're not equal ->  penalty applies. That makes sense.

 Also I see your PR #4095 already solved the "BR in query -> country wins" issue at the SQL level. After I rebase, is that approach in #4095 sufficient ?

lonvia: #4095 is just an approach on a different level and if won't solve the problem in all cases. Being more tight about the penalty check still makes sense on top of that.

Itz-Agasta:
Ig its good to go... pls have a look now

src/nominatim_api/search/geocoder.py
```diff
@@ -207,12 +231,11 @@ def rerank_by_query(self, query: QueryStruct, results: SearchResults) -> None:
             # to offset this.
             if result.rank_address == 4:
                 if self.params.locales and result.names:
-                    loc_names = [result.names[t] for t in self.params.locales.name_tags
-                                 if t in result.names]
-                    if loc_names:
-                        norm_loc = self.query_analyzer.normalize_text(' '.join(loc_names))
-                        loc_words = set(w for w in re.split('[-,: ]+', norm_loc) if w)
-                        if loc_words and loc_words.isdisjoint(qwords):
+                    locale_name = self.params.locales.display_name(result.names)
```
lonvia: Oh, I meant match the query fully against any of the names we had previously and probably against country code as well. Essentially, with my current setting of 'en,de' language searching for 'Germany', 'Deutschland' and 'de' should favour the country result. Anything else shouldn't.

Itz-Agasta: > probably against country code as well 

mm what’s the  reason for including the country code in the exact-match check too?  do poeple query by codes too ? lets say somone type "br" isnt it falls under typo?

lonvia: Searching a country by two-letter code is not uncommon. People usually know the code at least for their own country and writing "us" beats writing "united states of america".

lonvia:
All good. Thank you!

PR #4092, opened 2026-05-10
Allow importing areas for external postcode data

### LINKED ISSUE #4080: Allow importing areas for external postcodes
With the reorganised postcode handling from #3926 it is now possible to have area geometries for external postcodes. It would be nice to offer to import either custom extents or full polygons.

Currently the external postcodes are in CSV format. Adding support for an additional extent should be trivial. However, I'm not a big fan of adding geometry columns in there, especially polygons that can become quite large WKTs. Therefore I'm leaning towards adding support for a json line format with one geojson object per line.

Side issue: external postcodes are always fallbacks at the moment. Once we allow to rely on external postcode areas, it would be good to have an option to make them the authoritative source and ignore postcodes from OSM.

AyushDharDubey:
## Summary
Closes #4080
This PR introduces support for importing external postcode data in both csv and jsonl (json lines) formats, allowing for richer postcode geometry import. In JSONL files, we can now provide full geometries, and centroids (optional). 

The new JSONL files would need to placed in the project directory with name `XX_postcodes_geometry.jsonl`.

Currently OSM imported postcodes precede over jsonl imports which precedes csv imports.

The import logic, storage, and documentation are updated accordingly, and comprehensive tests are added for the new functionality.

### AI Usage
LLM based CLI tools were extensively used to understand the codebase and find out the relevant files pertaining the targeted changes.  

### Contributor guidelines (mandatory)

### DIFF
--- a/docs/customize/Postcodes.md
+++ b/docs/customize/Postcodes.md
@@ -1,31 +1,51 @@
 # External postcode data
 
-Nominatim creates a table of known postcode centroids during import. This table
-is used for searches of postcodes and for adding postcodes to places where the
-OSM data does not provide one. These postcode centroids are mainly computed
-from the OSM data itself. In addition, Nominatim supports reading postcode
-information from an external CSV file, to supplement the postcodes that are
-missing in OSM.
+Nominatim creates a table of known postcode centroids and geometries during import.
+This table is used for searches of postcodes and for adding postcodes to places where
+the OSM data does not provide one. These postcode centroids are mainly computed
+from the OSM data itself. In addition, Nominatim supports reading postcode information
+from external files to supplement the postcodes that are missing in OSM.
 
-To enable external postcode support, simply put one CSV file per country into
-your project directory and name it `<CC>_postcodes.csv`. `<CC>` must be the
-two-letter country code for which to apply the file. The file may also be
-gzipped. Then it must be called `<CC>_postcodes.csv.gz`.
+## Supported file formats
 
-The CSV file must use commas as a delimiter and have a header line. Nominatim
-expects three columns to be present: `postcode`, `lat` and `lon`. All other
-columns are ignored. `lon` and `lat` must describe the x and y coordinates of the
-postcode centroids in WGS84.
+Nominatim supports 2 data formats for external import:
 
-The postcode files are loaded only when there is data for the given country
-in your database. For example, if there is a `us_postcodes.csv` file in your
-project directory but you import only an excerpt of Italy, then the US postcodes
-will simply be ignored.
+### JSONL format
+
+The JSONL format may contain the postcode data along with a POLYGON or MULTIPOLYGON geometry
+of the postcode area. To enable external postcode support, put one file per country into
+your project directory and name it `<CC>_postcodes_geometry.<ext>`. `<CC>` must be the
+two-letter country code for which to apply the file. File type may be either `.jsonl` or
+`.jsonl.gz` (gzip compressed). The file must be in UTF-8 encoding and contain one JSON object
+per line with the following structure:
+
+#### GeoJSON Feature
+
+A standard [RFC7946](https://geojson.org) GeoJSON Feature object .
+
+* `geometry`: A GeoJSON geometry (Polygon/MultiPolygon).
+* `properties`: An object containing:
+  * `postcode`: (required) The postcode string.
+  * `lat`, `lon`: (optional) Coordinates for the centroid. If not provided, the centroid
+  is computed from the geometry. Thus centroid coordinates and geometry independently optional,
+  however at least one of them is required.
+
+### CSV format
+
+To enable external postcode support, put one file per country into
+your project directory and name it `<CC>_postcodes.<ext>`. `<CC>` must be the
+two-letter country code for which to apply the file. File type may be either `.csv` or `.csv.gz`
+(gzip compressed). The CSV file must use commas as a delimiter and have a header line. Nominatim
+expects three columns to be present: `postcode`, `lat` and `lon`. All other columns are ignored.
+`lon` and `lat` must describe the x and y coordinates of the postcode centroids in WGS84.
+
+The postcode area is assumed to be in a buffer around the centroid (typically 5km).
+
+## Usage
 
 As a rule, the external postcode data should be put into the project directory
 **before** starting the initial import. Still, you can add, remove and update the
-external postcode data at any time. Simply
-run:
+external postcode data at any time. Simply run:
 
 ```
 nominatim refresh --postcodes
--- a/src/nominatim_db/tools/postcodes.py
+++ b/src/nominatim_db/tools/postcodes.py
@@ -14,6 +14,7 @@ from pathlib import Path
 import csv
 import gzip
 import logging
+import json
 from math import isfinite
 
 from psycopg import sql as pysql
@@ -59,9 +60,12 @@ class _PostcodeCollector:
         self.extent = default_extent
         self.exclude = exclude
         self.collected: Dict[str, PointsCentroid] = defaultdict(PointsCentroid)
+        self.geometries: Dict[str, str] = {}
+        self.is_authoritative: bool = False
         self.normalization_cache: Optional[Tuple[str, Optional[str]]] = None
 
-    def add(self, postcode: str, x: float, y: float) -> None:
+    def add(self, postcode: str, x: Optional[float], y: Optional[float],
+            extent: Optional[int] = None, geometry: Optional[str] = None) -> None:
         """ Add the given postcode to the collection cache. If the postcode
             already existed, it is overwritten with the new centroid.
         """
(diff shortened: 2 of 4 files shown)


AyushDharDubey:
I have conserved the support for the existing csv import workflow alongside the new jsonl geometry support because for most countries (including my own🫠️), the centroid-tagged postcode dataset is rather dominant in open space than a maturely marked boundary dataset.

docs/customize/Postcodes.md
```diff
@@ -1,31 +1,56 @@
...
+#### GeoJSON Feature
+
+A standard [RFC7946](https://geojson.org) GeoJSON Feature object .
+
+* `geometry`: A GeoJSON geometry (Point, Polygon, etc.).
+* `properties`: An object containing:
+  * `postcode`: (required) The postcode string.
+  * `lat`, `lon`: (optional) Coordinates for the centroid. If not provided, the centroid 
+  is computed from the geometry. Thus centroid coordinates and geometry are mutually optional, but exclusive
+  * `extent`: (optional) An approximate geographic extent of the postcode in meters.
```
lonvia: Extent is mutual to supplying a full geometry.

lonvia:
I've skimmed through the PR and realise that we need to discuss a bit more how exactly the areas play with the other postcode sources.

Right now the order is as follows: first postcodes areas from OSM are imported. They are considered authoritative. That means postcode points from OSM with the same postcode are ignored and so are any postcode point inside such an area. Then the remaining postcode points are used to compute centroids. The area for such a centroid is a rectangle according to extent. Finally, postcodes are added from the external source but only if they don't appear at all in the OSM data.

If we now allows postcode areas in the external data, then it would make sense to assume that their area is authoritative as well. So the order would be something along those lines:

1. Import postcode areas from OSM.
2. Add postcode areas from external data for postcodes not found as areas in OSM.
3. Add postcode centroids from OSM and external data as before but also using external postcode areas as exclusion zones/postcodes.

The other thing I noticed is that in your reading of GeoJSON, the geometry can be optional. That doesn't quite match with its specification. A geometry must always be present. So the rules for interpretation should go somewhere along this:

* if geometry is POINT, interpret as postcode centroid. Also parse extent if present.
* If geometry is POLYGON or MULTIPOLYGON, use as exact postcode area. Also parse lat/lon if present and use as centroid. Ignore extent.
* If geometry is any other geometry type, ignore as invalid.

AyushDharDubey:
Thanks for the context. It will overhaul this PR completely.

Now, please correct me if I’m deviating here:

Instead of allowing `POINT` geometry in JSONL, we could rather restrict JSONL files to `POLYGON` and `MULTIPOLYGON` geometries only, and treat this file as authoritative, on par with OSM postcode area data (only if the postcode area is not already present from OSM).

If a user has a repository of postcode centroids, they can always import it through the existing CSV import-flow.

the CSV files could continue to be handled by `_update_guessed_postcode()` & JSONL files would be handled by a new method executed after `_update_postcode_areas()` (OSM postcode import). ?

lonvia:
> Instead of allowing POINT geometry in JSONL, we could rather restrict JSONL files to POLYGON and MULTIPOLYGON geometries only, and treat this file as authoritative, on par with OSM postcode area data (only if the postcode area is not already present from OSM).

Hmm, yes, I see how this would make things less confusing and the code a lot easier. I'd then give the file a different name though, not just a different extension. `XX_postcode_areas.jsonl`, for example.

Question still remains how they interact with the areas from OSM itself. Possible solutions that come to mind:

1. Only import when no area with the same postcode exists.
2. Only import the parts where they do not overlap with postcode areas from OSM. (Potentially making the area smaller to avoid overlaps.)
3. Make the external postcodes completely authoritative and ignore all postcodes from OSM. (i.e. skipping over the entire postcode generation process for the given country and just import the external ones)

The third point is something that will be requested as a feature sooner or later. So it remains to decide, if there is a use-case for 1) or 2). @mtmail, any opinion?

lonvia:
I'd say lets go with option 1) then for now. We can always add a command-line switch for 3) later.

src/nominatim_db/tools/postcodes.py
```diff
@@ -291,12 +382,46 @@ def _update_postcode_areas(conn: Connection, analyzer: AbstractAnalyzer,
...
+        areas.
+    """
+
+    if project_dir is None:
+        return
+
+    todo_countries: set[str] = set()
+
+    with conn.cursor() as cur:
+        cur.execute("""SELECT DISTINCT country_code FROM placex
```
lonvia, line 398: Getting them from placex is really expensive. Use the `country_name` table instead which has the full list of countries that Nominatim can handle.

src/nominatim_db/tools/postcodes.py
```diff
@@ -132,50 +162,107 @@ def commit(self, conn: Connection, analyzer: AbstractAnalyzer,
...
+        csvfile = self._open_external_centroid_file(project_dir)
+        if csvfile is not None:
             reader = csv.DictReader(csvfile)
             for row in reader:
                 if 'postcode' not in row or 'lat' not in row or 'lon' not in row:
                     LOG.warning("Bad format for external postcode file for country '%s'."
                                 " Ignored.", self.country)
                     return
                 postcode = analyzer.normalize_postcode(row['postcode'])
+                try:
```
lonvia, line 252: Any particular reason why you moved this block out of the `if` of line 261?

src/nominatim_db/tools/postcodes.py
```diff
@@ -132,50 +162,107 @@ def commit(self, conn: Connection, analyzer: AbstractAnalyzer,
...
+            fname = project_dir / f'{self.country}_postcodes.{ext}'
+
+            if fname.is_file():
+                LOG.info("Using external postcode file '%s'.", fname)
+                if ext.endswith('.gz'):
+                    return gzip.open(fname, 'rt', encoding='utf-8')
+                else:
+                    return open(fname, 'r', encoding='utf-8')
 
-        try:
```
lonvia, line 143: You seem to have removed the try/finally block. This is necessary when working with file descriptors without a `with ` context. There might be other reasons something throws in the code except the parts that are explicitly caught right now.

Also needed above in `_update_from_external_geometry()`.

src/nominatim_db/tools/postcodes.py
```diff
@@ -291,12 +382,46 @@ def _update_postcode_areas(conn: Connection, analyzer: AbstractAnalyzer,
...
+                       WHERE osm_id is not null
+                       ORDER BY country_code""")
+        for cc, pc in cur:
+            area_pcs[cc].add(pc)
+
+    for country in todo_countries:
+        _PostcodeCollector(country, matcher.get_matcher(country),
+                           matcher.get_postcode_extent(country),
+                           exclude=area_pcs[country]).commit(conn, analyzer, project_dir,
+                                                             is_initial, True)
```
lonvia, line 416: I don't think it is worth reusing the PostcodeCollector class here. It makes the code of the class a lot more complicated and you don't need the core functionality of this class at all (collecting lots of points into a single centroid). Can you give it a try and see if something more similar to `_update_postcode_areas()` possibly with slightly changing `_insert_postcode_areas()` to make it work for that use case would simplify the code?

src/nominatim_db/tools/postcodes.py
```diff
@@ -308,22 +433,30 @@ def _update_guessed_postcode(conn: Connection, analyzer: AbstractAnalyzer,
...
 
-    # Next, get the list of postcodes that are already covered by areas.
-    area_pcs = defaultdict(set)
+    # Next, get the list of authoritative postcodes comprising of postcodes that are already
+    # covered by areas and JSONL postcode import.
+    auth_pcs = defaultdict(set)
     with conn.cursor() as cur:
         cur.execute("""SELECT country_code, postcode
                        FROM location_postcodes WHERE osm_id is not null
+                            OR is_authoritative
```
lonvia, line 442: In #4067 I went for looking if the geometry is a rectangle to determine if it is made from an area of not. I see how this is a bit hacky, so I don't mind adding an additional column instead. However, I would rather go for a more explicit `is_area` column which is also set to true for the postcode areas from OSM. Area or not is the information that we are after here, after all. Two things to keep in mind with the new column: A new column needs a migration (see https://github.com/osm-search/Nominatim/blob/master/src/nominatim_db/tools/migration.py) and you should adapt the code from #4067 to make proper use of the new column.

src/nominatim_db/tools/postcodes.py
AyushDharDubey: I believed self.add is the helper function designed for this task: to handle the normalization part and add the coordinate/geometry in self.collected. So it would be better to reuse the add method.

Would this pose a larger issue?

src/nominatim_db/tools/postcodes.py
AyushDharDubey: I did consider the possibility of using the already persistent `is_area` column, but the issue was that (as discussed earlier) we intend to prioritize OSM areas over imported postcode areas. If we group both under a single boolean flag (`is_area`), we can't discriminate between the two (in order to prioritize during the import).

checking rectangle is also good, but as you did mention, not very clean...

Having a separate flag for imported geometries, would let us have semantic distinction between the two.

lonvia: You can still make the distinction looking at the osm_id column where necessary. It is only set for areas coming from OSM.

src/nominatim_db/tools/postcodes.py
lonvia: Is it possible we are talking about different things? Reusing add() is fine.

What I wanted to say: if the postcode is not in collected, then self.add() will not be called, so no need to parse `row['lon']` and `row['lat']`.

AyushDharDubey: Ah yes, that was my mistake, the parsing should be inside the if block as before.

AyushDharDubey:
Thanks for the reviews.
I will need to close this one and open a revised PR, rebased over master (this branch is already 52 commits behind🥲️) to accomodate #4115

PR #4101, opened 2026-05-21
Add stable postcode refs for details lookup and search exclusion

### LINKED ISSUE #4079: Details offers no stable link to artificial postcodes
## What did you search for?

https://nominatim.openstreetmap.org/ui/search.html?q=94110

## What result did you get?

Several postal code and postcode results (as expected). When I click on the San Francisco, CA postcode result, I see a details link. The link is to https://nominatim.openstreetmap.org/ui/details.html?place_id=352986712 which is not 94110 in San Francisco but 22214 in Tijuana.

This result is not the same every time. Sometimes I get a link to the SF postcode. Sometimes I get a link to another incorrect location.

Searching again gives me the same result. Tossing cookies and clearing history does not affect the result. I think the result changes after minutes or hours have passed and then is stable (whether correct or incorrect) for minutes or hours.

The issue is not specific to 94110. The detail links for other US postcodes (for example 94112, 94114, 94134, 18370) also sometimes link to different postcodes in other countries.

## Side note

The display of 94110 in SF on https://nominatim.openstreetmap.org/ui/search.html?q=94110 shows it in the middle of San Francisco Bay. This is incorrect. Yesterday I corrected an address, that incorrectly had the postcode 94110, that might have been responsible for the current incorrect location of 94110. I'm waiting for the correction to propagate through the systems involved. I don't mean to pursue that issue here; I just mention it in case anyone wonders.

Itz-Agasta:
## Summary
This change adds first-class support for stable postcode references in the API.
Artificial postcode rows do not always have a stable OSM object reference, so this introduces a dedicated ref format:

`P<country_code>:<postcode>`

eg:
- `Pus:94110`
- `Pgb:EH4 7EA`

closes #4079 

## AI usage
The test files  rework is done by ai according to my changes


docs/api/Details.md
```diff
@@ -39,6 +39,15 @@ for a place is different between Nominatim installation (servers) and
 changes when data gets reimported. Therefore it cannot be used as
 a permanent id and shouldn't be used in bug reports.
 
+``` xml
+https://nominatim.openstreetmap.org/details?place_ref=P<country_code>:<postcode>
```
lonvia: If this introduces a new parameter anyway and only applies to postcodes, should we simply go with `postcode=<country_code>:<postcode>` here?

src/nominatim_api/v1/server_glue.py
```diff
@@ -156,10 +156,17 @@ async def details_endpoint(api: NominatimAPIAsync, params: ASGIAdaptor) -> Any:
...
+            except UsageError as err:
+                params.raise_error(str(err))
         osmtype = params.get('osmtype')
-        if osmtype is None:
-            params.raise_error("Missing ID parameter 'place_id' or 'osmtype'.")
-        place = OsmID(osmtype, params.get_int('osmid'), params.get('class'))
+        if place_ref is None:
+            if osmtype is None:
+                params.raise_error("Missing ID parameter 'place_id', 'place_ref' or 'osmtype'.")
+            place = OsmID(osmtype, params.get_int('osmid'), params.get('class'))
```
lonvia: You can simplify this piece of code with the [walrus opeator](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions):

```
if place_id:
    ...
elif (place_ref := params.get('place_ref')) is not None:
    ...
elif (osmtype := params.get('osmtype')) is not None:
    ...
```

src/nominatim_api/lookup.py
```diff
@@ -301,6 +310,25 @@ async def find_in_postcode(conn: SearchConnection, collector: Collector) -> bool
...
+        ref_tab = sa.func.JsonArrayEach(sa.type_coerce(postcode_refs, sa.JSON))\
+                    .table_valued(sa.column('value', type_=sa.JSON))
+        t = conn.t.postcode
+        sql = sa.select(ref_tab.c.value['i'].as_integer().label('_idx'),
+                        t.c.osm_id, t.c.place_id, t.c.parent_place_id,
+                        t.c.rank_search,
+                        t.c.indexed_date, t.c.postcode, t.c.country_code,
+                        t.c.centroid)\
+                .where(t.c.country_code == ref_tab.c.value['cc'].as_string())\
+                .where(t.c.postcode == ref_tab.c.value['pc'].as_string())
```
lonvia: We explicitly exclude the OSM-derived postcodes here: `t.c.osm_id == None`

src/nominatim_api/types.py
```diff
@@ -439,6 +472,8 @@ def format_excluded(ids: Any) -> List[PlaceRef]:
             elif len(i) > 1 and i[0].upper() in ('N', 'W', 'R') and i[1:].isdigit():
                 if int(i[1:]) > 0:
                     result.append(OsmID(i[0].upper(), int(i[1:])))
+            elif len(i) > 4 and i[0] == 'P' and i[1:3].isalpha() and i[3] == ':' and i[4:]:
```
lonvia: I wonder if a regex might be more readable here.

src/nominatim_api/v1/server_glue.py
lonvia: This needs to emit the new postcode IDs. Given how complicated it has become, a helper function is probably in order.

lonvia (commented):
When I mentioned the spaces in postcode, I did so because I don't think it is a good idea to have spaces in an id, i.e. they should be replaced. Sorry, I should have been more specific. I think replacing with underscore should be fairly safe?

docs/api/Details.md
Itz-Agasta: yes  i was thinking about it too ... it will be much cleaner

Itz-Agasta:
yep.../details now takes postcode=<cc>:<id> directly..... The P prefix is only kept for exclude_place_ids (where the entries are mixed with OSM IDs and need disambiguation)... pls check the regx those are ai generated i never quite learned it well.

2 ci test are failing due to SQLAlchemy 2.x incompatibility tough

Itz-Agasta:
unrelated to this .. was checking the cicd code to figure out the upper failed job but 

https://github.com/osm-search/Nominatim/blob/master/.github/workflows/ci-tests.yml#L412

```yaml             
          - name: Migrate to master version
            run: |
              ./master/bin/nominatim admin --migrate
              ./release/bin/nominatim add-data --file Nominatim/test/testdb/additional_api_test.data.osm
```

here the `migrate` job ran `./master/bin/nominatim admin --migrate` and then used `./release/bin/nominatim add-data ...` again..... so a  migration could succeed but the current code path could still fail afterward..cuz  this does not actually validate that the migrated installation works with the new code and CI would give a false sense of safety for upgrade compatibility..... so shoudlnt this be `/master/bin/nominatim add-data ...` ?

src/nominatim_api/search/geocoder.py
```diff
@@ -53,6 +60,20 @@ async def _resolve_excluded_osm_ids(self) -> None:
...
         osm_ids = [e for e in excluded if isinstance(e, OsmID)]
+        postcode_refs = [e for e in excluded if isinstance(e, PostcodeRef)]
+
+        if postcode_refs:
+            p = self.conn.t.postcode
+            conditions = [
+                sa.and_(p.c.osm_id.is_(None),
+                        p.c.country_code == ref.country_code,
+                        _normalized_postcode_expr(p.c.postcode)
+                        == _normalized_postcode_expr(sa.literal(ref.postcode)))
```
lonvia: Sadly, that won't work because the normalization expression will prevent PostgreSQL from using the index over the postcode column. So, lets just do the conversion space <=> underscore between the postcode reference and the actual saved postcode. I think that should be a 1:1 conversion because there are no underscores in postcodes, so that you can simply do:

```
p.c.postcode == ref.postcode.replace('_', ' ')
```

and we are back in business with the index.

Even better if the PostcodeRef does the replacement in its constructor, so that at this point ref already contains a postcode with spaces.

src/nominatim_api/lookup.py
```diff
@@ -139,10 +143,22 @@ def enumerate_free_osm_ids(self) -> Iterable[Tuple[int, ntyp.OsmID]]:
...
+    def enumerate_free_postcode_refs(self) -> Iterable[Tuple[int, ntyp.PostcodeRef]]:
+        if self.result is None and isinstance(self.place, ntyp.PostcodeRef):
+            return [(0, self.place)]
+        return []
+
 
 Collector = Union[LookupCollector, DetailedCollector]
 
 
+def _normalized_postcode_expr(expr: sa.ColumnElement[str]) -> sa.ColumnElement[str]:
```
lonvia: SQLAlchemy types are difficult because they change so much with the different versions. There are a couple of aliases defined in https://github.com/osm-search/Nominatim/blob/master/src/nominatim_api/typing.py. Use those preferably and things should work out with the CI.

src/nominatim_api/search/geocoder.py
Itz-Agasta: ok so we can removed the SQL normalization helpers entirely..... and use  `PostcodeRef.__post_init__` to convert `_` to spaces on construction, so lookup/exclusion paths compare directly on postcode without any expression wrapping...am i thinking it correctly ? ya it will indeed be much cleaner

src/nominatim_api/lookup.py
Itz-Agasta: If we go with the SQL normalization approach above.. so these helper  funs can be  deleted along it... but ya i will keep these aliases in mind

lonvia:
Looks good now, thanks.

PR #4116, opened 2026-06-03
Add Support for External Postcode Geometry Import via JSONL Files

### LINKED ISSUE #4080: Allow importing areas for external postcodes

AyushDharDubey:
## Summary
Closes #4080
This PR introduces support for importing external postcode data in both csv and jsonl (json lines) formats, allowing for richer postcode geometry import. In JSONL files, we can now provide full geometries, and centroids (optional).

The new JSONL files would need to placed in the project directory with name XX_postcodes_geometry.jsonl.

Currently OSM imported postcodes precede over jsonl imports which precedes csv imports (csv imports are at the same confidence level with guessed postcode areas).

The import logic, storage, and documentation are updated accordingly, and comprehensive tests are added for the new functionality.

Also introduces a new `is_area` flag in `location_postcodes` table.

## AI usage
LLM based CLI tools were extensively used to understand the codebase and find out the relevant files pertaining the targeted changes.

### DIFF
--- a/docs/customize/Postcodes.md
+++ b/docs/customize/Postcodes.md
@@ -1,31 +1,50 @@
 # External postcode data
 
-Nominatim creates a table of known postcode centroids during import. This table
-is used for searches of postcodes and for adding postcodes to places where the
-OSM data does not provide one. These postcode centroids are mainly computed
-from the OSM data itself. In addition, Nominatim supports reading postcode
-information from an external CSV file, to supplement the postcodes that are
-missing in OSM.
+Nominatim creates a table of known postcode centroids and geometries during import.
+This table is used for searches of postcodes and for adding postcodes to places where
+the OSM data does not provide one. These postcode centroids are mainly computed
+from the OSM data itself. In addition, Nominatim supports reading postcode information
+from external files to supplement the postcodes that are missing in OSM.
 
-To enable external postcode support, simply put one CSV file per country into
-your project directory and name it `<CC>_postcodes.csv`. `<CC>` must be the
-two-letter country code for which to apply the file. The file may also be
-gzipped. Then it must be called `<CC>_postcodes.csv.gz`.
+## Supported file formats
 
-The CSV file must use commas as a delimiter and have a header line. Nominatim
-expects three columns to be present: `postcode`, `lat` and `lon`. All other
-columns are ignored. `lon` and `lat` must describe the x and y coordinates of the
-postcode centroids in WGS84.
+Nominatim supports 2 data formats for external import:
 
-The postcode files are loaded only when there is data for the given country
-in your database. For example, if there is a `us_postcodes.csv` file in your
-project directory but you import only an excerpt of Italy, then the US postcodes
-will simply be ignored.
+### JSONL format
+
+The JSONL format may contain the postcode data along with a POLYGON or MULTIPOLYGON geometry
+of the postcode area. To enable external postcode support, put one file per country into
+your project directory and name it `<CC>_postcodes_geometry.<ext>`. `<CC>` must be the
+two-letter country code for which to apply the file. File type may be either `.jsonl` or
+`.jsonl.gz` (gzip compressed). The file must be in UTF-8 encoding and contain one JSON object
+per line with the following structure:
+
+#### GeoJSON Feature
+
+A standard [RFC7946](https://geojson.org) GeoJSON Feature object.
+
+* `geometry`: (required) A GeoJSON geometry (Polygon/MultiPolygon).
+* `properties`: (required) An object containing:
+  * `postcode`: (required) The postcode string.
+  * `lat`, `lon`: (optional) Coordinates for the centroid. If not provided, the centroid
+  is computed from the geometry.
+
+### CSV format
+
+To enable external postcode support, put one file per country into
+your project directory and name it `<CC>_postcodes.<ext>`. `<CC>` must be the
+two-letter country code for which to apply the file. File type may be either `.csv` or `.csv.gz`
+(gzip compressed). The CSV file must use commas as a delimiter and have a header line. Nominatim
+expects three columns to be present: `postcode`, `lat` and `lon`. All other columns are ignored.
+`lon` and `lat` must describe the x and y coordinates of the postcode centroids in WGS84.
+
+The postcode area is assumed to be in a buffer around the centroid (typically 5km).
+
+## Usage
 
 As a rule, the external postcode data should be put into the project directory
 **before** starting the initial import. Still, you can add, remove and update the
-external postcode data at any time. Simply
-run:
+external postcode data at any time. Simply run:
 
 ```
 nominatim refresh --postcodes
--- a/docs/develop/Database-Layout.md
+++ b/docs/develop/Database-Layout.md
@@ -119,13 +119,15 @@ the placex table. Only the following columns are special:
 Address interpolations are always ways in OSM, which is why there is no column
 `osm_type`.
 
-The **location_postcodes** table holds computed postcode assembled from the
-postcode information available in OSM. When a postcode has a postcode area
-relation, then the table stores its full geometry. For all other postcode
-the centroid is computed using the position of all OSM object that reference
-the same postoce. The `osm_id` field can be used to distinguish the two.
-When set, it refers to the OSM relation with the postcode area.
-The meaning of the columns in the table is again the same as that of the
+The **location_postcodes** table holds computed postcode assembled from the postcode information
+available in OSM. When a postcode has a postcode area relation, or when the postcode geometry is
+[imported via JSONL files](../customize/Postcodes.md#jsonl-format) then the table stores
+its full geometry. For all other postcodes the centroid is computed using the position of all OSM
+objects that reference the same postcode. The `osm_id` and `is_area` fields can be used to
+distinguish the two. When `osm_id` is set, it refers to the OSM relation with the postcode area,
+and `is_area` is `true` for postcodes with a mature geometry (either from a postcode OSM area
+relation or imported via JSONL), `false` for postcodes without a mature geometry (guessed
+postcode geometries). The meaning of other columns in the table is again the same as that of the
 placex table.
 
 Every place needs an address, a set of surrounding places that describe the
--- a/src/nominatim_db/tools/migration.py
+++ b/src/nominatim_db/tools/migration.py
@@ -507,3 +507,15 @@ def create_associated_street_triggers(conn: Connection, config: Configuration, *
         cur.execute("""CREATE OR REPLACE TRIGGER place_associated_street_update
                        AFTER INSERT OR DELETE ON place_associated_street
                        FOR EACH ROW EXECUTE FUNCTION invalidate_associated_street_members()""")
+
+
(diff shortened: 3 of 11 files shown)


src/nominatim_db/tools/migration.py
```diff
@@ -507,3 +507,15 @@ def create_associated_street_triggers(conn: Connection, config: Configuration, *
...
+@_migration(5, 4, 99, 0)
+def alter_location_postcode_add_area_flag(conn: Connection, **_: Any) -> None:
+    """ Add is_area flag to location_postcode table
+    """
+    with conn.cursor() as cur:
+        cur.execute("""ALTER TABLE location_postcodes
+                       ADD COLUMN IF NOT EXISTS is_area BOOLEAN NOT NULL DEFAULT FALSE""")
+        cur.execute("""UPDATE location_postcodes
+                          SET is_area = true
+                          WHERE osm_id IS NOT NULL""")
```
lonvia: Can you check if the column already exists using [`table_has_column()`](https://github.com/osm-search/Nominatim/blob/1938aeb977d1aafb8b04e43d046c13988508c1ab/src/nominatim_db/db/connection.py#L53). The update is slightly expensive and we wouldn't want to do it again. (There are certain constellations where a migration might be run again.)

src/nominatim_db/tools/postcodes.py
```diff
@@ -291,12 +295,140 @@ def _update_postcode_areas(conn: Connection, analyzer: AbstractAnalyzer,
...
+
+                pcs.append({
+                    'postcode': normalized,
+                    'geometry': json.dumps(geometry),
+                    'lon': lon,
+                    'lat': lat,
+                })
+
+        _insert_external_postcode_areas(conn, country,
+                                        matcher.get_postcode_extent(country),
```
lonvia, line 440: I think we should rather compute the actual extent from the area geometries here. Get the maximum of width and height for the bbox of each postcode area. That's easy to compute. And then, I don't know what's a good guess, use the 75 percentile over all values maybe.

src/nominatim_db/tools/postcodes.py
```diff
@@ -291,12 +295,140 @@ def _update_postcode_areas(conn: Connection, analyzer: AbstractAnalyzer,
...
+                                    is_initial: bool) -> None:
+    if not pcs:
+        return
+
+    if not is_initial:
+        with conn.cursor() as cur:
+            # Wipe previous external polygons for this country during an incremental update
+            cur.execute("""DELETE FROM location_postcodes
+                           WHERE country_code = %s AND osm_id IS NULL AND is_area
+                        """, (country_code,))
```
lonvia: Only updating the ones that changed isn't an option?

docs/customize/Postcodes.md
```diff
@@ -1,31 +1,51 @@
...
+#### GeoJSON Feature
+
+A standard [RFC7946](https://geojson.org) GeoJSON Feature object .
+
+* `geometry`: A GeoJSON geometry (Polygon/MultiPolygon).
+* `properties`: An object containing:
+  * `postcode`: (required) The postcode string.
+  * `lat`, `lon`: (optional) Coordinates for the centroid. If not provided, the centroid
+  is computed from the geometry. Thus centroid coordinates and geometry independently optional,
+  however at least one of them is required.
```
lonvia: That latter sentence isn't correct anymore, right? Geometry is now required.

docs/develop/Database-Layout.md
```diff
@@ -119,13 +119,15 @@ the placex table. Only the following columns are special:
...
-relation, then the table stores its full geometry. For all other postcode
-the centroid is computed using the position of all OSM object that reference
-the same postoce. The `osm_id` field can be used to distinguish the two.
-When set, it refers to the OSM relation with the postcode area.
-The meaning of the columns in the table is again the same as that of the
+The **location_postcodes** table holds computed postcode assembled from the postcode information
+available in OSM. When a postcode has a postcode area relation, or when the postcode geometry is
+[imported via JSONL files](../customize/Postcodes.md#jsonl-format) then the table stores
+its full geometry. For all other postcode the centroid is computed using the position of all OSM
+object that reference the same postcode. The `osm_id` and `is_area` fields can be used to
```
lonvia: Would you mind fixing my awful use of grammar here: other postcode -> other postcodes, all OSM object -> all OSM objects

src/nominatim_db/tools/migration.py
```diff
@@ -507,3 +507,15 @@ def create_associated_street_triggers(conn: Connection, config: Configuration, *
         cur.execute("""CREATE OR REPLACE TRIGGER place_associated_street_update
                        AFTER INSERT OR DELETE ON place_associated_street
                        FOR EACH ROW EXECUTE FUNCTION invalidate_associated_street_members()""")
+
+
+@_migration(5, 4, 99, 0)
```
lonvia: It should be better documented how versions with migration work. The [docs for NominatimVersion](https://github.com/osm-search/Nominatim/blob/1938aeb977d1aafb8b04e43d046c13988508c1ab/src/nominatim_db/version.py#L16) explain it a bit. When adding a migration you must:

* Raise the [db-patch-level of the version](https://github.com/osm-search/Nominatim/blob/1938aeb977d1aafb8b04e43d046c13988508c1ab/src/nominatim_db/version.py#L58). If the current version is X.Y.0.0, then use X.Y.99.0, otherwise increase the last number by 1.
* Here in the migration use the new version you have just changed to.

So this migration should be for 5.3.99-1.

lonvia (commented):
Design-wise this is in good shape now. The only large question that remains open is that of minimizing changes during update. To that end it would also good to have a few more tests for updates, tests similar to `test_postcodes_add_point_within_area()`.

src/nominatim_db/tools/postcodes.py
AyushDharDubey: how about calculating individual extents for each postcode and using that to get search ranks explicitly for each postcode?

docs/customize/Postcodes.md
AyushDharDubey: Ah yes, that's a mistake.

src/nominatim_db/tools/postcodes.py
AyushDharDubey: I think the intended behavior should be to delete any postcodes that were imported during the previous db refresh call but no longer appear in the latest JSONL file; So as to ensure that postcodes removed (as a result of may be a delimitation exercise in the country) from the authoritative source no longer persist.

AyushDharDubey: So perhaps it would be better to simply truncate all geometry equipped postcodes for the country and recreate them from the latest jsonl file?

src/nominatim_db/tools/postcodes.py
lonvia: It would make for simpler code but different search ranks for postcode areas in the same country would be confusing. Oh well, lets stay with the generic extent from the matcher for now.

src/nominatim_db/tools/postcodes.py
lonvia: You are currently wiping all postcodes and reinserting them, aren't you? Or am I understanding this code wrong?

There are actually two reasons, why an external postcode area disappears: it is no longer in the file or it is now mapped as a postcode area in OSM (which has precedence). I think the latter case is handled by the [postcode trigger](https://github.com/osm-search/Nominatim/blob/400e0cabb647a1d42c93ca3a5369115b86f4a808/lib-sql/functions/postcode_triggers.sql#L62) during the update of the OSM postcode areas. You need to only catch the former.

Once you have the deletion restricted to the once really needed deleting, you can simply insert everything unconditionally and the trigger will do the right thing to insert/update.

At least that what the intention of the code was. I feel much better if we have tests that proof that it really does get the updates right.

AyushDharDubey: My bad, actually I was stupidly thinking deleting all entries and inserting new postcodes would be optimal than searching for the postcodes which are actually abandoned and need deletion.
But the (2*n+x) db operations in the former case would be more expensive than searching for the abandoned postcodes, right?
I will update this chunk and add the tests too.

src/nominatim_db/tools/postcodes.py
AyushDharDubey: Umm I can give it a try to do it in this PR only, otherwise it will be messy to update any search ranks of already imported area with generic extents in production, between the two PRs.

src/nominatim_db/tools/postcodes.py
lonvia: There is a hidden cost here: updates also invalidate rows in the placex table to make sure they get any changes for postcodes and these queries can get expensive.

AyushDharDubey: Okay, one last question:

do we also need to segregate such postcodes which have not changed (are the same as that is location_postcodes) and filter out such from our sql query?
OR
we only need to segregate the postcodes which are not in location_postcodes (which need deletion) and the postcodes which are in location_postcodes (no matter if the geometry has changed or not, send them to insert)?

lonvia: You only need to minimize the DELETE query to those actually needing deletion. Everything else you can just reinsert. The insert trigger will compare with any existing entry and take appropriate action (including just ignoring the entry if nothing has changed).

src/nominatim_db/tools/postcodes.py
lonvia: It's fine. Lets leave this as is. People can configure the extend in the country settings if necessary.

src/nominatim_db/tools/postcodes.py
```diff
@@ -129,26 +140,26 @@ def commit(self, conn: Connection, analyzer: AbstractAnalyzer,
...
+        with _open_external_file(fname) as csvfile:
             reader = csv.DictReader(csvfile)
             for row in reader:
                 if 'postcode' not in row or 'lat' not in row or 'lon' not in row:
                     LOG.warning("Bad format for external postcode file for country '%s'."
                                 " Ignored.", self.country)
                     return
                 postcode = analyzer.normalize_postcode(row['postcode'])
-                if postcode not in self.collected:
+                if postcode not in self.collected and postcode not in self.exclude:
```
AyushDharDubey, line 162: The exclude parameter was not being checked against during the csv import, I assume this was a mistake and not an expected behavior.

lonvia: Yes, your fix looks to be correct.

lonvia (approved):
Looks good now.

Many thanks for baring with me while sorting out the finer details of the design.

PR #4124, opened 2026-06-10
Add sanitizer to create 'Rodovia BR-XXX' name in Brazil

### LINKED ISSUE #4075: Systematic mismatches between addr:street and the name of the highway
Please take a look in all these addresses besides the BR-116 highway: https://nominatim.org/qa/#map=18.13/-3.77/-38.52&layer=addr_street_wrong_name

In Brazil, highways can have a name (for the whole highway, for just a part, or just in an urban area). The BR-116 case is interesting because it's the longest one in the country, so therefore it has multiple names. So, the addresses alongside it can be either "Rodovia BR-116" ("Rodovia" + space + ref), or the name if that section has a name=* tag (in that bbox, it could be "Rodovia Santos Dumont"). Highways only have name=* tag if there is actually a name for it. In the case below, the highway is properly mapped because that segment has no name, only ref tag.

https://nominatim.org/qa/#map=17.17/-28.65/-53.58&layer=addr_street_wrong_name

(you can find it everywhere in the whole country, it's the official address standard)

Therefore, can't we not flag these addresses? I might help with it, but I'm not sure where? Is that here? https://github.com/osm-search/Nominatim-Data-Analyser/blob/main/src/nominatim_data_analyser/rules_specifications/addr_street_wrong_name.yaml Or shall a specific country rule be created?

Also, I suppose this kind of stuff might happen in many countries (didn't have time to look into it), so perhaps it's useful to have a more general rule, like Highway (any language) + ref, or something like that?

lonvia:
Adds a sanitizer specifically for Brazil: when a major highway has a ref of the form `BR-*`, then a searchable name `Rodovia BR-*` is added. This allows the street to be found under the name with the Rodovia prefix and, more importantly, ensures that addresses with `addr:street=Rodovia BR-*'` are correctly matched up with the street.

There is a corresponding PR for Nominatim QA at https://github.com/osm-search/Nominatim-Data-Analyser/pull/73

Closes #4075.

### DIFF
--- a/settings/country_settings.yaml
+++ b/settings/country_settings.yaml
@@ -220,6 +220,12 @@ br:
     postcode:
       pattern: "(ddddd)-?(ddd)"
       output: \1-\2
+    sanitizers:
+      - step: derive-names
+        filter-kind: ref
+        filter-rank: "26"
+        name-pattern: ([A-Z]{2,3}-\d+)
+        variants: Rodovia \1
 
 
 # The Bahamas (The Bahamas)


matheusgomesms:
Hi Sarah, thanks for the fix!

It's going to be a life changer for the BR community! However, one little thing: can it be made more general, to encompass all the highways (not only the federal ones)? Highways in Brazil are like this:

Federal highways: `BR-XXX`
State highways: usually `AA-XXX`, where AA is the acronym for each state, like São Paulo = SP, so `SP-123`. However, there are some cases where the state highways have three letters instead of two (depending on state and type of highway).

And there are tons of addresses with these highways (example: https://nominatim.org/qa/#map=16.63/-28.67/-53.60&layer=addr_street_wrong_name) so perhaps the safest bet here would be `"Rodovia" + ref` instead? With that, not only federal highways will have this fix, but also all highways in the entire country.

santamariense:
There are cases of refs being used at the municipality level as well. Generically, it is
 _prefix + <any ref value (or one of its values, if multivalued)>_
where the prefix can be "Rodovia" (more common). But it can also be "Estrada" sometimes. 

Thanks for working on this.

lonvia:
I've extended the regex to accept any 2-3 letters. That should be okay according to the data. I don't want to extend it to any `ref` because that tag tends to get a lot of odd content.

I can't find a lot of "Estrada" in the data. And the ones that are there are either dubious values for `addr:street` or something like "Estrada Para Araticum e Acesso A BR-110". Maybe this is better tagged as an exception then.

santamariense:
I agree with your statement. Regarding "Estrada", it may occur in some cases, but I'm not sure how often. For now, I think it is solved.

PR #4132, opened 2026-06-21
Fix Windows `ProactorEventLoop` incompatibility with psycopg async

### LINKED ISSUE #3598: Appropriate asyncio.set_event_loop_policy to make scripts usable on Windows

AyushDharDubey:
## Summary
Closes #3598 

Replaced the past attempt of fixing this in the test conftest using `asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())`, as loop policies are in line to be deprecated with forthcoming Python 3.16.

Added a small helper (`get_loop_factory()`) that returns a `SelectorEventLoop`-based factory on Windows and `None` everywhere else, on other platforms this leaves `asyncio.run()` to choose it's own optimal `loop_factory` depending on platform.

The `pytest-asyncio` suite gets a [pytest_asyncio_loop_factories](https://pytest-asyncio.readthedocs.io/en/v1.4.0/how-to-guides/run_test_with_specific_loop_factories.html) hook in conftest.

Added Windows install jobs to CI basically mirroring the existing Ubuntu ones.

Also encountered a regex issue with `flex-base.lua` while running on Windows: the topic file path matching only handled `/` and broke on Windows paths `\\`.

## AI usage
Not used.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -33,7 +33,7 @@ jobs:
                   path: nominatim-src.tar.bz2
                   retention-days: 1
 
-    tests:
+    tests-ubuntu:
         needs: create-archive
         strategy:
             matrix:
@@ -142,6 +142,14 @@ jobs:
 
     tests-windows:
         needs: create-archive
+        strategy:
+            matrix:
+                python: ['3.10', '3.14']
+                include:
+                    - python: '3.10'
+                      pyicu: 'v2.15/pyicu-2.15-cp310-cp310-win_amd64.whl'
+                    - python: '3.14'
+                      pyicu: 'v2.16.2/pyicu-2.16.2-cp314-cp314-win_amd64.whl'
         runs-on: windows-latest
 
         steps:
@@ -159,7 +167,13 @@ jobs:
             - name: Set up Python
               uses: actions/setup-python@v6
               with:
-                  python-version: '3.14'
+                  python-version: ${{ matrix.python }}
+
+            - uses: actions/cache@v5
+              with:
+                  path: |
+                     C:\osm2pgsql
+                  key: osm2pgsql-windows-1
 
             - name: Install Spatialite
               run: |
@@ -169,8 +183,10 @@ jobs:
 
             - name: Install osm2pgsql
               run: |
-                  Invoke-WebRequest -Uri "https://osm2pgsql.org/download/windows/osm2pgsql-latest-x64.zip" -OutFile "osm2pgsql.zip"
-                  Expand-Archive -Path "osm2pgsql.zip" -DestinationPath "C:\osm2pgsql"
+                  if (-not (Test-Path "C:\osm2pgsql")) {
+                    Invoke-WebRequest -Uri "https://osm2pgsql.org/download/windows/osm2pgsql-latest-x64.zip" -OutFile "osm2pgsql.zip"
+                    Expand-Archive -Path "osm2pgsql.zip" -DestinationPath "C:\osm2pgsql"
+                  }
                   $BinDir = Get-ChildItem -Path "C:\osm2pgsql" -Recurse -Filter "osm2pgsql.exe" | Select-Object -ExpandProperty DirectoryName | Select-Object -First 1
                   if (-not $BinDir) {
                       Write-Error "Could not find osm2pgsql.exe"
@@ -186,19 +202,19 @@ jobs:
 
             - name: Install PyICU from wheel
               run: |
-                  python -m pip install https://github.com/cgohlke/pyicu-build/releases/download/v2.16.0/pyicu-2.16-cp314-cp314-win_amd64.whl
-                  
+                  python -m pip install https://github.com/cgohlke/pyicu-build/releases/download/${{ matrix.pyicu }}
+
             - name: Install test prerequisites
               run: |
                   python -m pip install -U pip
-                  python -m pip install pytest pytest-asyncio "psycopg[binary]!=3.3.0" python-dotenv pyyaml jinja2 psutil sqlalchemy pytest-bdd falcon starlette uvicorn asgi_lifespan aiosqlite osmium mwparserfromhell
+                  python -m pip install pytest pytest-asyncio async-timeout "psycopg[binary]!=3.3.0" python-dotenv pyyaml jinja2 psutil sqlalchemy pytest-bdd falcon starlette uvicorn asgi_lifespan aiosqlite osmium mwparserfromhell
 
             - name: Python unit tests
               run: |
                   python -m pytest test/python -k "not (import_osm or run_osm2pgsql)"
               working-directory: Nominatim
 
-    install:
+    install-ubuntu:
         runs-on: ubuntu-latest
         needs: create-archive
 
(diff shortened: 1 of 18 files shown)


.github/workflows/ci-tests.yml
```diff
@@ -145,12 +145,13 @@ jobs:
...
-                  name: full-source
+            - name: Enable git symlinks
+              run: git config --global core.symlinks true
 
-            - name: Unpack Nominatim
-              run: tar xf nominatim-src.tar.bz2
+            - uses: actions/checkout@v6
+              with:
+                  path: Nominatim
+                  submodules: true
```
lonvia: I presume you switch back to git checkout here because the archive doesn't do well with symlinks? Any chance we can change the way the archive works to make it work on Windows? The idea here was to reduce the number of downloads from https://www.nominatim.org/data/country_grid.sql.gz

.github/workflows/ci-tests.yml
```diff
@@ -329,7 +335,120 @@ jobs:
...
+            - name: Set up Python
+              uses: actions/setup-python@v6
+              with:
+                  python-version: '3.14'
+
+            - name: Install Spatialite
+              run: |
+                  Invoke-WebRequest -Uri "https://www.gaia-gis.it/gaia-sins/windows-bin-amd64/mod_spatialite-5.1.0-win-amd64.7z" -OutFile "spatialite.7z"
+                  7z x spatialite.7z -o"C:\spatialite"
+                  echo "C:\spatialite\mod_spatialite-5.1.0-win-amd64" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
```
lonvia: I don't think spatialite is needed for this step?

.github/workflows/ci-tests.yml
```diff
@@ -329,7 +335,120 @@ jobs:
...
+                  Invoke-WebRequest -Uri "https://osm2pgsql.org/download/windows/osm2pgsql-latest-x64.zip" -OutFile "osm2pgsql.zip"
+                  Expand-Archive -Path "osm2pgsql.zip" -DestinationPath "C:\osm2pgsql"
+                  $BinDir = Get-ChildItem -Path "C:\osm2pgsql" -Recurse -Filter "osm2pgsql.exe" | Select-Object -ExpandProperty DirectoryName | Select-Object -First 1
+                  if (-not $BinDir) {
+                      Write-Error "Could not find osm2pgsql.exe"
+                      exit 1
+                  }
+                  echo "$BinDir" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+                  $FullExePath = Join-Path $BinDir "osm2pgsql.exe"
+                  echo "NOMINATIM_OSM2PGSQL_BINARY=$FullExePath" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
```
lonvia, line 396: This would also be a candidate for a cache action to avoid repeated download.

src/nominatim_db/utils/asyncio_utils.py
```diff
@@ -0,0 +1,22 @@
...
+"""
+import sys
+import asyncio
+import selectors
+from typing import Callable, Optional
+
+
+def get_loop_factory() -> Optional[Callable[[], asyncio.AbstractEventLoop]]:
+    """Returns a compatible loop factory for Python 3.12+ on Windows,
+    or None for Unix based systems."""
```
lonvia: Nitpicking: please follow formatting conventions:

```
""" Returns a compatible loop factory for Python 3.12+ on Windows,
    or None for Unix based systems.
"""
```

test/bdd/test_osm2pgsql.py
```diff
@@ -86,7 +87,14 @@ def update_from_osm_file(db_conn, def_config, osm2pgsql_options, opl_writer, doc
...
     """
     create_table_triggers(db_conn, def_config)
-    asyncio.run(load_data(def_config.get_libpq_dsn(), 1))
+
+    loop_factory = get_loop_factory()
+    if loop_factory is not None:
+        asyncio.run(load_data(def_config.get_libpq_dsn(), 1),
+                    loop_factory=loop_factory)  # type: ignore[call-arg]
+    else:
+        asyncio.run(load_data(def_config.get_libpq_dsn(), 1))
```
lonvia: Given, how often this code is now duplicated, would it make sense to have a `asyncio_run()` helper function in the new `asyncio_utils` module?

test/python/conftest.py
```diff
@@ -10,7 +10,9 @@
 from pathlib import Path
 
 if sys.platform == 'win32':
-    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
+    # pytest-asyncio hook for event loop factory.
+    def pytest_asyncio_loop_factories(config, item):
```
lonvia: This hook is a recent addition to pytest-asyncio in [version 1.4](https://github.com/pytest-dev/pytest-asyncio/releases/tag/v1.4.0). I think we need a version switch here, at least for a couple of months.

src/nominatim_db/clicmd/replication.py
```diff
@@ -185,5 +185,11 @@ def run(self, args: NominatimArgs) -> int:
         if args.check_for_updates:
             return self._check_for_updates(args)
 
-        asyncio.run(self._update(args))
+        from ..utils.asyncio_utils import get_loop_factory
```
lonvia: I'd lean towards moving this include into the global import section at the beginning of the file. It's just a small utility module not a big one like the other modules that are locally imported within the functions.

Same for all the other uses of this import.

lonvia (commented):
This looks good in general. Given that the `loop_factory` parameter only exists in Python >= 3.12, does that mean we get a minimum requirement of 3.12 for Windows? That wouldn't be a showstopper, it just should be noted in the [documentation](https://github.com/osm-search/Nominatim/blob/a9eb079fc139dfeb6367bc84ccaae5520f623dbf/docs/admin/Installation.md?plain=1#L30). (And possibly in the pyproject.toml files? Not sure if the version spec allows that.)

.github/workflows/ci-tests.yml
AyushDharDubey: Sorry, this was a mistake actually, during my testing phase, I think I had got the path wrong at some step and assumed that the symlinks are failing.

I just checked the normal extraction of the archive is preserving the symlinks just fine here.

AyushDharDubey:
> does that mean we get a minimum requirement of 3.12 for Windows?

We could also fall back to our old `asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())` on lower versions.

AyushDharDubey:
Would you happen to know why we are excluding these tests in windows job?

https://github.com/osm-search/Nominatim/blob/a9eb079fc139dfeb6367bc84ccaae5520f623dbf/.github/workflows/ci-tests.yml#L198

lonvia:
> Would you happen to know why we are excluding these tests in windows job?

They are using a shell script for mocking osm2pgsql which wouldn't work under windows, see https://github.com/osm-search/Nominatim/blob/master/test/python/tools/conftest.py#L11. Rather something for a different PR.

lonvia (commented):
Looks good, just one tiny thing: now that Windows runs against two different Python version, it would be good to make them smallest supported and latest release, i.e. at the moment that be 3.9 and 3.14.

lonvia:
Oh, I see the problem. Use pyicu 3.15 instead with 3.9?

AyushDharDubey:
While testing, I hit a separate compatibility gap: spatialite is breaking on Windows Python 3.9 with `updateTableTriggers: "no such module: rtree"`, on further digging up it came up that the bundled SQLite for the python Windows 3.9 build doesn't have `SQLITE_ENABLE_RTREE` compiled in, it is supported in 3.10 onward.

Figure we will need to upgrade the minimum requirement anyways, so should we go with directly 3.12 for windows to line up with the loop_factory constraint above.

lonvia:
If you can make it work with 3.10 that would be great. 3.9 is on the way out, I don't care about that, but 3.10 is still an officially supported Python version. If there is no major blocker then it will be still at least 2 years before 3.12 becomes the minimal requirement for Nominatim  Linux.

That said, if there is more compatibility issues, then a 3.12 requirement for Windows is still perfectly fine.

AyushDharDubey:
3.10 can work then.

lonvia (commented):
Looks good now. Thank you.

Issue #4134, opened 2026-06-25
Documentation Ambiguity and Outdated Scripts for Updating Multi-Region Databases


andymilga:
I am currently setting up a Nominatim instance (v5.3) with multiple region imports and want to enable automatic updates. However, I've encountered significant difficulties due to unclear documentation and what appear to be outdated processes.

The Problem:

The official documentation for setting up a multi-region database with updates (https://nominatim.org/release-docs/latest/admin/Advanced-Installations/#importing-multiple-regions-with-updates) relies on two scripts from the utils directory: import_multiple_regions.sh and update_database.sh .

This approach presents several issues:

Outdated Scripts: These scripts appear to be legacy PHP scripts that are no longer compatible with the current command structure of Nominatim (e.g., nominatim import --continue). Their primary purpose seems to be orchestrating the download and merging of PBF files, but the underlying commands are outdated and fail.

Lack of Clarity on Update Mechanism: The documentation for multi-region updates is too brief. It states that update_database.sh will "get diffs from the replication server" but provides little detail on how this process works for multiple regions . For instance:
  - How does the script handle separate sequence.state files for each region?
  - How can this process be run continuously, similar to the nominatim replication command used for single-region setups ?

# PBF files in my project directory
europe-latest.osm.pbf (34.5 GB)
kazakhstan-latest.osm.pbf
kyrgyzstan-latest.osm.pbf
mongolia-latest.osm.pbf
tajikistan-latest.osm.pbf
turkmenistan-latest.osm.pbf
uzbekistan-latest.osm.pbf

# My import command (works fine)
nominatim import --osm-file europe-latest.osm.pbf --osm-file kazakhstan-latest.osm.pbf --osm-file kyrgyzstan-latest.osm.pbf --osm-file mongolia-latest.osm.pbf --osm-file tajikistan-latest.osm.pbf --osm-file turkmenistan-latest.osm.pbf --osm-file uzbekistan-latest.osm.pbf --threads $(nproc)

mtmail:
> These scripts appear to be legacy PHP scripts that are no longer compatible with the current command structure of Nominatim (e.g., nominatim import --continue). 

The script are shell scripts (`*.sh`), not PHP. I don't see `--continue` used in the scripts.

> Their primary purpose seems to be orchestrating the download and merging of PBF files, but the underlying commands are outdated and fail.

Downloading and merging PBF files is indeed required. Which failures did you see?

> How does the script handle separate sequence.state files for each region?

[...]

PR #4135, opened 2026-06-25
utils - documentation still mentioned PHP scripts

mtmail:
The documentation inside `utils/*.sh` script still mentioned old Nominatim commands. The scripts themselves were updated 5 years ago already.

My automatic linter added whitespace changes, too. I hope that's fine.

No AI used.

For https://github.com/osm-search/Nominatim/issues/4134 where I suspect an LLM might've gotten confused about PHP being mentioned.

### DIFF
--- a/utils/import_multiple_regions.sh
+++ b/utils/import_multiple_regions.sh
@@ -8,7 +8,7 @@
 
 #     *) Set up sequence.state for updates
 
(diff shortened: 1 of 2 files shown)


lonvia:
> I suspect an LLM might've gotten confused about PHP being mentioned.

Oh, that would be a good reason to keep the bad comments. ;)

PR #4106, opened 2026-05-27
Add ltree categories column for multi-tag OSM object support

Itz-Agasta:
## Summary
 
Adds a `categories ltree[]` column to the `place` and `placex` tables that stores derived OSM categories as ltree paths (for example, `osm.amenity.restaurant`).

This replaces the previous one-row-per-tag model. Multi-tag OSM objects (for example, a node tagged with both `amenity=restaurant` and `tourism=hotel`) are now merged into a single row containing all categories in the `categories` array.

<p align="center">
  Lua merge model
  <br>
  <em>OLD: N tags → N rows. NEW: N tags → 1 row — all categories collected in a single Lua pass.</em>
</p>

### Schema
- Added a `categories ltree[]` column to the `place` and `placex` tables.
- Added a GiST index on the `categories` column.
- Requires the PostgreSQL `ltree` extension.

### Lua import
- Added `sanitize_label()` and `get_category()` to derive ltree categories from OSM class/type tags.
- Updated `process_tags()` to collect all categories and write a single merged row.
- Replaced the previous `write_place()` / `write_row()` one-row-per-tag import model.

### SQL ranking & triggers
- Updated `is_rankable_place()`, `compute_place_rank()`, and the `placex_update` trigger to use ltree containment operators (`<@`) instead of class/type comparisons.
- `compute_place_rank()` now iterates over all categories and selects the best ranking category.

### Indexer
- Updated `BoundaryRunner` queries to use:

  ```sql
  categories <@ 'osm.boundary.administrative'
  ```

### Migration
- Added a migration for existing databases.
- Backfills `categories` from existing `class`/`type` values and `extratags->'place'` for rows with `rank_address < 26`.
- Creates indexes after the bulk `UPDATE`.

## AI usage
Most of the bulk test file rework was generated with AI.... Tough I manually reviewed every change, verified its correctness, and decided which tests should be updated vs skipped. The bug fixes and the final implementation decisions were my own......I also used AI to help with some code simplifications, preparing small patches, analyzing logs to better understand issues, and brainstorming possible approaches and solutions.

## Contributor guidelines (mandatory)

- [X] I have adhered to the [coding style](https://github.com/osm-search/Nominatim/blob/master/CONTRIBUTING.md#coding-style)
- [X] I have [tested](https://github.com/osm-search/Nominatim/blob/master/CONTRIBUTING.md#testing) the proposed changes
[...]

PR #4146, opened 2026-07-06
Use ltree categories column instead of place_classtype tables

Itz-Agasta:
## Summary
Follow-up to #4106. The search side no longer reads the obsolete `place_classtype_*` tables. `PoiSearch` and `NearSearch` now query the `placex.categories` ltree[] column via `categories <@ 'osm.<class>.<type>'`, backed by the `idx_placex_categories` GiST index.

The `place_classtype_*` tables and their generation code (special-phrase importer, sqlite export) are untouched here and will be removed in a separate PR .

- New `CategoryArray` SQLAlchemy type: native `ltree[]` on Pg, comma-separated text on SQLite.
- The lookup ltree is built with the same normalization as the Lua  importer (`sanitize_label`/`get_category`:  hyphen ------>underscore, invalid ----->`yes`).
- **SQLite has no ltree**, so `CategoryMatch` compiles to class/type equality there (columns still present and populated) (May change).
- **Merged rows:** an object tagged e.g. `club=scout` + `amenity=scout_hut`  is a single placex row whose class/type winner is `amenity`. The `<@` query matches it correctly, but the result must report the *searched* category, so both searches now set `result.category` explicitly. This fixes and un-skips the `[club=scout]` BDD scenario.

## AI usage
None

### DIFF
--- a/src/nominatim_api/search/db_searches/base.py
+++ b/src/nominatim_api/search/db_searches/base.py
@@ -7,15 +7,15 @@
 """
 Interface for classes implementing a database search.
 """
-from typing import Any, Callable, List, Optional
+from typing import Callable, List
 import abc
 import re
 
 import sqlalchemy as sa
-from sqlalchemy.ext.compiler import compiles
 
 from ...typing import SaFromClause, SaSelect, SaColumn, SaExpression, SaLambdaSelect
 from ...sql.sqlalchemy_types import Geometry
+from ...sql.sqlalchemy_functions import CategoryMatch
 from ...connection import SearchConnection
 from ...types import SearchDetails, DataLayer, GeometryFormat
 from ...results import SearchResults
@@ -34,59 +34,28 @@ def _sanitize_label(part: str) -> str:
     return part.replace('-', '_')
 
 
-def category_ltree(cls: str, typ: str) -> Optional[str]:
+def is_ltree_category(cls: str) -> bool:
+    """ Check whether the given class can form a valid ltree category label.
+        Types are always sanitized, so only the class may render a category
+        unrepresentable. Non-conforming categories are rejected during token
+        collection so that search never has to fall back to an unindexed query.
+    """
+    return bool(cls) and not _LTREE_INVALID.search(cls)
+
+
+def category_ltree(cls: str, typ: str) -> str:
     """ Build the 'osm.<class>.<type>' ltree string for a class/type pair,
-        mirroring the Lua get_category(). Returns None when the class cannot
-        form a valid ltree label, in which case no stored category can match.
+        mirroring the Lua get_category(). The class must be a valid ltree
+        label (see is_ltree_category()).
     """
-    if not cls or _LTREE_INVALID.search(cls):
-        return None
     return f'osm.{_sanitize_label(cls)}.{_sanitize_label(typ)}'
-
-
-class CategoryMatch(sa.sql.expression.FunctionElement[Any]):
-    """ Match a placex row against a single (class, type) category.
-
-        On PostgreSQL this queries the ltree 'categories' column using the
-        GiST index (``categories <@ 'osm.<class>.<type>'``). On SQLite, which
-        has no ltree support, it falls back to class/type equality; those
-        columns are still present and populated.
-    """
-    inherit_cache = True
-
-    def __init__(self, table: SaFromClause, ltree: str, cls: str, typ: str) -> None:
-        super().__init__(table.c.categories, sa.literal(ltree),
-                         table.c.class_, sa.literal(cls),
-                         table.c.type, sa.literal(typ))
-
-
-@compiles(CategoryMatch)
-def _default_category_match(element: CategoryMatch,
-                            compiler: 'sa.Compiled', **kw: Any) -> str:
-    cats, ltree, _, _, _, _ = list(element.clauses)
-    return "(%s <@ (%s)::ltree)" % (compiler.process(cats, **kw),
-                                    compiler.process(ltree, **kw))
-
-
-@compiles(CategoryMatch, 'sqlite')
-def _sqlite_category_match(element: CategoryMatch,
-                           compiler: 'sa.Compiled', **kw: Any) -> str:
-    _, _, cls_col, cls_lit, typ_col, typ_lit = list(element.clauses)
-    return "(%s = %s AND %s = %s)" % (compiler.process(cls_col, **kw),
-                                      compiler.process(cls_lit, **kw),
-                                      compiler.process(typ_col, **kw),
-                                      compiler.process(typ_lit, **kw))
 
 
 def category_filter(table: SaFromClause, cls: str, typ: str) -> SaExpression:
     """ Build a boolean expression selecting rows of the given class/type
-        category. Uses the ltree categories column on PostgreSQL and falls
-        back to class/type equality where no ltree category can exist.
+        category via the ltree categories column (class/type on SQLite).
     """
-    ltree = category_ltree(cls, typ)
-    if ltree is None:
-        return sa.and_(table.c.class_ == cls, table.c.type == typ)
-    return CategoryMatch(table, ltree, cls, typ)
+    return CategoryMatch(table, category_ltree(cls, typ), cls, typ)
 
 
 class AbstractSearch(abc.ABC):
--- a/src/nominatim_api/search/db_searches/base.py
+++ b/src/nominatim_api/search/db_searches/base.py
@@ -7,16 +7,86 @@
 """
 Interface for classes implementing a database search.
 """
-from typing import Callable, List
+from typing import Any, Callable, List, Optional
 import abc
+import re
 
 import sqlalchemy as sa
+from sqlalchemy.ext.compiler import compiles
 
 from ...typing import SaFromClause, SaSelect, SaColumn, SaExpression, SaLambdaSelect
 from ...sql.sqlalchemy_types import Geometry
 from ...connection import SearchConnection
 from ...types import SearchDetails, DataLayer, GeometryFormat
 from ...results import SearchResults
+
+
+_LTREE_INVALID = re.compile(r'[^A-Za-z0-9_-]')
+
+
+def _sanitize_label(part: str) -> str:
+    """ Turn an OSM key or value into a valid ltree label, mirroring the
+        Lua sanitize_label(): invalid characters map to 'yes', hyphens to
+        underscores (for PostgreSQL < 16 compatibility).
+    """
+    if not part or _LTREE_INVALID.search(part):
+        return 'yes'
+    return part.replace('-', '_')
+
+
+def category_ltree(cls: str, typ: str) -> Optional[str]:
+    """ Build the 'osm.<class>.<type>' ltree string for a class/type pair,
+        mirroring the Lua get_category(). Returns None when the class cannot
+        form a valid ltree label, in which case no stored category can match.
+    """
+    if not cls or _LTREE_INVALID.search(cls):
+        return None
+    return f'osm.{_sanitize_label(cls)}.{_sanitize_label(typ)}'
+
+
+class CategoryMatch(sa.sql.expression.FunctionElement[Any]):
+    """ Match a placex row against a single (class, type) category.
+
+        On PostgreSQL this queries the ltree 'categories' column using the
+        GiST index (``categories <@ 'osm.<class>.<type>'``). On SQLite, which
+        has no ltree support, it falls back to class/type equality; those
+        columns are still present and populated.
+    """
+    inherit_cache = True
+
+    def __init__(self, table: SaFromClause, ltree: str, cls: str, typ: str) -> None:
+        super().__init__(table.c.categories, sa.literal(ltree),
+                         table.c.class_, sa.literal(cls),
+                         table.c.type, sa.literal(typ))
+
+
+@compiles(CategoryMatch)
+def _default_category_match(element: CategoryMatch,
+                            compiler: 'sa.Compiled', **kw: Any) -> str:
+    cats, ltree, _, _, _, _ = list(element.clauses)
+    return "(%s <@ (%s)::ltree)" % (compiler.process(cats, **kw),
+                                    compiler.process(ltree, **kw))
+
+
+@compiles(CategoryMatch, 'sqlite')
+def _sqlite_category_match(element: CategoryMatch,
+                           compiler: 'sa.Compiled', **kw: Any) -> str:
+    _, _, cls_col, cls_lit, typ_col, typ_lit = list(element.clauses)
+    return "(%s = %s AND %s = %s)" % (compiler.process(cls_col, **kw),
+                                      compiler.process(cls_lit, **kw),
+                                      compiler.process(typ_col, **kw),
+                                      compiler.process(typ_lit, **kw))
+
+
+def category_filter(table: SaFromClause, cls: str, typ: str) -> SaExpression:
+    """ Build a boolean expression selecting rows of the given class/type
+        category. Uses the ltree categories column on PostgreSQL and falls
+        back to class/type equality where no ltree category can exist.
+    """
+    ltree = category_ltree(cls, typ)
+    if ltree is None:
+        return sa.and_(table.c.class_ == cls, table.c.type == typ)
+    return CategoryMatch(table, ltree, cls, typ)
 
 
 class AbstractSearch(abc.ABC):
--- a/src/nominatim_api/search/db_searches/poi_search.py
+++ b/src/nominatim_api/search/db_searches/poi_search.py
@@ -12,7 +12,7 @@ import sqlalchemy as sa
 from . import base
 from ..db_search_fields import SearchData
 from ... import results as nres
-from ...typing import SaBind, SaRow
+from ...typing import SaBind
 from ...sql.sqlalchemy_types import Geometry
 from ...connection import SearchConnection
 from ...types import SearchDetails, Bbox
@@ -46,54 +46,32 @@ class PoiSearch(base.AbstractSearch):
 
         t = conn.t.placex
 
-        rows: list[SaRow] = []
+        # All searches go through the categories column, which is backed by the
+        # combined centroid/categories GiST index. Filter and order on the
+        # centroid, so that the index can serve both conditions at once.
+        sql = base.select_placex(t)\
+                  .where(sa.or_(*(base.category_filter(t, *category)
+                                  for category in self.qualifiers.values)))\
+                  .where(t.c.linked_place_id == None)
 
-        category_match = sa.or_(*(base.category_filter(t, *category)
-                                  for category in self.qualifiers.values))
-
-        if details.near and details.near_radius is not None and details.near_radius < 0.2:
-            # simply search in placex table
-            sql = base.select_placex(t) \
-                      .add_columns((-t.c.centroid.ST_Distance(NEAR_PARAM))
-                                   .label('importance'))\
-                      .where(t.c.linked_place_id == None) \
-                      .where(t.c.geometry.within_distance(NEAR_PARAM, NEAR_RADIUS_PARAM)) \
-                      .where(category_match) \
-                      .order_by(t.c.centroid.ST_Distance(NEAR_PARAM)) \
-                      .limit(LIMIT_PARAM)
-
-            if self.countries:
-                sql = sql.where(t.c.country_code.in_(self.countries.values))
-
-            if details.viewbox is not None and details.bounded_viewbox:
-                sql = sql.where(t.c.geometry.intersects(VIEWBOX_PARAM))
-
-            if details.excluded:
-                sql = sql.where(base.exclude_places(t))
-
-            rows.extend(await conn.execute(sql, bind_params))
+        if details.near is not None and details.near_radius is not None:
+            sql = sql.add_columns((-t.c.centroid.ST_Distance(NEAR_PARAM))
+                                  .label('importance'))\
+                     .where(t.c.centroid.within_distance(NEAR_PARAM, NEAR_RADIUS_PARAM))\
+                     .order_by(t.c.centroid.ST_Distance(NEAR_PARAM))
         else:
-            # use the categories column, backed by the combined
-            # centroid/categories GiST index
-            sql = base.select_placex(t)\
-                      .add_columns(t.c.importance)\
-                      .where(category_match)
+            sql = sql.add_columns(t.c.importance)
 
-            if details.viewbox is not None and details.bounded_viewbox:
-                sql = sql.where(t.c.centroid.intersects(VIEWBOX_PARAM))
+        if details.viewbox is not None and details.bounded_viewbox:
+            sql = sql.where(t.c.centroid.intersects(VIEWBOX_PARAM))
 
-            if details.near and details.near_radius is not None:
-                sql = sql.order_by(t.c.centroid.ST_Distance(NEAR_PARAM))\
-                         .where(t.c.centroid.within_distance(NEAR_PARAM, NEAR_RADIUS_PARAM))
+        if self.countries:
+            sql = sql.where(t.c.country_code.in_(self.countries.values))
 
-            if self.countries:
-                sql = sql.where(t.c.country_code.in_(self.countries.values))
+        if details.excluded:
+            sql = sql.where(base.exclude_places(t))
 
-            if details.excluded:
-                sql = sql.where(base.exclude_places(t))
-
-            sql = sql.limit(LIMIT_PARAM)
-            rows.extend(await conn.execute(sql, bind_params))
+        rows = await conn.execute(sql.limit(LIMIT_PARAM), bind_params)
 
         results = nres.SearchResults()
         for row in rows:
--- a/src/nominatim_api/search/db_searches/poi_search.py
+++ b/src/nominatim_api/search/db_searches/poi_search.py
@@ -73,17 +73,18 @@ class PoiSearch(base.AbstractSearch):
 
             rows.extend(await conn.execute(sql, bind_params))
         else:
-            # use the categories column, backed by the ltree GiST index
+            # use the categories column, backed by the combined
+            # centroid/categories GiST index
             sql = base.select_placex(t)\
                       .add_columns(t.c.importance)\
                       .where(category_match)
 
             if details.viewbox is not None and details.bounded_viewbox:
-                sql = sql.where(t.c.geometry.intersects(VIEWBOX_PARAM))
+                sql = sql.where(t.c.centroid.intersects(VIEWBOX_PARAM))
 
             if details.near and details.near_radius is not None:
                 sql = sql.order_by(t.c.centroid.ST_Distance(NEAR_PARAM))\
-                         .where(t.c.geometry.within_distance(NEAR_PARAM, NEAR_RADIUS_PARAM))
+                         .where(t.c.centroid.within_distance(NEAR_PARAM, NEAR_RADIUS_PARAM))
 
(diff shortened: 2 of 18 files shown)


Itz-Agasta:
`PoiSearch`'s near/small-radius branch changed from a single combined `OR` query to one query per category, so each returned row can be tagged with the category it matched. I tried mt best to keep the behaviour same..... flagging it since it touches the hot near-search path.

src/nominatim_api/search/db_searches/base.py
```diff
@@ -19,6 +21,74 @@
...
+    """ Build the 'osm.<class>.<type>' ltree string for a class/type pair,
+        mirroring the Lua get_category(). Returns None when the class cannot
+        form a valid ltree label, in which case no stored category can match.
+    """
+    if not cls or _LTREE_INVALID.search(cls):
+        return None
+    return f'osm.{_sanitize_label(cls)}.{_sanitize_label(typ)}'
+
+
+class CategoryMatch(sa.sql.expression.FunctionElement[Any]):
```
lonvia: Custom SQLAlchemy functions should rather go into `src/nominatim_api/sql/sqlalchemy_functions.py`.

src/nominatim_api/search/db_searches/base.py
```diff
@@ -19,6 +21,74 @@
...
+
+
+@compiles(CategoryMatch, 'sqlite')
+def _sqlite_category_match(element: CategoryMatch,
+                           compiler: 'sa.Compiled', **kw: Any) -> str:
+    _, _, cls_col, cls_lit, typ_col, typ_lit = list(element.clauses)
+    return "(%s = %s AND %s = %s)" % (compiler.process(cls_col, **kw),
+                                      compiler.process(cls_lit, **kw),
+                                      compiler.process(typ_col, **kw),
+                                      compiler.process(typ_lit, **kw))
```
lonvia: This means that we'd need slightly different indexes on the placex table for the SQLite variant. Instead of an index on category it needs an index on (class, type).

src/nominatim_api/search/db_searches/base.py
```diff
@@ -19,6 +21,74 @@
...
+
+
+def category_filter(table: SaFromClause, cls: str, typ: str) -> SaExpression:
+    """ Build a boolean expression selecting rows of the given class/type
+        category. Uses the ltree categories column on PostgreSQL and falls
+        back to class/type equality where no ltree category can exist.
+    """
+    ltree = category_ltree(cls, typ)
+    if ltree is None:
+        return sa.and_(table.c.class_ == cls, table.c.type == typ)
```
lonvia: You can't just fall back to class/type because in some cases an index is required.

The code will probably become easier, if you simply reject any class/type searches that don't conform to the category syntax. You can do that in the [token collection](https://github.com/osm-search/Nominatim/blob/master/src/nominatim_api/search/icu_tokenizer.py#L171) in the ICU tokenizer.

src/nominatim_api/search/db_searches/poi_search.py
```diff
@@ -48,7 +46,10 @@ async def lookup(self, conn: SearchConnection,
 
         t = conn.t.placex
 
-        rows: List[SaRow] = []
+        # Rows tagged with the category they were matched on. A placex row may
+        # carry several categories, so the matched one is not necessarily its
+        # class/type and must be tracked explicitly.
```
lonvia: It's fine to keep returning the class/type in the category field in `result`. The field is meant to be the main type of the object and not related to our new "categories". There is a bit of an unfortunate name clash here.

Also makes your code lots simpler, if you don't need to track the match category.

lonvia (commented):
The main question mark for me right now is that the new search functions really hit the indexes as expected. Make sure you run this code on the planet test server and run a few searches that use categories manually. You should recognise immediately if indexes are missing.

Also delete all `place_classtype_*` tables from the database and see if queries still run through and get you the right results. Especially run the geocoder_tester suite that way. (You can recreate the table any time by running `nominatim special-phrases` when you want to test migration in the next PR.)

Itz-Agasta:
```
psql -h localhost -p 5433 -U rupam -d nominatim -c "\di+ idx_placex_categories"
                                                List of indexes
 Schema |         Name          | Type  | Owner | Table  | Persistence | Access method |  Size   | Description 
--------+-----------------------+-------+-------+--------+-------------+---------------+---------+-------------
 public | idx_placex_categories | index | rupam | placex | permanent   | gist          | 2637 MB | 
(1 row)

```

Now results are not too good. on my first geotester run (without droping the table)
```
======== 8119 failed, 10913 passed, 3264 skipped in 3342.79s (0:55:42) =========
```
I droped the table and ran again, now 

```
======== 8119 failed, 10913 passed, 3264 skipped in 2708.70s (0:45:08) =========
```

I think this can be due to our lazy backfill design only 1,694,610 of 354,000,123 POI-level rows (rank_search >= 26) have categories populated -- under 0.5%. This is the same story as the restaurant-specific check, just at full scale.

all logs at https://filebin.net/dkpj4y1okwjtdpe1

Itz-Agasta:
Pulled the actual SQL from a debug-format search for `restaurants near Alexanderplatz` and ran it with `EXPLAIN (ANALYZE, BUFFERS)`:

```text
Nested Loop  (cost=172.38..113990.84 rows=7 width=72) (actual time=76.601..76.602 rows=0 loops=3)
  Join Filter: st_coveredby(placex.centroid, CASE WHEN (pgeom.rank_address > 9
    AND st_geometrytype(pgeom.geometry) = ANY ('{ST_Polygon,ST_MultiPolygon}'))
    THEN pgeom.geometry ELSE st_expand(pgeom.centroid, 0.05) END)
  Rows Removed by Join Filter: 17591
  ->  Parallel Bitmap Heap Scan on placex  (cost=171.81..8522.08 rows=3462 width=40)
        (actual time=18.195..38.241 rows=8795 loops=3)
        Recheck Cond: (categories <@ 'osm.amenity.restaurant'::ltree)
        ->  Bitmap Index Scan on idx_placex_categories  (cost=0.00..169.73 rows=8308 width=0)
              (actual time=7.876..7.876 rows=27057 loops=1)
              Index Cond: (categories <@ 'osm.amenity.restaurant'::ltree)
  ->  Index Scan using idx_place_id on placex pgeom  (cost=0.57..5.18 rows=2 width=217)
        Index Cond: (place_id = ANY ('{112292795,135962171}'::bigint[]))
Planning Time: 12.647 ms
Execution Time: 138.661 ms
```

The category index is being used correctly. `idx_placex_categories` gets hit with a Bitmap Index Scan and gives 27,057 candidate rows.

The expensive part seems to be the outer Nested Loop. For those candidates, `ST_CoveredBy(placex.centroid, ...)` is applied as a plain Join Filter instead of an index lookup, because `placex.centroid` doesn't have a spatial index, only `geometry` does. In this run, 17,591 rows are removed by that filter. The query returns 0 rows and takes around 140ms.

For comparison, I ran the same logical join against the old `place_classtype_amenity_restaurant` table, which has its own GiST index on `centroid`:

```text
Nested Loop  (cost=0.99..2134.97 rows=1557 width=72) (actual time=2.862..3.177 rows=18 loops=1)
  ->  Index Scan using idx_place_id on placex pgeom (cost=0.57..2.59 rows=1 width=217)
        Index Cond: (place_id = 73489205)
  ->  Index Scan using idx_place_classtype_amenity_restaurant_centroid on
        place_classtype_amenity_restaurant c (cost=0.42..2130.82 rows=156 width=40)
        (actual time=2.742..3.054 rows=18 loops=1)
        Index Cond: (centroid @ CASE WHEN (...) THEN pgeom.geometry
                     ELSE st_expand(pgeom.centroid, 0.05) END)
        Filter: st_coveredby(centroid, CASE WHEN (...) ...)
        Rows Removed by Filter: 43
Execution Time: 3.399 ms
```

Here the `centroid @ box` condition drives an actual Index Scan using the GiST index on `centroid`. It finds 18 matching restaurants in 3.4ms instead of scanning all category candidates and filtering most of them afterward.

So the category index itself looks fine..... The performance difference seems to come from losing the spatial index on `centroid` that the old `place_classtype_*` tables had.

Itz-Agasta:
hey @mtmail  good news from my side... the reports are in 

**master:**

```text
======== 7919 failed, 11113 passed, 3264 skipped in 2006.45s (0:33:26) =========

============================ slowest 500 durations =============================
5.82s call     geocoder_tester/world/france/lorraine/test_addresses.csv::9  57130
4.58s call     geocoder_tester/world/france/lorraine/test_addresses.csv::21  57160
4.05s call     geocoder_tester/world/france/lorraine/test_addresses.csv::29  57155
3.76s call     geocoder_tester/world/france/lorraine/test_addresses.csv::31  57000
3.59s call     geocoder_tester/world/france/lorraine/test_addresses.csv::89  57050
```

**PR branch:**

```text
========= 7919 failed, 11113 passed, 3264 skipped in 988.62s (0:16:28) =========

=========================== slowest 500 durations =============================
2.42s call     geocoder_tester/world/france/alsace/test_addresses.csv::34 Rue de la Gare 68380
2.30s call     geocoder_tester/world/france/centre/test_addresses.csv::14 Rue de la Gare 36310
1.81s call     geocoder_tester/world/france/alsace/test_addresses.csv::3A Rue de l'Église 67530
1.76s call     geocoder_tester/world/france/paysdelaloire/test_addresses.csv::1 Rue de l'Église 49260
1.53s call     geocoder_tester/world/france/paysdelaloire/test_addresses.csv::3 Rue de l'Église 85670
```

yep... you guessed it right.... the tests are now running about **2x faster** on the PR branch.

for comparison, this was the last PR's test status before it was merged:

```text
======== 7923 failed, 11109 passed, 3264 skipped in 1759.55s (0:29:19) =========

============================ slowest 500 durations =============================
4.80s call     geocoder_tester/world/france/lorraine/test_addresses.csv::9  57130
3.73s call     geocoder_tester/world/france/lorraine/test_addresses.csv::21  57160
3.20s call     geocoder_tester/world/france/lorraine/test_addresses.csv::23  57130
3.18s call     geocoder_tester/world/france/lorraine/test_addresses.csv::89  57050
3.08s call     geocoder_tester/world/france/lorraine/test_addresses.csv::29  57155
```

waiting for your master and PR branch test results 

logs @ https://filebin.net/69ykp2uruzjmkyin

Itz-Agasta:
but indexing is still an issue... i need to investigate it more.

I ran the query `restaurants near Alexanderplatz`, which returns 160 rows.


Plan | Time
-- | --
Old place_classtype_amenity_restaurant (GiST on centroid) | 8.2 ms
PR: BitmapAnd(idx_placex_geometry, idx_placex_categories) | 655 ms (warm) / 2618 ms (cold)
PR: enable_bitmapscan=off (geometry index scan + filter) | 630 ms


The bottleneck is unambiguous:

```text
-> BitmapAnd (actual time=327.220..327.221 loops=2)
     -> Bitmap Index Scan on idx_placex_geometry
          (actual time=10.899 rows=26842 loops=2)
     -> Bitmap Index Scan on idx_placex_categories
          (actual time=311.521 rows=1792883 loops=2)   <------- every restaurant on Earth
```

The geometry index scans 26,842 rows in about 11 ms, while the category index scans 1.79 million rows in about 311 ms. It looks like pg builds the bitmap for every restaurant in the database first and only then intersects it with the geometry bitmap, which ends up dominating the query time :(

Itz-Agasta:
As discussed ... i built both column orders on my db

**`(centroid, categories gist__ltree_ops(siglen=8))` is the one to go with.** 5125 MB vs 5376 MB for the other order, and 12:22 vs 19:45 to build. The current `idx_placex_categories` is 11 GB, so siglen=8 roughly halves it.

Numbers below are for Berlin `amenity=restaurant`. Every row is 7 runs on a fresh connection with the first discarded, median of the rest - cold vs warm moves these by 10-40x, so mixing the two is meaningless.

| | POI bounded viewbox | near search |
|---|---|---|
| master, via `place_classtype` | 0.69 ms | 22.6 ms |
| this PR as it stands | 106.5 ms | 510 ms |
| centroid + existing categories index | 183 ms | 4530 ms |
| centroid + combined index | **1.28 ms** | **75 ms** |

Cold, the PR's POI search is 3.2 s and near search 1.7 s, which is presumably what turned up as a timeout in marcs test.

Two things worth being straight about. The combined index does **not** fully match `place_classtype` - it's about 2x slower on POI and 3x on near search. But it is the same order of magnitude, against 428 tables and 8.2 GB replaced by a single 5.1 GB index, and it takes near search from 510 ms to 75 ms.

The second is that **neither half of the fix works on its own**. Switching back to centroid with the categories index we have is row 3, and it is *worse* than what's in the PR now. `place_classtype_*` is `(place_id, centroid)` with a GiST index on centroid, so master's category search was always centroid-based; this PR moved both searches onto `placex.geometry`, which changes the filter semantics *and* trades a 1.5M-row centroid index for the 22 GB `idx_placex_geometry` over 376M rows. Both have to change together.

So the proposed change is:

- `poi_search.py` - four viewbox filters back from `geometry` to `centroid`
- `near_search.py` - drop the `geometry &&` prefilter, keep the centroid `ST_CoveredBy`
- index definition in `lib-sql/indices.sql` + the migration --> `USING GIST (centroid, categories gist__ltree_ops(siglen=8))`




 lemme know what y think

Itz-Agasta:
One correction on the reasoning for centroid-first in my last mail, since it wasn't what I expected.

The argument was that a leading `centroid` column would let the index also serve centroid-only queries elsewhere. In practice **both** orders serve a centroid-only query fine - which matches the docs, a multicolumn GiST index [can be used with any subset of its columns](https://www.postgresql.org/docs/current/indexes-multicolumn.html), unlike btree.

Ig what actually blocks the reuse is the **partial predicate**. A plain `centroid && bbox` can't use either index and falls back to a parallel seq scan taking tens of seconds - the query has to also carry `categories IS NOT NULL`....So this doesn't give us a general centroid index; that would still want a separate non-partial one.

centroid-first is still the right pick in my opinion, on size and build time. On query time the two orders were within noise of each other on the POI case, with centroid-first slightly ahead on near search, on a centroid-only query, and on a second case with a rarer category (London bbox + `amenity.pharmacy`) - so it's not specific to Berlin or to common categories.

lonvia:
Ah, didn't know multicolumn GIST behaves differently. Good to know because it leaves open other index-based category filtering options in the future, even if we don't need them right now.

Could you check if the queries become significantly slower when dropping the partial predicate and building a full GIST index? I've created this kind of index (with order category, centroid) on a planet placex table with a full category set (as you get after an import) and it comes out at 25GB. That's an upper bound for the size to expect. It is a bit on the high side but still doable, especially if we can drop the 14GB geometry index in the future.

The slow-down for the queries is acceptable given the added functionality. So lets go ahead with that plan.

One more thing: there is still `idx_placex_categories` created by the migration that won't be needed anymore. Make sure you give the combined index a different name. Then you can simply delete the `idx_placex_categories` index creation from the older migration. That way the index won't be created for anybody doing a migration after this PR has been merged.

Itz-Agasta:
> Could you check if the queries become significantly slower when dropping the partial predicate and building a full GIST index? I've created this kind of index (with order category, centroid) on a planet placex table with a full category set (as you get after an import) and it comes out at 25GB. That's an upper bound for the size to expect.

Done. Short answer: no, and the full index is clearly the better one.

I rebuilt the partial index alongside the full one rather than comparing against my earlier numbers, so both are measured in the same session against the same cache state.

| | partial | full |
|---|---|---|
| build | 12:46 | 1:00:32 |
| size | 5134 MB | **22 GB** |

Query cost of dropping the predicate:

| query | partial | full | |
|---|---|---|---|
| POI bounded viewbox, Berlin `amenity=restaurant` | 1.06 ms | 1.18 ms | +11% |
| POI bounded viewbox, London `amenity=pharmacy` | 1.84 ms | 2.16 ms | +18% |
| near search, base place Berlin | 71.6 ms | 73.5 ms | +3% |

And what it buys - a plain `centroid && bbox` with no category condition at all:

| | `LIMIT 50` | `count(*)` over the bbox |
|---|---|---|
| no combined index (i.e. today) | 770.6 ms | 104.9 s |
| combined, partial | 613.9 ms | 95.9 s |
| **combined, full** | **0.40 ms** | **0.54 s** |

The partial index cannot serve those queries - the predicate makes it unusable and the planner falls back to a parallel seq scan, which is why it sits level with having no index at all. The full index turns the same queries into index scans. So dropping the predicate costs 10-20% on the category queries and takes the centroid-only case from a seq scan to an index scan.

That is the trade I thought would need a *separate* non-partial index in my last comment. It doesn't - dropping the predicate on this one index gets us both.

Method as before: 7 runs per row on a fresh connection with the first discarded, median of the remaining 6, every configuration wrapped in the same `BEGIN; DROP INDEX <the others>; EXPLAIN ...; ROLLBACK;` so the wrapper cost is
identical across rows. All near-search configurations return the same 4710 rows and all POI ones 50, so nothing is being won by a plan that quietly returns something different. As a cross-check on the method, the partial index measured 1.06 ms / 71.6 ms here against 1.28 ms / 75 ms in the run above.

One caveat on the 22 GB, because it is easy to misread against your 25 GB. Mine server is a migrated planet, so only 72,570,558 of 376,223,510 rows have a category - 19.3%. **(maybe this can be a reason too why PR branch on Marc's server gives less fails)**. The full index is 22 GB *at that coverage*, which says the size is dominated by indexing all 376M centroids rather than by the category signatures. Going from 19.3% to full coverage only adds siglen=8 signature bytes, so your 25 GB looks about right, and I'd expect it to be fairly insensitive to how many rows actually carry a category.

Itz-Agasta:
Btw what to do about idx_placex_categories on databases that already ran the 5.3.99-2 migration?

Itz-Agasta:
Interesting all my tests pass locally tough 

image

Itz-Agasta:
`gist__ltree_ops(siglen=8)` needs pg 13. Opclass parameters didn't exist in 12, and ci-tests.yml runs ubuntu-22 on pg 12,... ahhhh  😥

lonvia:
> gist__ltree_ops(siglen=8) needs pg 13

Hmmm, given the index inflation without that parameter, I'm leaning towards deprecating support for pg 12. If we do that, can you please open a separate PR that moves the minimum version to 13 with all updates to documentation and CI.

src/nominatim_api/search/db_searches/poi_search.py
```diff
@@ -48,28 +46,21 @@ async def lookup(self, conn: SearchConnection,
...
-                    lambda: _base_query().where(t.c.class_ == cclass)
-                                         .where(t.c.type == ctype))
-            else:
-                sql = _base_query().where(sa.or_(*(sa.and_(t.c.class_ == cls, t.c.type == typ)
-                                                   for cls, typ in classtype)))
+            sql = base.select_placex(t) \
+                      .add_columns((-t.c.centroid.ST_Distance(NEAR_PARAM))
+                                   .label('importance'))\
+                      .where(t.c.linked_place_id == None) \
+                      .where(t.c.geometry.within_distance(NEAR_PARAM, NEAR_RADIUS_PARAM)) \
```
lonvia: Shouldn't that be on centroid now so it hits the index?

src/nominatim_api/search/db_searches/poi_search.py
```diff
@@ -82,32 +73,27 @@ def _base_query() -> SaSelect:
...
+                         .where(t.c.centroid.within_distance(NEAR_PARAM, NEAR_RADIUS_PARAM))
+
+            if self.countries:
+                sql = sql.where(t.c.country_code.in_(self.countries.values))
+
+            if details.excluded:
+                sql = sql.where(base.exclude_places(t))
+
+            sql = sql.limit(LIMIT_PARAM)
+            rows.extend(await conn.execute(sql, bind_params))
```
lonvia: On a more general note. Is it even necessary to distinguish these two cases now that they use exactly the same mechanism?

src/nominatim_db/tools/migration.py
```diff
@@ -608,3 +603,19 @@ def backfill_categories(conn: Connection, **_: Any) -> None:
...
+def add_centroid_categories_index(conn: Connection, **_: Any) -> None:
+    """ Add the combined centroid/categories index used by category search.
+
+    The index is not partial, so that it can also serve queries on the
+    centroid alone. A partial index cannot be used for those.
+    """
+    if not table_exists(conn, 'placex') or not table_has_column(conn, 'placex', 'categories'):
+        return
+
+    if table_exists(conn, 'search_name'):
```
lonvia: You will need this index also for reverse queries in the next phase when implementing include/exclude. So better create the index unconditionally here.

lonvia (commented):
The main thing missing at the moment seems to be the code for backfilling categories for POIs.

Itz-Agasta:
See the skipped test ?..  On previous master, [club=scout] returned `category=club` via the classtype table. Now it returns `category=amenity` for merged rows (our alphabetic winner model).

So what to do about it ? i mean alphabetic order is temporary anyways.

Itz-Agasta:
btw `near_search` still doesn't filter linked places while `poi_search` now does (master had the same asymmetry, so I left it)

Itz-Agasta:
> The main thing missing at the moment seems to be the code for backfilling categories for POIs.

Now the problem is my migrated db is already back filled with the data using raw sql while i was experimenting... and i forgot to track the time :( ......yk 303,652,952 rows need filling so  ~7.5 hrs - 6.5 -9 hrs ig? idk

should i do some kind of reverse sql and then run the migration to track? ... btw wherever we do the fun will be 1hrs +

test/bdd/features/api/search/queries.feature
lonvia: Just switch this to checking for presence of the specific objects by OSM type/id using "Then the result set contains". Probably best to restrict to W351713277 and W408844120. The third result is subject to deduplication which is not deterministic.

We need to discuss if it makes sense to expose categories at API level or not. Photon doesn't do it but that is mainly because it doesn't save them in a way that they are retrievable.

src/nominatim_db/tools/migration.py
```diff
@@ -608,3 +603,65 @@ def backfill_categories(conn: Connection, **_: Any) -> None:
...
+                    END
+                 || '.'
+                 || CASE
+                      WHEN placex.type IS NULL OR placex.type = ''
+                           OR placex.type !~ '^[A-Za-z0-9_-]+$'
+                        THEN 'yes'
+                      ELSE replace(placex.type, '-', '_')
+                    END
+               )::ltree]
+        WHERE categories IS NULL
```
lonvia, line 660: I thought we had decided to go for a backfill is only done for places with class/type that has a corresponding `place_classtype_*` table?

Itz-Agasta:
how can i e2e test the backfill migration fun? my server db is already migrated (and i forget the sql & timing aswell, cuz i was experimenting back then)

lonvia:
To test the migration itself, you should probably best do that on a smaller database from an extract on your laptop: import with the latest release of Nominatim (don't forget to add special phrases). Then check out your branch and run migration. That must work without errors.

For getting some timing you will need to resort to timing the SQL: you can clear the category column and then run first the backfill SQL from the last's PR migration and then the new one. Don't forget to disable triggers. Clearing the column will also be very long-running unfortunately. (If you haven't run migrations yet on the current planet database, you can also restore the 'master' state of the category column by deleting everything and running the last PR's backfill SQL. And then run and time migration on that.)

Itz-Agasta:
Setup (untimed): `UPDATE placex SET categories = NULL` touched 72,570,558 rows in **1:49:07**, then `VACUUM (ANALYZE) placex` took **28:56**. You were right that clearing the column is slow.

`nominatim admin --migrate`, exit 0, total **3:03:47**:

| step | version | rows | duration |
|---|---|---|---|
| `add_categories_schema` | 5.3.99-2 | — | 0 s (no-op, column existed) |
| `backfill_categories` + its 3 partial indexes | 5.3.99-2 | 22,223,275 | **35:17** |
| `backfill_poi_categories` | 5.3.99-3 | 46,222,253 | **1:26:56** |
| `add_centroid_categories_index` | 5.3.99-3 | 22 GB index | **1:01:32** |

So this PR adds **2:28:28** on top of #4106's 35:17 (last time it was ~40 mins). 376,223,510 `placex` rows, 12 cores, 125 GB RAM.

The index figure cross-checks against two earlier standalone builds of the same index on this machine at 1:00:32 and 1:01:07. GiST has no parallel build, so the cores don't help there.

Post-migration verification on the planet came out the same as on my local pg Switzerland: `missing_rank_lt26` 0, `missing_classtype` 0, `disagreeing` 0, `pending_index` 0, version 5.3.99-3. `with_categories` is 68,445,528, which is exactly 22,223,275 + 46,222,253 - the two backfills filled precisely their declared scopes and nothing else.

Disk: `placex` heap stayed at 208 GB and free space went 479 GB → 477 GB. The dead tuples from the NULL pass got reused by the backfills rather than accumulating, so there was no lasting bloat beyond the 22 GB index itself.

Itz-Agasta:
**A migrated db ends up with different data than a freshly imported one.** The Lua style fills `categories` for every place it inserts, so a fresh import has the column populated on every row. A migrated database only gets what the two backfills cover (This pr adds the last addon backfill) :

| | rows | with `categories` |
|---|---|---|
| fresh import on this branch (Monaco, no special phrases) | 4,095 | 4,095 (100%) |
| v5.3.2 import + special phrases + migrate (Switzerland) | 3,651,653 | 779,510 (21%) |
| this planet, after migrating | 376,223,510 | 68,445,528 (18%) |

The gap is the `dfae1304` restriction, which limits the POI backfill to class/type combinations that have a `place_classtype_*` table. The 1:26:56 above is a good argument for keeping it - without it the backfill is ~300M rows instead of 46M.

But it means a POI whose class/type has no classtype table is findable by category search on a freshly imported database and invisible on a migrated one. It's sharpest for anyone who never imported special phrases: there are no
classtype tables, the backfill returns immediately, and they get no POI categories at all - while the Monaco row above shows a fresh import gives 100% coverage with no special phrases involved.

what y think?

src/nominatim_db/tools/migration.py
```diff
@@ -608,3 +603,95 @@ def backfill_categories(conn: Connection, **_: Any) -> None:
...
+    maintain it.
+    """
+    if not table_exists(conn, 'placex') or not table_has_column(conn, 'placex', 'categories'):
+        return
+
+    with conn.cursor() as cur:
+        cur.execute("""SELECT tablename FROM pg_tables
+                       WHERE schemaname = 'public'
+                             AND tablename LIKE 'place\\_classtype\\_%'""")
+        classtype_tables = [row[0] for row in cur]
```
lonvia, line 635: If you filter out here all class/type combinations that do not fit the category pattern, then lines 643-655 can be greatly simplified (you'd already know that the class and type will match the regex).

The outcome would not be 100% the same because we'd loose some category values instead of falling back to 'yes' but that would be very much acceptable.

lonvia (commented):
I think we are mostly good with this (minus that one comment).

3h migration is very long but acceptable for such a huge refactoring change. We'll have to add a warning in the migration notes but I'd like to defer that until after this has run in production and we have numbers from there.

Not running the migration when there are classtype tables is exactly what I was shooting for. That means that at least the installation that do not care for special-phrase search are saved from the expensive backfill.

lonvia:
This change is now live on the production servers. Migration took between 3.5 and 8.5h, depending on how powerful the server is. The good news is that there was no issue doing the migration under load, at least on a quiet Saturday evening. After migration it took a while to get back to the optimal working set using the new index but now query times are back in the same range as before.

In short: all good. Well just have to add a warning about the long migration in the documentation.

PR #4149, opened 2026-07-15
Add `clean_multilingual_names` Sanitizer

### LINKED ISSUE #4125: Add sanitizer for multi-lingual names
In areas where multiple languages are spoken, OSM usually has a name tag which contains all the language variants. This can for example be seen in [Brussels](https://www.openstreetmap.org/relation/58274) not only in the name of the city itself but also on every street within the city. The language variants then appear again in separate language-specific name tags.

It would be nice to have a sanitizer which detects when a `name` tag only consists of a concatenation of names already present in language-specific name tags and removes the name in that case.

Note that the rules on how the names are concatenated are region-specific and can be complex. [This wiki page](https://wiki.openstreetmap.org/wiki/Multilingual_names) list some current practice. Essentially, there are four ways of concatenating: slash, hyphen, semicolon and simple space. Ideally the sanitizer can just deal with all of them. We definitely need lots of tests derived from examples on this site.

AyushDharDubey:
Closes #4125

## Summary
Adds a new sanitizer, `clean_multilingual_names`, to detect and remove redundant base name tags that are simple concatenations of language-specific tags.

Default delimiters used: ["/", " - ", " — " ";", " "].

## AI usage
Unit tests were extensively co-authored with AI assistance.

### DIFF
--- a/src/nominatim_db/tokenizer/sanitizers/clean_multilingual_names.py
+++ b/src/nominatim_db/tokenizer/sanitizers/clean_multilingual_names.py
@@ -1,0 +1,107 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+# This file is part of Nominatim. (https://nominatim.org)
+#
+# Copyright (C) 2026 by the Nominatim developer community.
+# For a full list of authors see the git log.
+"""
+Sanitizer that removes name tags which are a simple concatenation of
+language-specific name tags.
+
+In areas where multiple languages are spoken, OSM often has a `name` tag
+which contains all language variants joined by a delimiter (e.g.
+``name=Rue du Marché aux Poulets - Kiekenmarkt``). These language variants
+also appear in separate tags like ``name:fr`` and ``name:nl``. The
+combined name is redundant and adds noise to the search index.
+
+This sanitizer detects when a bare ``name`` tag (i.e. one without a
+language suffix) consists entirely of a concatenation of names that are
+already present in language-specific name tags, and removes it. Exact
+duplicates are also considered redundant and removed, even when no
+concatenation is involved.
+
+Different delimiters used within a single name, i.e. ``Foo / Bar - Baz``
+is recognised as a concatenation as well. A trailing/leading delimiter is ignored, i.e.
+``- Foo - Bar - `` is still recognised.
+
+Arguments:
+    delimiters: A list of delimiter strings to try when checking if the
+                name is a concatenation.
+                (default: ``["/", "-", " "]``)
+
+    filter-kind: Define which 'kind' of names are considered for
+                 removal. Takes a string or list of strings where each
+                 string is a regular expression.
+                 (default: ``name``)
+"""
+import re
+from typing import Sequence
+
+from .base import ProcessInfo, SanitizerFunc
+from .config import SanitizerConfig
+
+
+def create(config: SanitizerConfig) -> SanitizerFunc:
+    """ Create a function to remove multilingual concatenated names.
+    """
+    delimiters: Sequence[str] = config.get_string_list('delimiters',
+                                                       ['/', '-', ' '])
+    filter_kind = config.get_filter('filter-kind', ['name'])
+
+    # Suffix having less than 4 letters before ending or being followed by a non-letter
+    # (eg. `name:en`, `name:fr-CA`, `name:zh-Hans`) is considered a valid language code.
+    illegal_lang_pattern = re.compile('(?:^[A-Za-z]{4,}.*)')
+
+    delim_alts = [r'\s*' + re.escape(d) + r'\s*' for d in delimiters]
+    delim_pattern = re.compile('(?:' + '|'.join(delim_alts) + ')')
+
+    def _is_concatenated(name_value: str, lang_names: set[str]) -> bool:
+        """ Check if name_value can be fully decomposed into values
+            from lang_names, optionally separated by delimiters.
+        """
+
+        lang_names_by_length = sorted(lang_names, key=len, reverse=True)
+        name_pattern = re.compile(
+            '(?:' + '|'.join(re.escape(n) for n in lang_names_by_length) + ')'
+        )
+
+        idx = 0
+        m = delim_pattern.match(name_value, idx)
+        if m:
+            idx = m.end()
+
+        while idx < len(name_value):
+            m = name_pattern.match(name_value, idx)
+            if not m:
+                return False
+            idx = m.end()
+
+            if idx < len(name_value):
+                m = delim_pattern.match(name_value, idx)
+                if not m:
+                    return False
+                idx = m.end()
+
+        return True
+
+    def _process(obj: ProcessInfo) -> None:
+        if not obj.names:
+            return
+
+        # Collect language-specific name values for each kind.
+        # A language-specific name has the same kind but a non-empty suffix.
+        lang_names_by_kind: dict[str, set[str]] = {}
+        for name in obj.names:
+            if name.suffix and not illegal_lang_pattern.match(name.suffix) \
+               and filter_kind(name.kind):
+                lang_names_by_kind.setdefault(name.kind, set()).add(name.name)
+
+        if not lang_names_by_kind:
+            return
+
+        obj.names = [name for name in obj.names
+                     if not (name.suffix is None
+                             and name.kind in lang_names_by_kind
+                             and _is_concatenated(name.name, lang_names_by_kind[name.kind]))]
+
+    return _process
--- a/test/python/tokenizer/sanitizers/test_clean_multilingual_names.py
+++ b/test/python/tokenizer/sanitizers/test_clean_multilingual_names.py
@@ -1,0 +1,221 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+# This file is part of Nominatim. (https://nominatim.org)
+#
+# Copyright (C) 2026 by the Nominatim developer community.
+# For a full list of authors see the git log.
+"""
+Tests for the sanitizer that removes multilingual concatenated names.
+"""
+import pytest
+
+from nominatim_db.config import Configuration
+from nominatim_db.data.place_info import PlaceInfo
+from nominatim_db.tokenizer.place_sanitizer import PlaceSanitizer
+
+
+@pytest.fixture
+def run_sanitizer(def_config):
+    def _run(extra_args=None, **kwargs):
+        place = PlaceInfo({'name': {k.replace('_', ':'): v
+                                    for k, v in kwargs.items()},
+                           'country_code': 'be', 'rank_address': 26})
+
+        sanitizer_args = {'step': 'clean-multilingual-names'}
+        if extra_args:
+            sanitizer_args.update(extra_args)
+
+        PlaceSanitizer([sanitizer_args], def_config).process_names(place)
+
+        return sorted([(p.name, p.kind, p.suffix or '')
+                       for p in place.searchable_names])
+
+    return _run
+
+
+class TestDelimiter:
+    """ Test removal with some common delimiters used worlwide.
+    """
+
+    def test_hyphen_delimiter(self, run_sanitizer):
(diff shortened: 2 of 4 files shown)


test/python/tokenizer/sanitizers/test_clean_multilingual_names.py
```diff
@@ -0,0 +1,181 @@
...
+                            name_fi='Helsinki',
+                            name_sv='Helsingfors',
+                            name_hi='हेलसिंकी')
+
+        assert ('Helsinki ; Helsingfors; हेलसिंकी', 'name', '') not in res
+        assert ('Helsinki', 'name', 'fi') in res
+        assert ('Helsingfors', 'name', 'sv') in res
+        assert ('हेलसिंकी', 'name', 'hi') in res
+
+    def test_extra_whitespace_in_parts(self, run_sanitizer):
```
mtmail: Use the relevant languages in the key names, not fr (French), nl (Dutch).

test/python/tokenizer/sanitizers/test_clean_multilingual_names.py
```diff
@@ -0,0 +1,181 @@
...
+        be stripped before comparison."""
+        res = run_sanitizer(name='गढ़वा ;    Garhwa',
+                            name_fr='गढ़वा',
+                            name_nl='Garhwa')
+
+        assert ('गढ़वा ;    Garhwa', 'name', '') not in res
+        assert ('गढ़वा', 'name', 'fr') in res
+        assert ('Garhwa', 'name', 'nl') in res
+
+    def test_space_delimiter(self, run_sanitizer):
```
mtmail, line 62: same as above. The field names should match the language of the values. I think that's English and Hindi here.

test/python/tokenizer/sanitizers/test_clean_multilingual_names.py
```diff
@@ -0,0 +1,181 @@
...
+        res = run_sanitizer(name='Rue du Poulet - Unknown',
+                            name_fr='Rue du Poulet',
+                            name_nl='Kiekenstraat')
+
+        assert ('Rue du Poulet - Unknown', 'name', '') in res
+
+    def test_single_language_tag_exact_match(self, run_sanitizer):
+        """A name identical to the only language tag is not a concatenation of
+        multilingual parts."""
+        res = run_sanitizer(name='गढ़वा', name_en='गढ़वा')
```
mtmail: name_en should be Hindi I think

test/python/tokenizer/sanitizers/test_clean_multilingual_names.py
```diff
@@ -0,0 +1,181 @@
...
+        res = run_sanitizer(name='गढ़वा - Garhwa-گڑھوا',
+                            name_hi='गढ़वा',
+                            name_en='Garhwa',
+                            name_ur='گڑھوا')
+
+        assert ('गढ़वा - Garhwa-گڑھوا', 'name', '') not in res
+        assert ('गढ़वा', 'name', 'hi') in res
+        assert ('Garhwa', 'name', 'en') in res
+        assert ('گڑھوا', 'name', 'ur') in res
+
```
mtmail, line 50: Add test where each name also already contains a hyphen. Example: "Berchem-Sainte-Agathe - Sint-Agatha-Berchem" (https://www.openstreetmap.org/node/66194565)

test/python/tokenizer/sanitizers/test_clean_multilingual_names.py
mtmail: Actually, can you use this place as test? "Oran ⵡⴻⵀⵔⴰⵏ وهران" (https://www.openstreetmap.org/relation/1259187). Then we test 3 names and another right-to-left language. `ⵡⴻⵀⵔⴰⵏ` is https://en.wikipedia.org/wiki/Berber_languages

mtmail (commented):
Looks good and comprehensive.

Throughout please make sure the language in the field names matches the values. E.g. `         name_fr='गढ़वा',
                            name_nl='Garhwa')` is neither French or Dutch.

Please add emdash and endash as separator, too (https://en.wikipedia.org/wiki/Dash#Em_dash). It's probably rare to see them in real data, often a mistake by the mapper/editor.

test/python/tokenizer/sanitizers/test_clean_multilingual_names.py
AyushDharDubey: Thanks for the review, I will fix the test cases,

I was doubtful on this specific case, should we remove the name tag in this scenario, when the name is the same as the only language name present?

src/nominatim_db/tokenizer/sanitizers/clean_multilingual_names.py
```diff
@@ -0,0 +1,83 @@
...
+            if name.suffix and filter_kind(name.kind):
+                lang_names_by_kind.setdefault(name.kind, set()).add(name.name)
+
+        if not lang_names_by_kind:
+            return
+
+        new_names = []
+        for name in obj.names:
+            if name.suffix is None \
+               and filter_kind(name.kind) \
```
lonvia: That check isn't necessary. It is included in the next one due to how `lang_names_by_kind` is constructed.

test/python/tokenizer/sanitizers/test_clean_multilingual_names.py
```diff
@@ -0,0 +1,199 @@
...
+        assert ('Oran ⵡⴻⵀⵔⴰⵏ وهران', 'name', '') not in res
+        assert ('Oran', 'name', 'en') in res
+        assert ('ⵡⴻⵀⵔⴰⵏ', 'name', 'ber') in res
+        assert ('وهران', 'name', 'ar') in res
+        assert ('Аран', 'name', 'be') in res
+
+    def test_extra_whitespace_in_parts(self, run_sanitizer):
+        """Parts with leading/trailing whitespace after splitting should
+        be stripped before comparison."""
```
lonvia: Nitpicking: please follow the usual formatting convention for comments:

```
        """ Parts with leading/trailing whitespace after splitting should
            be stripped before comparison.
        """
```

Same problem with other comments below.

test/python/tokenizer/sanitizers/test_clean_multilingual_names.py
```diff
@@ -0,0 +1,199 @@
...
+        assert ('गढ़वा    Garhwa', 'name', '') not in res
+
+    def test_space_delimiter_multiword_name(self, run_sanitizer):
+        """A normal multi-word name should NOT be split and removed if its parts
+        don't exactly match the language-specific names."""
+        res = run_sanitizer(name='New York City',
+                            name_en='New York City',
+                            name_fr='Ville de New York')
+
+        assert ('New York City', 'name', '') not in res
```
lonvia, line 99: There is a somewhat unintended side effect demonstrated here: the non-suffixed name will also be deleted when it is simply exactly the same as another without being a concatenation. That should be okay (and actually might save a bit of processing down the line) but best to explicitly document it.

lonvia (commented):
Looks good in general. Seeing this in code I have some more thoughts on the choice of delimiters:

I don't agree with @mtmail on the inclusion of m-dashes. The sanitizer doesn't need to catch all cases, just the most frequent ones and each additional delimiter adds another round of splitting for every single name tag on the planet.

The semicolon delimiter is actually not needed. There is a sanitizer in place that splits the name tags on semicolon, which also solves the underlying problem.

The space delimiter will not work correctly when a name contains a space, e.g. in this case: https://www.openstreetmap.org/node/2289263602 I am a bit worried in general that the space delimiter will hit on all multi-word names. I wonder if it would be possible to change this to a special check which breaks the string according to the script. I don't think you can do something like that with regular expression but maybe it is possible to use the RuleBasedBreakIterator from the [ICU library](https://unicode-org.github.io/icu/userguide/boundaryanalysis/break-rules.html) to create custom rules to that end. We already have pyICU as a required dependency, so that would be fine to use.

AyushDharDubey:
> The space delimiter will not work correctly when a name contains a space, e.g. in this case: https://www.openstreetmap.org/node/2289263602 I am a bit worried in general that the space delimiter will hit on all multi-word names. I wonder if it would be possible to change this to a special check which breaks the string according to the script. I don't think you can do something like that with regular expression but maybe it is possible to use the RuleBasedBreakIterator from the [ICU library](https://unicode-org.github.io/icu/userguide/boundaryanalysis/break-rules.html) to create custom rules to that end. We already have pyICU as a required dependency, so that would be fine to use.


I'm not quite well versed in ICU, so I'll have to look into it. I could solve this using recursive calls with memoization, which I believe would be O(N^2) on length of the name?

lonvia:
Here is another idea using a completely different algorithm: build a regex from the localized names instead of putting them in a set: `(<name:de>)|(<name:fr>)|...`. Also build a similar regex of all configured delimiters. Now start matching the name regex at the beginning of the general name. When it matches, match the delimiter regex on the beginning of the reminder string. If it matches, you have the delimiter in use. Now continue to match name regex followed by delimiter until you reach the end.

This way, you don't need to care at all about the content of the local names. They can switch scripts, contain supported delimiters, etc. It'd still work.

src/nominatim_db/tokenizer/sanitizers/clean_multilingual_names.py
```diff
@@ -0,0 +1,98 @@
...
+        m = name_pattern.match(name_value, idx)
+        if not m:
+            return False
+        idx = m.end()
+
+        while idx < len(name_value):
+            m = delim_pattern.match(name_value, idx)
+            if not m:
+                return False
+            idx = m.end()
```
lonvia, line 71: Can you add a test for a name with delimiter at the end: `name:fr=Foo` + `name:es=Bar` + `name=Foo - Bar - `

I think it will come out right from reading that code. Just want to be sure Python doesn't raise an exception somewhere.

src/nominatim_db/tokenizer/sanitizers/clean_multilingual_names.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _process(obj: ProcessInfo) -> None:
+        if not obj.names:
+            return
+
+        # Collect language-specific name values for each kind.
+        # A language-specific name has the same kind but a non-empty suffix.
+        lang_names_by_kind: dict[str, set[str]] = {}
+        for name in obj.names:
+            if name.suffix and filter_kind(name.kind):
```
lonvia: Just a tiny bit of paranoia here: can you check that suffix doesn't follow the pattern `[A-Za-z]{4,}.*`, i.e has less than 4 letters before ending or being followed by a non-letter. OSM has a few strange suffixes that actually denote non-names, for example [`name:prefix` and `name:suffix`](https://wiki.openstreetmap.org/wiki/Key:name:prefix). They might cause false positives together with the space delimiter.

lonvia (commented):
Nice. Depending on how fast Python is with regex compilation this should be faster as well.

The one major change to the previous version is that now different delimiters can be mixed. I can't think of a case right now where this would lead to additional problems. It might still be a good idea to document it because it is not entirely expected.

Two tiny comments, then this would be good to go.

lonvia:
Oh, I forgot: lets add this directly into the list of [default sanitizers](https://github.com/osm-search/Nominatim/blob/master/settings/sanitizers.yaml). I'd say before `split-name-list`.

src/nominatim_db/tokenizer/sanitizers/clean_multilingual_names.py
AyushDharDubey: This specific regex would allow suffixes similar to `name:1ab` too, but I don't think that should pose an issue.

lonvia: Yeah, I should have been more precise: `^[A-Za-z]{4,}.*`. I keep thinking in Python `match()` terms which matches the beginning.

lonvia:
Looks good, and seems to work well on a Brussels extract. Thanks for bearing with me.

PR #4152, opened 2026-07-25
Fix 500 error on category search without matching results

### LINKED ISSUE #4148: 500 Internal Server Error on category search with zero matching results
**Describe the bug**
Any `/search` request using the hidden `[key=value]` category syntax combined with a free-text near-phrase crashes with an uncaught IndexError, returned to the client as a generic 500 Internal Server Error, whenever the category search legitimately returns zero results.

Basically thr 	`ForwardGeocoder.sort_and_cut_results()` in src/nominatim_api/search/geocoder.py unconditionally indexes results[0]:
```
def sort_and_cut_results(self, results: SearchResults) -> SearchResults:
    results.sort(key=lambda r: (r.ranking, 0 if r.bbox is None else -r.bbox.area))
    final = SearchResults()
    min_rank = results[0].rank_search   # IndexError if results is empty
    min_ranking = results[0].ranking
    ...
```
git blame shows the guard against empty results was removed in commit 87a8c246a097cd9c218eb87871ba059d0435965a ("improve result cutting when a POI comes out with top importance", 2025-06-01), which previously wrapped this block in if results:. Still present and unfixed on current master (2991c61b).

This is confirmed independent of any specific category backend (old class/type place_classtype_* tables or the newer ltree-based categories column) -- it reproduces on production, which uses the old mechanism.

**To Reproduce**
```
curl -G "https://nominatim.openstreetmap.org/search" \
  --data-urlencode "q=[amenity=xyzxyz] Berlin" \
  --data-urlencode "format=json"
```
Response: {"title": "500 Internal Server Error"} (HTTP 500). amenity=xyzxyz is not a real OSM tag value, so this query is guaranteed zero category matches everywhere, ruling out a data/import-specific cause.

**Software Environment (please complete the following information):**
- Nominatim version: 5.3.99-2 (current master, commit 2991c61b)
- Postgresql version: 17.9 (Debian 17.9-0+deb13u1)
- Postgis version: 3.5.2
- OS: Debian (x86_64-pc-linux-gnu 

**Hardware Configuration (please complete the following information):**
- RAM: 
- number of CPUs:
- type and size of disks:
> Ahhh i cant recall

**Postgresql Configuration:**

NIL

**Nominatim Configuration:**

NIL

**Additional context**

Confirmed on two independent instances:
- nominatim.openstreetmap.org (production) -- [amenity=xyzxyz] Berlin -> HTTP 500
- our planet test instance -- same guaranteed-zero-match category, and the originally-noticed [club=scout] Vaduz -> HTTP 500

Itz-Agasta:
## Summary
Fixes #4148.

A `/search` request using the `[key=value]` category syntax together with a near-phrase returned a 500 error whenever the search legitimately found nothing:

```
curl -G "https://nominatim.openstreetmap.org/search" \
  --data-urlencode "q=[amenity=xyzxyz] Berlin" --data-urlencode "format=json"
{"title": "500 Internal Server Error"}
```

There turned out to be two independent `IndexError`s on this code path, both in `ForwardGeocoder.lookup_pois()`:

* `sort_and_cut_results()` accessed `results[0]` without checking that any results were found. This is the regression reported in the issue; the `if results:` guard was lost in 87a8c246.
* `execute_searches()` accessed `searches[0]` when the near-phrase produced no  search interpretations at all. `lookup_pois()` guarded this with `if query:`,  which is always true because `QueryStruct` defines neither `__bool__` nor  `__len__`. Replaced with `if searches:`, which is what `lookup()` already  does.

Which of the two triggers depends on whether the near-phrase is known to the tokenizer, so both need fixing for the query in the issue to work everywhere. `lookup()` was never affected: it calls `sort_and_cut_results()` under `if len(results) > 1:` and `execute_searches()` under `if searches:`.

Adds a BDD scenario in `test/bdd/features/api/search/queries.feature` covering both cases: a near-phrase that exists in the test data (`Vaduz`) and one that does not (`Berlin`).

## AI usage

None

### DIFF
--- a/src/nominatim_api/search/geocoder.py
+++ b/src/nominatim_api/search/geocoder.py
@@ -175,6 +175,9 @@ class ForwardGeocoder:
         """ Remove badly matching results, sort by ranking and
             limit to the configured number of results.
         """
+        if not results:
+            return results
+
         results.sort(key=lambda r: (r.ranking, 0 if r.bbox is None else -r.bbox.area))
 
         final = SearchResults()
@@ -271,7 +274,7 @@ class ForwardGeocoder:
         if phrases:
             query, searches = await self.build_searches(phrases)
 
-            if query:
+            if searches:
                 searches = [wrap_near_search(categories, s) for s in searches[:50]]
                 results = await self.execute_searches(query, searches)
                 results = self.pre_filter_results(results)
--- a/test/bdd/features/api/search/queries.feature
+++ b/test/bdd/features/api/search/queries.feature
@@ -113,6 +113,16 @@ Feature: Search queries
           | category | type |
           | leisure  | firepit |
 
+    # github #4148
+    Scenario Outline: Key/value search without any match returns no results
+        When geocoding "<query>"
+        Then exactly 0 results are returned
+
+        Examples:
+          | query                   |
+          | [amenity=xyzxyz] Vaduz  |
+          | [amenity=xyzxyz] Berlin |
+
     Scenario: POI search in a bounded viewbox
         When geocoding "restaurants"
           | viewbox                           | bounded |


lonvia:
Good catches, both of them!

PR #4163, opened 2026-08-06
Remove the place_classtype tables

Itz-Agasta:
## Summary

Follow-up to #4146. Now that category search goes through the ltree `categories` column, nothing reads the per-category `place_classtype_*` tables anymore, so this removes them and everything that maintained them.

On a planet database that is 428 tables and around ~8+GB of disk, plus a pg_tables lookup on every placex insert and delete during diff updates.

### Breaking change

The `--min` option of `nominatim special-phrases` is removed. It only decided which of the tables were created and never filtered the phrases themselves, so it has no meaning without them. `get_classtype_pairs()` goes with it. Noted in `docs/admin/Migration.md`. (i will make a sperate pr for docs aswell)

## AI usage

Yes the PR has mostly AI involvement. I asked it to confirm that no search service touches the `place_classtype_*` tables, audit what still references them, and remove those references. (Reviewed by me before commit btw)


Itz-Agasta:
btw `SPImporter` no longer touches the db at all, so the conn it takes and the now empty commit() call could go.... It is a small cleanup but it drags in the CLI and tests, so will do a follow up pr for it soon

src/nominatim_db/clicmd/special_phrases.py
```diff
@@ -58,8 +58,6 @@ def add_args(self, parser: argparse.ArgumentParser) -> None:
                            help='Import special phrases from a CSV file')
         group.add_argument('--no-replace', action='store_true',
                            help='Keep the old phrases and only add the new ones')
-        group.add_argument('--min', type=int, default=0,
-                           help='Restrict special phrases by minimum occurance')
```
lonvia, line 62: Rather than completely removing the option, just mark it as "Deprecated, no longer has any effect" and simply print a warning when it is used. That gives people a bit of time to adopt their scripts.

lonvia (commented):
Looks okay. I can't find more than the AI could.

Regarding the special phrase importer: it should actually be switched to add category names into the `word` table instead of class/type at some point. But that is a larger refactoring as well.

Itz-Agasta:
pls have a look now

lonvia:
I've been thinking some more about the migration that drops the `place_classtype_*` tables. It's better to just leave the tables in place and say in the migration guide that they can be manually dropped if desired. The reason for this is how migration works on a live server: you first run the migration script and then restart the API with the new code. That means there is a short period of time where the API server runs the old code but the migration has already run. The API servers have a cache to know which `place_classtype_*` tables exist. Thus, after the migration has run, there is a time where the API servers might produce SQL against the tables that no longer exist. I'd like to avoid that.

Can you remove the migration? Sorry about the back and forth. Should have thought about that earlier.

mtmail:
That reminds me https://github.com/ankane/strong_migrations (Ruby) which has some rules which operations in Postgresql data migrations are safe in a production environment and which should be split, e.g. when the migration would cause a full table lock. Depending on database size that could lock the application. The Ruby gem was written for companies like Shopify/Github/Gitlab.

Itz-Agasta:
mmm didnt though like that .. but ya  make sense

lonvia (commented):
Looks good now.

PR #4164, opened 2026-08-12
Add category filters to the search API

Itz-Agasta:
## Summary

Adds `include` and `exclude` to `/search`, filtering on the `categories` ltree column from #4106/#4146/#4163. Semantics followed [Photon's doc](https://github.com/komoot/photon/blob/master/docs/categories.md) + some stuff i thought out.

```
/search?q=restaurant+berlin&include=osm.amenity.restaurant
/search?q=berlin&include=osm.amenity                                    # matches descendants
/search?q=hilton&include=osm.tourism.hotel&include=osm.amenity.restaurant
/search?q=restaurants+in+berlin&exclude=osm.amenity.fast_food
```

| Form | Meaning |
|------|---------|
| `include=a.b` | place has `a.b` or a descendant of it |
| `include=a.b,c.d` | place has `a.b` **or** `c.d` |
| `include=a.b&include=c.d` | place has `a.b` **and** `c.d` |
| `exclude=a.b` | drop places having `a.b` |
| `exclude=a.b,c.d` | drop places having `a.b` **and** `c.d` |
| `exclude=a.b&exclude=c.d` | drop places having `a.b` **or** `c.d` |

Comma being OR for `include` and AND for `exclude` is Photon's behaviour: it falls out of De Morgan on their `should(mustNot(...))`. Note their `osm_tag=` is deprecated in favour of these, so I did not copy that syntax.

Invalid categories are a 400. A category needs at least two dot-separated labels, so `include=osm` is rejected. Hyphens are folded to underscores because the Lua importer stores them that way (Yk pg limitation and blh blh); without it `osm.shop.car-repair` would silently match nothing.

i decided to keep`SearchDetails.categories`  untouched and kept alongside. It is not redundant acc to me: it also restricts which special-phrase tokens are used (`56:427,444`), which `include` as a pure row filter cannot do.

The filter is applied to every search returning placex rows  - a parameter that silently does nothing on plain name queries seemed worse than none. Sources with no categories (postcodes, interpolations, TIGER, the country fallback tables) return nothing under `include`.

## AI usage

I was getting out of creativity for tests...so test cases +  its params are done by ai. Also for some sqllite changes ai helped me.. im not sure its correct or not as i didnt test them.

## Contributor guidelines (mandatory)

- [X] I have adhered to the [coding style](https://github.com/osm-search/Nominatim/blob/master/CONTRIBUTING.md#coding-style)
[...]

PR #4166, opened 2026-08-14
Update the documentation for the category series

Itz-Agasta:
## Summary

Documentation follow-up for #4106, #4146, #4163, and #4164 as part of the GSoC period.

## AI usage

After I was done,  asked ai to review it against the codebase and check if I was missing anything....It  made a few patches on its own.


docs/api/Search.md
```diff
@@ -229,6 +229,56 @@ to the address layer (see above).
...
+
+
+!!! note
+    Categories are only used for filtering, they are never part of the output.
+    The `category`/`class` field of a result contains the main tag of the place,
+    which is just one of its categories.
+
+!!! note
+    A place without any category cannot satisfy `include`. Postcodes, address
+    interpolations, TIGER housenumbers and countries that are only known from
```
mtmail: I don't think we need "that are only known from..." since countries never have type/class/category.

docs/customize/Import-Styles.md
```diff
@@ -82,6 +83,25 @@ and may be returned with the result [on request](../api/Search.md#output-details
...
 
+### Categories
+
+A category is derived from a main tag and written as `osm.<key>.<value>`.
+An object tagged with `tourism=hotel` and `amenity=restaurant` therefore gets
+the categories `osm.tourism.hotel` and `osm.amenity.restaurant`. Categories are
+hierarchical and can be used to filter search results, see the
+[search API](../api/Search.md#result-restriction).
+
+Only letters, digits, underscore and hyphen may appear in a category, where
```
mtmail: Is it letters or a stricter list of [a-z]?

mtmail:
Well caught that documentation for place_entrance and place_associated_street was missing so far.

docs/api/Search.md
Itz-Agasta: I don't think it's wrong to drop entirely..... it's actually a meaningful distinction. `CountrySearch.lookup` queries placex and applies `filter_by_category`, so a country backed by a real OSM boundary `(boundary=administrative+admin_level=2)` has a category and gets filtered like any other place. It's only the `lookup_in_country_table` fallback (no matching OSM boundary, backed by country_name/country_grid which have no categories column) that unconditionally returns nothing when include is set. 

tried to reword it pls check the new commit

docs/customize/Import-Styles.md
Itz-Agasta: yes it's letters, not just [a-z]...  the sanitizer in init.lua uses [^A-Za-z0-9_-] to detect invalid chars, so both cases pass through unchanged. 

https://github.com/osm-search/Nominatim/blob/f8da9c614a8a146145bbbf32796bd2da04e119e7/lib-lua/themes/nominatim/init.lua#L288-L298

docs/api/Search.md
```diff
@@ -229,6 +229,58 @@ to the address layer (see above).
...
+    boundary was found) therefore disappear from the results as soon as
+    `include` is used. `exclude` has no effect on them. Countries backed by
+    an OSM boundary keep their usual category and are filtered normally.
+
+!!! tip
+    Adding a filter may increase the number of results. Nominatim stops
+    searching once a search round has produced good enough matches. When the
+    filter removes such an early match, later rounds get to run and may find
+    places the unfiltered query never reached. The `countrycodes` and `layer`
+    filters behave the same way.
```
lonvia: That's an interesting (and correct) observation but likely raises more questions than it answers for the unsuspecting user. So more something for the developer section.

lonvia (commented):
Aaahhh, now we are really running into the ambiguity between the existing 'category' and the new 'categories'.

As there are already a lot of references to "main tag" when talking about the class/type columns, I would suggest that we stick to that name and drop all references to 'category' for that. So sooner or later we have the values in the category column as the only 'category'.

Also, can you please mark the 'categories' API parameter as deprecated. One of the next PRs should be to simply map it to include and some day it should go away completely (similar to Photon's `osm_tag`.)

lonvia (commented):
This looks good for now. Some further adaptions may be needed for follow-up changes, especially once we allow arbitrary categories.

PR #4168, opened 2026-08-16
Make POIs on unnamed streets searchable by address

### LINKED ISSUE #4133: POI object parented to an unnamed streets are not searchable by address
When a POI gets parented to a street, for example because it is a node on that street, and the street has no name, then the POI is only searchable by its name. The `nameaddress_vector` in the `search_name` table remains empty.

Example: https://nominatim.openstreetmap.org/ui/details.html?osmtype=N&osmid=8795855155&keywords=1

The reason for that is how we handle POI addresses in `create_poi_search_terms`. The first step is to retrieve the search terms for the parents:

https://github.com/osm-search/Nominatim/blob/6f74afef88a3a6209022b2e558448ebc47c173d5/lib-sql/functions/placex_triggers.sql#L412-L415

An unnamed street has no entry in `search_name`, so the variables are all set to NULL and the cheating further down

https://github.com/osm-search/Nominatim/blob/6f74afef88a3a6209022b2e558448ebc47c173d5/lib-sql/functions/placex_triggers.sql#L464-L467

doesn't work as expected.

If nothing is found, and the POI needs to be searchable (has name or housenumber), then the code should look up the address objects of the parent in `place_addressline`, retrieve their `name_vector` entries from `search_name` and merge them together to make up the nameaddress_vector.

Fix should come with a BDD test in `features/db/import/search_name.feature` that describes the situation.

pawank925:
Fixes #4133

POIs that are parented to an unnamed street do not inherit any address terms because the parent street has no entry in the search_name table. The nameaddress_vector in the search_name table stays empty and the POI can only be found by its own name.

When the parent has no search terms, look up the address objects of the parent in place_addressline, take their name_vector entries from search_name and merge them into the address terms of the POI.

Adds a BDD test that verifies a named POI on an unnamed street inside a city is found when searching for the name together with the city.

### DIFF
--- a/lib-sql/functions/placex_triggers.sql
+++ b/lib-sql/functions/placex_triggers.sql
@@ -409,6 +409,7 @@ DECLARE
 
   addr_item RECORD;
   addr_place RECORD;
+  parent_place RECORD;
   parent_address_place_ids BIGINT[];
 BEGIN
   nameaddress_vector := '{}'::INTEGER[];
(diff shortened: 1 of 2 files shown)


pawank925:
branch was wrong

PR #4171, opened 2026-08-17
Make POIs on unnamed streets searchable by address #4133

### LINKED ISSUE #4133: POI object parented to an unnamed streets are not searchable by address

pawank925:
## summary

Fixes #4133
When a POI is parented to an unnamed street (e.g. a node sitting on an unnamed residential way), the POI only shows up when searching by its own name. The `nameaddress_vector` in the `search_name` table stays empty because the parent street has no entry in `search_name`, so address terms are never inherited.

This means searching for "Blue house, Strange Town" doesn't return the POI even though it's inside the Strange Town boundary.

## Fix

In `create_poi_search_terms`, when the parent has no `search_name` row (`parent_address_vector IS NULL`) and the POI is searchable (has name or housenumber), look up the parent's address objects in `place_addressline`, get their `name_vector` entries from `search_name`, and merge them into the POI's address terms.

## Testing

- Added a BDD test scenario in `test/bdd/features/db/import/search_name.feature` that places a named POI on an unnamed street inside a city polygon and verifies geocoding by name + city returns the correct result.
- Verified locally: `make tests` passes with no regressions.

## Checklist

- [x] Tests pass locally
- [x] Single issue addressed (#4133)
- [x] No unrelated changes mixed in
 
## proof 

kkkk

### DIFF
--- a/lib-sql/functions/placex_triggers.sql
+++ b/lib-sql/functions/placex_triggers.sql
@@ -409,6 +409,7 @@ DECLARE
 
   addr_item RECORD;
   addr_place RECORD;
+  parent_place RECORD;
   parent_address_place_ids BIGINT[];
 BEGIN
   nameaddress_vector := '{}'::INTEGER[];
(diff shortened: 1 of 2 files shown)


pawank925:
@lonvia  please have  a look at  this pr thanks!

lonvia:
Looks good. Thank you.

PR #4184, opened 2026-09-03
keep secondary main tags in extratags

### LINKED ISSUE #4173: Secondary main tags need to go into extra tags
When the Lua code needs to choose between multiple main tags, it converts all main tags into categories but puts only one into class/type. The other main tags are dropped. It turns out that this means that the other main tags are no longer accessible in the results because we don't return categories. Secondary main tags should be pushed into extratags when the Lua script is configured to keep extra tags.

pawank925:
## summary

When an object carries several main tags, Nominatim converts all of them into categories but only places a single one into class/type. The remaining main tags were dropped entirely, so they were no longer accessible in the results (categories are not returned).

Keep secondary main tags in extratags instead of removing them from the extratags set. Only the winning main tag that became class/type is excluded from extratags. Because extratags are only collected when the Lua style is configured to keep extra tags, the behaviour in the default import (extra tags disabled) is unchanged.

Adds a BDD test covering secondary main tags being kept in extratags, and updates the existing main-with-extra and internally-used-tags tests for the new behaviour.

Fixes https://github.com/osm-search/Nominatim/issues/4173

## AI usage
This pull request was made manually and reviewed by Claude.  The change was verified by running the osm2pgsql BDD import tests (60 passed), and the previously-affected tests were updated accordingly.

### DIFF
--- a/lib-lua/themes/nominatim/init.lua
+++ b/lib-lua/themes/nominatim/init.lua
@@ -696,18 +696,12 @@ else
 end
 
 -- Build extratags for a merged row: filter via EXTRATAGS_FILTER,
--- remove sibling main tags (they belong in categories, not extratags),
--- then add any required extratags (wikipedia, wikidata, etc.).
-local function build_extratags(place, k, v)
+-- remove the winning main tag (it belongs in class/type, not extratags),
+-- keeping any secondary main tags, then add any required extratags.
+local function build_extratags(place, k, v, main_class)
     local extra = EXTRATAGS_FILTER(place, k, v) or {}
-    for ek, ev in pairs(extra) do
-        local ktable = MAIN_KEYS[ek]
-        if ktable ~= nil then
-            local transform = ktable[ev] or ktable[1]
-            if type(transform) == 'function' then
-                extra[ek] = nil
-            end
-        end
+    if main_class ~= nil then
+        extra[main_class] = nil
     end
     for tk, tv in pairs(place.object.tags) do
         if REQUIRED_EXTRATAGS_FILTER(tk, tv) and extra[tk] == nil then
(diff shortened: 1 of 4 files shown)


pawank925:
The change was verified by running the osm2pgsql BDD import tests (60 passed), and the previously-affected tests were updated accordingly.

lib-lua/themes/nominatim/init.lua
```diff
@@ -696,15 +696,13 @@ else
...
-        local ktable = MAIN_KEYS[ek]
-        if ktable ~= nil then
-            local transform = ktable[ev] or ktable[1]
-            if type(transform) == 'function' then
+    if main_class ~= nil then
+        for ek, ev in pairs(extra) do
+            if ek == main_class and ev == main_type then
                 extra[ek] = nil
             end
         end
```
lonvia: This looks overly complicated. You don't need to loop through extra to find a tag with the `main_class`. A simple `extra[main_class]` will do.

pawank925: Done, thanks for the suggestion. Simplified to `extra[main_class] = nil` and dropped the now-unused `main_type` parameter.

pawank925:
@lonvia Done, thanks for the suggestion.

lonvia (commented):
Can you please squash the fixup commit into the original one? Then this would be good to go.

lib-lua/themes/nominatim/init.lua
pawank925: Squashed the fixup into the original commit as requested. The branch now has two commits: the implementation and the test updates.

lonvia:
Thank you.

PR #4177, opened 2026-08-28
Fix 4172 drop unwanted categories

justpriyanshu:
## Summary

While the `is_rankable_place()` already excluded certain highway and boundary categories, it did not actually remove the problem category from the array. Instead, it simply decided to continue processing if the unwanted category was found via `CONTINUE`. This resulted in the `compute_place_rank()` seeing the unwanted category and assigning a rank.

The fix makes the `drop_unwanted_categories()` handle the filtering of the array, removing any problematic categories and passing what remains to the ranking function. Now the `is_rankable_place()` becomes a very simple function, just checking whether there is anything remaining after the filtration, without changing the filtration rules themselves.

Before, there were three calls to `compute_place_rank()` from `NEW.categories`, that did not filter the array first — `placex_insert()`, `placex_update()` and `place_insert()`. Now they all use the filtered array.

As noted above, this solution remains minimal, leaving `NEW.categories` array untouched and not changing the object class/type.

Issue : #4172

## AI usage

Used an AI assistant to help navigate the codebase, discuss the approach,
and Testing.

## Contributor guidelines (mandatory)

- [x] I have adhered to the [coding style](https://github.com/osm-search/Nominatim/blob/master/CONTRIBUTING.md#coding-style)
- [x] I have [tested](https://github.com/osm-search/Nominatim/blob/master/CONTRIBUTING.md#testing) the proposed changes
- [x] I have [disclosed](https://github.com/osm-search/Nominatim/blob/master/CONTRIBUTING.md#using-ai-assisted-code-generators) above any use of AI to generate code, documentation, or the pull request description

PR #4190, opened 2026-09-10
Make category assignment configurable via presets (#4180)

### LINKED ISSUE #4180: Allow to define custom categories in lua scripts
Now that we have categories, we also want to allow the user to customize them in the Lua import script.

Currently the user defines a main type by choosing one or more of the OSM tags of the object. This main type is used for two things:

* defining the primary tag of the object (class and type columns)
* defining the assigned categories in the `osm.` category class
  * These indirectly defines the base rank of the object.

Going forward, we'd want to break up the configuration:

* User chooses one OSM tag as the main type.
  * This indirectly defines the base rank of the object (aka going back to what we did before categories were introduced.)
* User chooses N categories.

The simple category configuration best works similar to what we have now: anything that looks like a 'main tag' becomes a category item in the `osm.` class. We just might allow some more refining, so that we get for example `osm.boundary.administrative.8` to distinguish admin levels or `osm.tourism.information.guidepost` to describe the information category better.

On top of that users should be able to plug in functions to add arbitrary other categorization schemas.

All this needs to be backwards compatible: when using a Lua script from an older version, things should remain working as before.

pawank925:
## summary
Allow to define custom categories in lua scripts
Currently the main tag of an object both defines the primary tag (class/type columns) and the assigned categories in the `osm.` category class. This makes it impossible to add arbitrary categorization to an object beyond the tags that are recognized as main tags.
This change allows users to plug in their own categorization functions. Register a Lua function via `add_custom_categories()` (append) or `set_custom_categories()` (replace), each returning arbitrary ltree category strings (e.g. `osm.boundary.administrative.8`). Main tag transforms may also return a `{categories = {...}}` table to attach custom categories while still setting class/type.
The default `osm.<key>.<value>` category continu es to be added automatically, and existing style files behave exactly as before.
Adds BDD tests covering custom category functions, multiple registered functions, the replacement behaviour of `set_custom_categories()`, and transforms returning custom category tables. All existing osm2pgsql import tests pass unchanged.

Fixes #4180

## AI usage

This pull request was made manually,tested manuallly every line  and reviewed by Claude. The change was verified by running the osm2pgsql BDD import tests (60 passed)

### DIFF
--- a/lib-lua/themes/nominatim/init.lua
+++ b/lib-lua/themes/nominatim/init.lua
@@ -32,6 +32,7 @@ local EXTRATAGS_FILTER
 local REQUIRED_EXTRATAGS_FILTER
 local POSTCODE_FALLBACK = true
 local ENTRANCE_FUNCTION = nil
+local CUSTOM_CATEGORY_FUNCS = {}
 
 -- This file can also be directly require'd instead of running it under
 -- the themepark framework. In that case the first parameter is usually
@@ -304,6 +305,18 @@ local function get_category(k, v)
         return nil
     end
     return 'osm.' .. sanitize_label(k) .. '.' .. sanitize_label(v)
+end
+
+local function sanitize_category_path(path)
+    if path == nil or path == '' then return nil end
+    local result = {}
+    for label in path:gmatch('[^%.]+') do
+        local s = sanitize_label(label)
+        if s == nil or s == '' then return nil end
+        table.insert(result, s)
+    end
+    if #result == 0 then return nil end
+    return table.concat(result, '.')
 end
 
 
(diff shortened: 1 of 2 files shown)


lib-lua/themes/nominatim/init.lua
```diff
@@ -306,6 +307,18 @@ local function get_category(k, v)
...
+    if path == nil or path == '' then return nil end
+    local result = {}
+    for label in path:gmatch('[^%.]+') do
+        local s = sanitize_label(label)
+        if s == nil or s == '' then return nil end
+        table.insert(result, s)
+    end
+    if #result == 0 then return nil end
+    return table.concat(result, '.')
+end
```
lonvia, line 319: I think it is better to leave sanitization to the category functions and just reject labels that do not comply with the format. We should also restrict the number of subcategories to 4 like Photon does.

lonvia (commented):
The approach looks sound in general, I would just suggest one fundamental change: the current category assignment should go into a custom category function and this function then is set in the current presets. That way, a user can easily decide to add further categories or use completely different ones.

However, given that some categories are used in address processing, it isn't really possible to completely drop the `osm.*` categories. So, in fact, there need to be two category assignment functions supplied by default: one implementing the current behaviour of getting all main tags and another one, which only adds the main categories that are necessary to ensure that address processing later goes smoothly.

lonvia:
I'm sorry but this is not making it better. What I was trying to say: the category computation needs to move out completely from the place where it is now and go into a separate function. Looking at the code I wonder if it is even possible to do without looking at the same time in the second part of #4180: assignment of the main category.

And looking at the mandatory minimal categories, it might be better to first move the exceptions in `compute_place_rank()` into the Lua code. (See discussion in #4177.)

pawank925:
@lonvia  category computation lives in its own function, and the rank exceptions are handled in Lua instead of SQL. Boundary-node tests updated accordingly

pawank925:
category computation is now fully in its own function, and moved the rank exceptions into Lua. Had to update 3 tests since boundary nodes now get filtered at the osm2pgsql level instead of in SQL

lonvia:
Would you mind making a separate PR for moving the rank exceptions? It is a lot easier to do a review when topics are separate. Once that is merged, we can come back to this one.

pawank925:
I split the rank exception port into its own pull request: #4196  The head branch of this PR now only contains the configurable-categories/presets work; it is built on top of the fork's master minus the rank-exception parts, and the full test_osm2pgsql.py suite passes (112 tests). Once #4196 is merged, this PR can be rebased on it without conflict

lonvia:
Closing as a violation of our AI policy.

PR #4195, opened 2026-09-14
Update fallback for mismatched class

### LINKED ISSUE #4181: Ignore class parameter on /details
With the move to categories we won't have multiple places for a single OSM object anymore. This makes the `class=` parameter useless in most cases. I wouldn't want to remove it completely but we should fall back to an object that matches with OSM id but has a different class, when the given class cannot be found.

davidbian1:
## Summary
Resolves #4181.

With the move to categories, the 'class=' parameter is useless in most cases. There should be a fall back to an object that matches with OSM id but has a different class, when the given class cannot be found. 

`/details` still matches against the 'class=' parameter, so a request for any other of the object's tags returns "not found" even though there is an object that matches the OSM id.

Updates `get_detailed_place` to fall back to returning the object by OSM ID when no entry matches the given class, instead of returning nothing, so `/details` request gives the correct result.

Adds a test scenario to verify the fallback and updates tests that previously relied on the no fallback scenario.

image

## AI usage

Claude plus manual review and testing


src/nominatim_api/lookup.py
```diff
@@ -168,6 +168,12 @@ async def get_detailed_place(conn: SearchConnection, place: ntyp.PlaceRef,
         if await func(conn, collector):
             break
 
+    if collector.result is None and isinstance(place, ntyp.OsmID) and place.osm_class:
+        fallback = DetailedCollector(ntyp.OsmID(place.osm_type, place.osm_id),
+                                     collector.with_geometry)
+        if await find_in_placex(conn, fallback):
+            collector.result = fallback.result
```
lonvia, line 175: This would create a second SQL call. Is this also solvable without sending a second query?

test/python/api/test_api_details.py
```diff
@@ -606,15 +619,17 @@ def test_lookup_postcode_with_address_details(apiobj, frontend, lookup):
...
+                                                     (napi.OsmID('N', 55, 'amenity'), 1)])
+def test_lookup_missing_object(apiobj, frontend, objid, expected_place_id):
     apiobj.add_placex(place_id=1, osm_type='N', osm_id=55,
                       class_='place', type='suburb')
 
     api = frontend(apiobj, options={'details'})
-    assert api.details(objid) is None
+    result = api.details(objid)
+
+    assert (result.place_id if result else None) == expected_place_id
```
lonvia: Split this test instead of extending with a parameter.

lonvia:
> Claude plus manual review and testing

I won't impose any restrictions on choice of workflow but if I may make an observation: if you are here because you are preparing for GSOC, then just reviewing work done by AI won't get you very far. The successful students turn this workflow around and have AI review their work instead.

src/nominatim_api/lookup.py
davidbian1: Yes, by stashing a possible candidate, but this requires modifying `DetailedCollector` to account for this. My understanding is any direct modification of the functions for the search tables changes the expected result (particularly when an entry exists in the `placex` and `location_postcodes` search tables but the class doesn't match in `placex`)

davidbian1:
- Test has been refactored
- Left comment about the changes needed to eliminate a second query, will go ahead if that matches the original intent

src/nominatim_api/lookup.py
lonvia: Fair enough. Duplicates between `placex` and `location_postcodes`/`osmline` are rare but not impossible. 

Alright, lets go with this solution.

Just  one thing: please do not assign internal attributes (i.e. `result`) from one collector to another. Save the result to a local variable instead and change the code below accordingly.

lonvia (commented):
One more feature request after looking at the tests: if the class looks like a number then do not fall back to placex. A number clearly indicates that an interpolation was meant.

